PR #4, opened 2020-09-10
Enhance stereo depth develop merge 

iamsg-luxonis:
Enhance stereo depth develop merge


include/depthai/device.hpp
```diff
@@ -36,6 +36,12 @@ class Device{
     );
     std::vector<std::string> get_available_streams();
 
+    std::vector<std::vector<float>> get_left_intrinsic();
+    std::vector<std::vector<float>> get_left_homography();
+    std::vector<std::vector<float>> get_right_intrinsic();
+    std::vector<std::vector<float>> get_right_homography();
+    std::vector<std::vector<float>> get_rotation();
+    std::vector<float> get_Translation();
```
themarpe, line 44: Rename to `get_translation`

include/depthai/host_data_packet.hpp
```diff
@@ -37,7 +37,10 @@ struct HostDataPacket
         // Check if metadata is valid
         if(metadata.isValid()){
             // copy only frame data
-            frameSize = metadata.frameSize;
+            if (stream_name_ == "disparity_color") // FIXME see if applicable to all streams
```
themarpe, line 40: metadata.frameSize should be set correctly. Maybe rather fix over at device side instead?

themarpe (changes requested):

include/depthai/host_data_packet.hpp
iamsg-luxonis: "disparity_color" is a virtual stream, created on the host ("disparity" colorized on host)

themarpe: Can you add the metadata to the frame while creating it and then send it onwards to "disparity_color" stream. Then decoded data will include everything needed - w/h, type, ...

iamsg-luxonis: I don't know why it was added here, but I agree, it should be added when it's created.
I will change it.

alex-luxonis: Fixed.

PR #5, opened 2020-09-10
Stream for retrieving raw color frames

themarpe:
This PR adds 'raw_color' stream for retrieving raw YUV420p frames of color camera


iamsg-luxonis (approved):
LGTM.

PR #7, opened 2020-09-18
Calibration fix when EEPROM is not programmed.

szabi-luxonis:
Corresponds to:
https://github.com/luxonis/depthai-python/pull/50


saching13 (approved):
LGTM

PR #13, opened 2020-10-12
Fix a scaling issue or crash with NN on rectified frames

alex-luxonis:
Should fix issue https://github.com/luxonis/depthai/issues/226


szabi-luxonis (approved):
LGTM

PR #14, opened 2020-10-12
Fetching USB speed and Myriad X serial number

saching13:
Created `usb_speed` and `mx_serial` private variables to store the usb speed and myriad x serial id.
Along with this added public functions `is_usb3`, `get_mx_id` and `is_eeprom_loaded`(factory testing purpose).
- `is_usb3` returns true if it is connected over usb3
- `get_mx_id` will return string of the Myriad X id
- `is_eeprom_loaded` returns true if both intrinsics are available after device init.


include/depthai/device.hpp
```diff
@@ -99,13 +112,16 @@ class Device{
     std::vector<float> d1_l;
     std::vector<float> d2_r;
     int32_t version;
-
+    bool device_changed = true;
```
szabi-luxonis, line 115: shouldn't this be false by default?

saching13: No. it helps considering the first device connected as also device change that's why I made it true by default. And I am resetting it to false once its purpose is served in testing.

src/device.cpp
```diff
@@ -593,6 +634,15 @@ std::vector<std::vector<float>> Device::get_right_homography()
     
 }
 
+bool Device::is_usb3()
+{
+        // if(usb_speed.find("Super") != std::string::npos)
+        if(g_xlink->getUSBSpeed().find("Super") != std::string::npos)
```
themarpe, line 640: Modify g_xlink->getUSBSpeed to return strong type (UsbSpeed_t) which makes this thing a tad nicer.

```
if(g_xlink->getUSBSpeed() == X_LINK_USB_SPEED_SUPER || g_xlink->getUSBSpeed() == X_LINK_USB_SPEED_SUPER_PLUS){
    return true;
}
return false;
```

Also either if statements with blocks or single line if when no block specified

saching13: Good idea. Changed it

themarpe (approved):

szabi-luxonis (approved):

PR #15, opened 2020-10-13
Remove depth confidence threshold

szabi-luxonis:
Removed `depth_confidence_threshold` since there is no reason to have both for decoding and depth.


src/pipeline/host_pipeline_config.cpp
```diff
@@ -105,17 +105,6 @@ bool HostPipelineConfig::initWithJSON(const nlohmann::json &json_obj)
                 depth.depth_limit_m = depth_obj.at("depth_limit_m").get<float>();
             }
 
-            if (depth_obj.contains("confidence_threshold"))
-            {
```
alex-luxonis, line 109: If users have custom code / python scripts still setting this field, not sure if we should warn them that it got removed.

alex-luxonis (approved):
LGTM

PR #16, opened 2020-10-15
Gen1 updates: FrameMetadata all streams, sync seqNo, IMX477 RPi HQcam

alex-luxonis:
Add an experimental option to synchronize the sequence numbers of RGB and Mono cameras MIPI source, and thus all packets should have `seqNo` in sync. RGB camera starts first and will stream frames. After several frames when the Mono camera start too, their `seqNo` will not begin with zero, but start from the current RGB `seqNo`.

Other device side changes:
- FrameMetadata is being populated and sent out for all streams. Previously it was missing for `jpegout, video, object_tracker`. The metadata for the new and existing streams should have all fields properly initialized.
- Fix device crash in some scenarios when too many streams were enabled.
- Add support for IMX477 RPi HQ camera. Allow operation with missing EEPROM / VCM ICs for the RGB cameras modules (IMX378 or IMX477).

Related PR: https://github.com/luxonis/depthai-shared/pull/8

### DIFF
--- a/cmake/Depthai/DepthaiDeviceSideConfig.cmake
+++ b/cmake/Depthai/DepthaiDeviceSideConfig.cmake
@@ -2,7 +2,7 @@
 set(DEPTHAI_DEVICE_SIDE_MATURITY "snapshot")  
 
 # "full commit hash of device side binary"
-set(DEPTHAI_DEVICE_SIDE_COMMIT "cc7b6c4296bd31e30dae65e4d5f1552947750dba")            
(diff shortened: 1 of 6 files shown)


szabi-luxonis (approved):
LGTM

Luxonis-Brandon:
Sweet, thanks.  All of these will be very valuable!

alex-luxonis:
XLink changes here: https://github.com/luxonis/XLink/pull/2

PR #20, opened 2020-10-23
Update device FW: fix X,Y,Z calc on NN bounding box

alex-luxonis:
Should fix issue https://github.com/luxonis/depthai/issues/244


szabi-luxonis (approved):
LGTM

PR #22, opened 2020-10-31
Fix crash on 2nd device object delete

alex-luxonis:
... due to watchdog thread not recreated.
Now running create+delete in a loop should work fine.

Should fix issue https://github.com/luxonis/depthai/issues/259


szabi-luxonis (approved):
LGTM

PR #24, opened 2020-11-23
Update device FW with fixes for RGB

alex-luxonis:
- fix crash with RGB 12MP + depth
- fix the cropping for 4K (make it centered, it was top-left)


saching13 (approved):
LGTM

PR #26, opened 2020-11-26
Adding SPI node for gen2

jonngai:


include/depthai/pipeline/node/SPIOut.hpp
```diff
@@ -0,0 +1,52 @@
...
+            return j;
+        }
+
+        std::shared_ptr<Node> clone(){
+            return std::make_shared<SPIOut>(*this);
+        }
+    
+    public:
+        SPIOut(const std::shared_ptr<PipelineImpl>& par, int64_t nodeId) : Node(par, nodeId) {
+        }
```
themarpe, line 37: Set default bus id, eg. 0

themarpe (changes requested):
Remove blobs from example (and uneeded jsons, etc...) to not bloat the repo.
Then squash the commits so that its removed from history (otherwise the repo will maintain 28MB extra size)

include/depthai/pipeline/node/SPIOut.hpp
jonngai: Set default to 0.

PR #30, opened 2020-12-22
Docs kickoff

VanDavv:
Bootstrap Gen2 API docs


themarpe (changes requested):
Thanks for this PR.
One extra thing - can you add documentation to README.md (or docs/README.md) on some basics on how to build and view these, etc...

PR #34, opened 2021-01-05
Gen2 common objdet

jonngai:


include/depthai/pipeline/node/CommonObjDet.hpp
```diff
@@ -0,0 +1,33 @@
+#pragma once
```
themarpe, line 1: Still used?

include/depthai/pipeline/node/DetectionNetwork.hpp
```diff
@@ -0,0 +1,55 @@
...
+    std::shared_ptr<Node> clone() override;
+
+    OpenVINO::Version networkOpenvinoVersion;
+
+   protected:
+    DetectionNetwork(const std::shared_ptr<PipelineImpl>& par, int64_t nodeId);
+    dai::DetectionNetworkProperties properties;
+
+   public:
+    Input input{*this, "in", Input::Type::SReceiver, {{DatatypeEnum::Buffer, true}}};
```
themarpe, line 30: Should only accept ImgFrame types (as input w&h are required, right?)
```
Input input{*this, "in", Input::Type::SReceiver, {{DatatypeEnum::ImgFrame, false}}};
```

include/depthai/pipeline/node/DetectionNetwork.hpp
```diff
@@ -0,0 +1,55 @@
...
+    Output out{*this, "out", Output::Type::MSender, {{DatatypeEnum::Buffer, false}}};
```
themarpe, line 31: Outputs ImgDetections type:
```
Output out{*this, "out", Output::Type::MSender, {{DatatypeEnum::ImgDetections, false}}};
```

themarpe (changes requested):
Above mentioned comments and please remove `standaloneYOLOExample`. Move to `depthai-experiments`

include/depthai/pipeline/node/CommonObjDet.hpp
jonngai: Oh, good call, I can remove all the old CommonObjDet stuff.

include/depthai/pipeline/node/DetectionNetwork.hpp
jonngai: The SPIOut node is expecting DatatypeEnum::Buffer for input though. Can I leave it as DatatypeEnum::Buffer or is there some other way around this?

include/depthai/pipeline/node/DetectionNetwork.hpp
jonngai: Changed.

include/depthai/pipeline/node/DetectionNetwork.hpp
themarpe: SPIOut is expecting Buffer or its descendants (think, derived objects / polymorphism), same as XLinkIn.

Therefore they can both accept any messages.

themarpe:
@jonngai One extra request - can you derive `DetectionNetwork` from `NeuralNetwork`?

jonngai:
> @jonngai One extra request - can you derive `DetectionNetwork` from `NeuralNetwork`?

Hey @themarpe , I set that up but still had to override a number of methods since DetectionNetwork has its own inputs, outputs and properties. Let me know if there's a better way of handling this or if this'll do for the time being. Thanks.

include/depthai/pipeline/node/DetectionNetwork.hpp
jonngai: Probably a noob question, but when using the inherited getOutputs and getInputs methods, won't they return the base class's input and out?

PR #35, opened 2021-01-05
Gen2 camera and API improvements

themarpe:
Related PR: https://github.com/luxonis/depthai-shared/pull/14
Some API changes in this PR. Mainly deprecated setCamId and switched terminology for all queue related things to 'blocking' instead of previous mixed 'overwrite'.

 - DEPTHAI_LEVEL environmental variable controls default device level if below level “WARN”
 - Renamed and switched queue to use “blocking” terminology instead of “overwrite” – Eg. getOutputQueue(name, maxSize, blocking) should now be ‘false’ to specify non-blocking queue
 - Added getAll and tryGetAll to DataOutputQueue which retrieve all messages currently in queue or block till first message in former case
 - Color & Mono camera – deprecated setCamId in favor for boardSocket (AUTO, RGB, LEFT, RIGHT)
 - Color camera – added capability to select video and still output
 - Color camera – added CameraControl input which can issue still shot (capture)
 - Modified examples to use setBoardSocket


src/pipeline/node/ColorCamera.cpp
```diff
@@ -59,24 +87,206 @@ void ColorCamera::setInterleaved(bool interleaved) {
...
+            break;
+
+        case ColorCameraProperties::SensorResolution::THE_12_MP:
+            return {4056, 3040};
+            break;
+
+        default:
+            break;
+    }
+}
```
alex-luxonis, line 231: There's a build warning here, no return at the end.

themarpe:
Updated C++ examples to accommodate for API changes

szabi-luxonis:
Can you add example for this PR?
Changing the crop region at runtime, similar to ImageManip.

themarpe:
Added example: `color_camera_control`
Uses 'WASD' controls to move the crop window and 'C' to capture a still image

PR #37, opened 2021-01-11
Color/MonoCamera: add API to configure image orientation

alex-luxonis:
Related PR: https://github.com/luxonis/depthai-shared/pull/15

### DIFF
--- a/src/pipeline/node/ColorCamera.cpp
+++ b/src/pipeline/node/ColorCamera.cpp
@@ -7,6 +7,7 @@ namespace node {
 
 ColorCamera::ColorCamera(const std::shared_ptr<PipelineImpl>& par, int64_t nodeId) : Node(par, nodeId) {
     properties.boardSocket = CameraBoardSocket::AUTO;
+    properties.imageOrientation = CameraImageOrientation::AUTO;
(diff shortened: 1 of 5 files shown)


src/pipeline/node/ColorCamera.cpp
```diff
@@ -72,6 +73,17 @@ int64_t ColorCamera::getCamId() const {
...
 }
 
+// Set camera image orientation
+void ColorCamera::setImageOrientation(CameraImageOrientation imageOrientation) {
+    properties.imageOrientation = imageOrientation;
+}
+
+// Get camera image orientation
+CameraImageOrientation ColorCamera::getImageOrientation() const {
+    // TODO: in case of AUTO, see if possible to return actual value determined by device?
```
themarpe, line 83: Right now unfortunately no - all of the node building happens before device is known.
That said, we could add device type (OAK-1, OAK-D, ....) as a part of pipeline constructor, which would in turn help us resolve some of these values. This still doesn't solve all of it, but might be a good addition down the road.

themarpe (approved):

szabi-luxonis (approved):

PR #38, opened 2021-01-12
ImageManip: rotated/arbitrary cropping, add camera controls

alex-luxonis:
- Extend ImageManipConfig with Warp related functionality:
    ```
    void setCropRotatedRect(RotatedRect rr, bool normalizedCoords = true);
    void setWarpTransformFourPoints(std::vector<Point2f> pt, bool normalizedCoords);
    void setWarpTransformMatrix3x3(std::vector<float> mat);
    void setWarpBorderReplicatePixels();
    void setWarpBorderFillColor(int red, int green, int blue);
    void setRotationDegrees(float deg);
    void setRotationRadians(float rad);
    ```
- Add ImageManipConfig API allowing to skip or reuse a frame (useful when ImageManip `inputImage` is connected directly to the camera, and it has to process a series of bounding-boxes, or none). Usable only with `setWaitForConfigInput(true)`
    ```
    void setReusePreviousImage(bool reuse);
    void setSkipCurrentImage(bool skip);
    ```
- Extended CameraControl with ISP / 3A commands: 
    https://github.com/luxonis/depthai-core/blob/58033d6/include/depthai/pipeline/datatype/CameraControl.hpp
    https://github.com/luxonis/depthai-shared/blob/7fd5976/include/depthai-shared/datatype/RawCameraControl.hpp
- ColorCamera / MonoCamera nodes: added `inputControl` input, that can take control commands during runtime. Applying a set of initial commands on the `Mono/ColorCamera->initialControl` is also possible, but only these are implemented for now (will be extended in future):
    ```
    void setAutoFocusMode(AutoFocusMode mode);
    void setManualFocus(uint8_t lensPosition);
    ````
- Deprecated ImageManip API that was common to ImageManipConfig, in favor of applying on `imageManip->initialConfig` (shared with runtime messages passed via `inputConfig`):
`imageManip->setCenterCrop(0.7f) -> imageManip->initialConfig.setCenterCrop(0.7f)`
- NeuralNetwork: added `void setNumInferenceThreads(int numThreads);` - could set more than 1, but usually not recommended more than 2. Was used to improve the performance on OCR demo (two networks: text detection and bounding-box text recognition -- the latter was configured with two threads).
- Added auto/manual exposure/focus controls to `color_camera_control_example.cpp`
- Added new example: `image_manip_warp_example.cpp`

Related PR: https://github.com/luxonis/depthai-shared/pull/16

### DIFF
--- a/src/pipeline/node/NeuralNetwork.cpp
+++ b/src/pipeline/node/NeuralNetwork.cpp
@@ -82,5 +82,9 @@ void NeuralNetwork::setNumInferenceThreads(int numThreads) {
     properties.numThreads = numThreads;
 }
 
+int NeuralNetwork::getNumInferenceThreads() {
+    return properties.numThreads;
+}
+
 }  // namespace node
 }  // namespace dai
--- a/src/pipeline/node/NeuralNetwork.cpp
+++ b/src/pipeline/node/NeuralNetwork.cpp
@@ -78,5 +78,9 @@ void NeuralNetwork::setNumPoolFrames(int numFrames) {
     properties.numFrames = numFrames;
 }
 
+void NeuralNetwork::setNumInferenceThreads(int numThreads) {
+    properties.numThreads = numThreads;
+}
+
 }  // namespace node
 }  // namespace dai
--- a/cmake/Depthai/DepthaiDeviceSideConfig.cmake
+++ b/cmake/Depthai/DepthaiDeviceSideConfig.cmake
@@ -2,7 +2,7 @@
 set(DEPTHAI_DEVICE_SIDE_MATURITY "snapshot")  
 
 # "full commit hash of device side binary"
-set(DEPTHAI_DEVICE_SIDE_COMMIT "cc6f54c05d5448c8de5c082608a2f80ca6474b13")
+set(DEPTHAI_DEVICE_SIDE_COMMIT "c1d766103180bf0e8893b07314ec3382e054f6e8")
 
 # "version if applicable"
 set(DEPTHAI_DEVICE_SIDE_VERSION "")
--- a/cmake/Depthai/DepthaiDeviceSideConfig.cmake
+++ b/cmake/Depthai/DepthaiDeviceSideConfig.cmake
@@ -2,7 +2,7 @@
 set(DEPTHAI_DEVICE_SIDE_MATURITY "snapshot")  
 
 # "full commit hash of device side binary"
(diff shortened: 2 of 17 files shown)


themarpe (changes requested):
Can you add a test/example of this as well?
And some additional helper functions like:
 - rotate (degrees / radians)
 - etc..

Luxonis-Brandon:
This is looking to have so much awesome functionality!  Thanks @alex-luxonis !  
When do you think will be ready to merge?  It looks like from the last commit you may be already preparing it for merge?

themarpe (changes requested):
For `ColorCamera` and `ImageManip` should we rather add setInitialConfig taking in an corresponding message? Or reference return `getInitialConfig`
This depends on whether all options can be configured initially. And I think that right now `ColorCamera` is missing some initial settings

themarpe (approved):

src/pipeline/node/NeuralNetwork.cpp
```diff
@@ -78,5 +78,9 @@ void NeuralNetwork::setNumPoolFrames(int numFrames) {
     properties.numFrames = numFrames;
 }
 
+void NeuralNetwork::setNumInferenceThreads(int numThreads) {
+    properties.numThreads = numThreads;
```
szabi-luxonis, line 82: Can you add a check and throw error if this is not in valid range? 1 or 2 should be the only valid values.

szabi-luxonis (changes requested):

src/pipeline/node/NeuralNetwork.cpp
themarpe: Also the 'AUTO' value (0 or -1, TBD). And a getter, for later depthai-gui usage

alex-luxonis: So discussed internally, and depending on the NN blob, if it only uses Shaves and no NCEs, it may make sense to run on more threads (e.g. we could have 12 Shaves available, and the blob needs 3, then could have up to 4 threads, while setting more would work, but may not be optimal).
So decided to add more complex checks on device side regarding this field, and issue a warning if the value is larger than optimal.

Added a getter for this field for now. Some other fields in NeuralNetwork properties also need getters, we'll add later.

alex-luxonis:
Updated PR description with a list of changes. Merging.

PR #41, opened 2021-01-18
Changed calibration file read and console information (gen1)

saching13:


src/device.cpp
```diff
@@ -747,42 +747,54 @@ std::shared_ptr<CNNHostPipeline> Device::create_pipeline(
...
+                std::cout << sz << std::endl;
+                
+                if (sz < homography_size) {
+                    if(version < 5 && config.board_config.store_to_eeprom){
+                        std::cerr << WARNING "Calibration file is outdated. Recalibration required before writing to EEPROM." << std::endl << "To continue using old calibration disable('-e') write to EEPROM.";
+                        abort();
+                    }
+                    else{
+                        std::cerr << WARNING "Calibration file size " << sz << ENDC " < smaller than expected, data ignored. May need to recalibrate\n";
+                    }
```
WhitherGo, line 781: I have always believed in meaningful error messages. If the calibration file is too small it is ignored. If the file is ignored, it will never be written as it has been presented. The user must either append the file with the missing data or generate a new complete calibration file through the recalibration process. I would suggest something like the following for the second sentence.  "Verify configuration file is complete and correct or recalibrate." 

This is my only comment or concern with commits from this file. 

Discord: @caltain
GitHub: WitherGo
IRL: Dan

saching13: Thanks alot for the feedback. This happens only if calibrated using old calibration. I will add that information at this point informing the user to update the calibration.

szabi-luxonis (approved):
LGTM

PR #42, opened 2021-01-19
Gen2 improvements

themarpe:
This PR adds a couple of improvements: (related PR: https://github.com/luxonis/depthai-shared/pull/17)
 - Queues are created from pipeline description and checks are performed if trying to access queues which don't correspond to pipeline
 - Added queue events, which enable to block on multiple queues at the same time
 - Added callback capability to DataOutputQueue
 - `Device` constructors have a better default - waits 3s for the device and as a last resort tries connecting to an already booted device (debugging case) - makes testing faster, removes the need to add delays.
 - Added functions to query devices system information
 - Added initial docstring documentation
 - Improved error messages and added more checks
 - Improved ImgFrame::getTimestamp by returning `time_point` (compared to synced `steady_clock`)
 - StreamPacketParser - added missing message types and modified build system to error out if missing in the future
 - Added `SystemInformation` message and `SystemLogger` node

### DIFF
--- a/src/pipeline/datatype/StreamPacketParser.cpp
+++ b/src/pipeline/datatype/StreamPacketParser.cpp
@@ -2,23 +2,33 @@
 
 // standard
 #include <memory>
+#include <sstream>
 
 // libraries
 #include <XLinkPublicDefines.h>
+#include <spdlog/spdlog.h>
 
 #include <nlohmann/json.hpp>
 
 // project
 #include "depthai/pipeline/datatype/ADatatype.hpp"
 #include "depthai/pipeline/datatype/Buffer.hpp"
+#include "depthai/pipeline/datatype/CameraControl.hpp"
+#include "depthai/pipeline/datatype/ImageManipConfig.hpp"
+#include "depthai/pipeline/datatype/ImgDetections.hpp"
 #include "depthai/pipeline/datatype/ImgFrame.hpp"
 #include "depthai/pipeline/datatype/NNData.hpp"
+#include "depthai/pipeline/datatype/SystemInformation.hpp"
 
 // shared
 #include "depthai-shared/datatype/DatatypeEnum.hpp"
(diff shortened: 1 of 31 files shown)


src/pipeline/datatype/StreamPacketParser.cpp
```diff
@@ -69,6 +86,22 @@ std::shared_ptr<RawBuffer> parsePacket(streamPacketDesc_t* packet) {
...
+            return parseDatatype<RawImgDetections>(jser, data);
+            break;
+
+        case DatatypeEnum::SystemInformation:
+            return parseDatatype<RawSystemInformation>(jser, data);
+            break;
+
         default:
             throw std::runtime_error("Bad packet, couldn't parse");
             break;
```
szabi-luxonis, line 107: Gives warning:`control reaches end of non-void function`
Add     `throw std::runtime_error("Bad packet, couldn't parse");` to suppress warning.

src/pipeline/datatype/StreamPacketParser.cpp
```diff
@@ -107,6 +139,22 @@ std::shared_ptr<ADatatype> parsePacketToADatatype(streamPacketDesc_t* packet) {
...
+            return std::make_shared<ImgDetections>(parseDatatype<RawImgDetections>(jser, data));
+            break;
+
+        case DatatypeEnum::SystemInformation:
+            return std::make_shared<SystemInformation>(parseDatatype<RawSystemInformation>(jser, data));
+            break;
+
         default:
             throw std::runtime_error("Bad packet, couldn't parse");
             break;
```
szabi-luxonis, line 160: Gives warning:control reaches end of non-void function
Add throw std::runtime_error("Bad packet, couldn't parse"); to suppress warning.

szabi-luxonis (changes requested):
Other than that LGTM.

PR #44, opened 2021-01-21
Updating firmware.

jonngai:


themarpe (changes requested):
Merge after FW is merged as well (and used the merged commit)

PR #52, opened 2021-02-10
Allow direct access to the node connections

kunaltyagi:
While writing a thin generic wrapper, I found it hard to perform a few operations without causing unnecessary state monitoring or memory allocations without this function

### DIFF
--- a/include/depthai/pipeline/Pipeline.hpp
+++ b/include/depthai/pipeline/Pipeline.hpp
@@ -56,8 +56,9 @@ class PipelineImpl {
     GlobalProperties globalProperties;
     // Optimized for adding, searching and removing connections
     std::unordered_map<Node::Id, std::shared_ptr<Node>> nodeMap;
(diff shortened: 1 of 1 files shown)


themarpe:
Idea - move `using NodeConnectionMap` to Pipeline so that function signature doesn't need to reference PipelineImpl

kunaltyagi:
If needed, we should re-define the type in `Pipeline`, but the core definition should stay in the impl class. Otherwise, a change in impl class will need a change in the wrapper class as well.

PS: Usually, the impl class is in a different header/library in order to reap the benefits of Pimpl idiom, so I'm a bit confused about the structure and my opinions might be ill informed in this situation

themarpe:
Yup, sounds good - lets do a `using` in `Pipeline` that points to `PipelineImpl::NodeConnectionMap`.

In this case hiding the implementation wasn't actually necessary and the overall idea was mostly a `shared_ptr` usage without having to deal with the actual `shared_ptr`. So its not actually the usual usage of Pimpl idiom, but borrows some concepts.

themarpe (approved):

PR #53, opened 2021-02-11
Gen2 OpenCV Support

themarpe:
Added support for cv::Mat types being requested from ImgFrame.
Support is optional and can also be completely disabled.

Additional target is now created, depthai::core::opencv, which links to depthai::core and adds the OpenCV functionality.

This enables/will enable better UX when dealing with ImgFrames coming/going to device.


src/opencv/ImgFrame.cpp
```diff
@@ -0,0 +1,105 @@
...
+            break;
+
+        case Type::YUV420p:
+        case Type::NV12:
+        case Type::NV21:
+            size = cv::Size(getWidth(), getHeight() * 3 / 2);
+            type = CV_8UC1;
+            break;
+
+        case Type::RAW8:
```
szabi-luxonis, line 32: RAW16, GRAY16 can also be added as CV_16UC1

szabi-luxonis (changes requested):

CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.2) # For Hunter
+cmake_minimum_required(VERSION 3.4) # For Hunter
```
szabi-luxonis, line 1: For what we need cmake 3.4 ? On slightly older system with the default cmake it won't work.

src/opencv/ImgFrame.cpp
```diff
@@ -0,0 +1,147 @@
...
+
+        case Type::NV12:
+            cv::cvtColor(frame, output, cv::ColorConversionCodes::COLOR_YUV2BGR_NV12);
+            break;
+
+        case Type::NV21:
+            cv::cvtColor(frame, output, cv::ColorConversionCodes::COLOR_YUV2BGR_NV21);
+            break;
+
+        case Type::RAW8:
```
szabi-luxonis, line 132: Should we throw error in case it's not possible, since it's not the desired effect? (Function not used properly)

CMakeLists.txt
themarpe: The change was for `IN_LIST` operator. And right now AFAIK a couple of our dependencies require higher anyway. I'd have to create a separate build with actual `cmake` of that version to verify correctness of this (some behavior can also change with newer versions)

src/opencv/ImgFrame.cpp
themarpe: For RAW8, RAW16, etc...?
Maybe the function name is ill formed, but I had in mind that the return of this function gives you frame that opencv can "atleast" display. So thats why I counted in grayscale, etc..

Should I rather rename the function?

PR #56, opened 2021-02-13
Reduce memory allocations by using the nodeMap

kunaltyagi:
- Add an accesor in `Pipeline` for the underlying node map
- Use the nodeMap acccessor, use early return to reduce indent


themarpe:
Thanks for the PR @kunaltyagi
Looks good - merging.

Luxonis-Brandon:
(And yes I second Martin's comments - thanks @kunaltyagi - much appreciated!)

PR #57, opened 2021-02-16
Openvino 2021.2 support

szabi-luxonis:
Added support for openvino 2021.2
On the FW side added additional checks in NN node if the blob is compatible with openvino version.
Added option to specify more than 1 NCE per inference thread. 

Related PR: https://github.com/luxonis/depthai-shared/pull/20

### DIFF
--- a/src/device/Device.cpp
+++ b/src/device/Device.cpp
@@ -1,7 +1,6 @@
 #include "depthai/device/Device.hpp"
 
 // std
-#include <fstream>
 #include <iostream>
 
 // shared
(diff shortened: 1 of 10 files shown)


src/device/Device.cpp
```diff
@@ -278,6 +279,10 @@ Device::~Device() {
 }
 
 void Device::init(const Pipeline& pipeline, bool embeddedMvcmd, bool usb2Mode, const std::string& pathToMvcmd) {
+    if(!pathToMvcmd.empty()) {
+        std::ifstream f(pathToMvcmd.c_str());
+        if(!f.good()) throw std::runtime_error("Error path doesn't exist. Note: Environment variables in path are not expanded. (E.g. '~', '$PATH').");
+    }
```
themarpe, line 285: Rather move into `XLinkConnection`s constructor.
Either in `bootAvailableDevice` or 
```
if(bootWithPath) {
    bootStatus = bootAvailableDevice(foundDeviceDesc, pathToMvcmd);
} else {
    bootStatus = bootAvailableDevice(foundDeviceDesc, mvcmd);
}
```

themarpe (changes requested):

PR #27, opened 2020-11-26
Change to relative submodule path

szabi-luxonis:
After switching from `develop` to `gen2` I got the following error : 
`fatal: not a git repository: shared/depthai-shared/../../../.git/modules/depthai-core/modules/shared/depthai-shared`

Deleting `shared` folder + switching to relative submodule path and running `git submodule sync` resolved the issue.

PR #58, opened 2021-02-17
Allow recursive-clone with forks

kunaltyagi:
Currently, forking private forks is an issue since the gitmodule path only moves 1 up the hierarchy.

This PR applies the change to main branch. If accepted, I'll create a similar PR for the gen2_develop branch

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "depthai-cpp/shared/depthai-shared"]
 	path = shared/depthai-shared
-	url = ../depthai-shared.git
+	url = ../../luxonis/depthai-shared.git


Luxonis-Brandon:
This seems like a great idea.  Thank you!  Bringing up to team and likely will see an answer tomorrow when the team is back online.

themarpe:
Seems like the solution to https://github.com/luxonis/depthai-core/pull/27#issuecomment-735962837 as well.
CC: @szabi-luxonis 

Any known drawbacks of this method @kunaltyagi?

kunaltyagi:
None on github (at least that I know of). On gitlab, it might be an issue since the hierarchies can be arbitrary deep there

szabi-luxonis (approved):
LGTM

PR #73, opened 2021-03-03
Jetson-compile-error-fix

saching13:
Added `position independent code` option to overcome compilation error on jetson.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -45,6 +45,7 @@ if(POLICY CMP0028)
 endif()
 
 # Set to export compile commands for tools like clang-tidy and format
+set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
 # Add module path


themarpe:
How was the compilation done to require that? By default, static libraries are build, if shared libraries are set, then PIC is enabled via toolchain.

saching13:
Building `depthai-python` crashed at linking with an error that it needs to be compiled with `-fpic` on jetson nano.

saching13:
```
[ 91%] Linking CXX shared library /home/sachin/Desktop/luxonis/ros_ws/devel_isolated/depthai_bridge/lib/libdepthai-core.so
/usr/bin/ld: /home/sachin/Desktop/luxonis/ros_ws/devel_isolated/depthai_bridge/lib/libdepthai-resources.a(lib.cpp.o): relocation R_X86_64_PC32 against symbol `_ZN4cmrc6detail9directoryD1Ev' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: final link failed: bad value
collect2: error: ld returned 1 exit status
make[2]: *** [depthai-core/CMakeFiles/depthai-core.dir/build.make:586: /home/sachin/Desktop/luxonis/ros_ws/devel_isolated/depthai_bridge/lib/libdepthai-core.so] Error 1
make[1]: *** [CMakeFiles/Makefile2:1532: depthai-core/CMakeFiles/depthai-core.dir/all] Error 2
make: *** [Makefile:141: all] Error 2
<== Failed to process package 'depthai_bridge': 
  Command '['/home/sachin/Desktop/luxonis/ros_ws/devel_isolated/depthai_ros_msgs/env.sh', 'make', '-j24', '-l24']' returned non-zero exit status 2.
```

This issue happens when I add in ros too. I had previously added `set (CMAKE_POSITION_INDEPENDENT_CODE ON)` in ros package's CMakeLists.txt.
@themarpe

themarpe:
@saching13 how are you building this?
The more appropriate fix is: https://github.com/vector-of-bool/cmrc/issues/17
But if you have a case where you are linking static depthai::core into some other shared library, then you'd have to resolve this for all statically build files with a `CMAKE_TOOLCHAIN_FILE` that sets `set(CMAKE_POSITION_INDEPENDENT_CODE ON)`, which in turn applies to all

If you are building shared library (using BUILD_SHARED_LIBS) then this should be the case already. If you are setting `CMAKE_TOOLCHAIN_FILE` already and don't have control over that, then I guess one way would be to perform the above mentioned fix, where the setting the PIC property depends on `BUILD_SHARED_LIBS` being set

Issue #71, opened 2021-03-02
several code warnings related to deprecated funcs, truncation, and unused vars


diablodale:
Build results in 10+ warnings of code style, deprecation, truncation, etc. These can likely be resolved easily by minor code changes, or hiding the warnings with pragmas.

### Setup

* Microsoft Windows [Version 10.0.19042.804]
* Visual Studio 2019 Community v16.8.6
* Current `develop` branch 126f5fcf64bb127c3f5c7f069410b3e2754d20d7

### Repro

1. clean git clone, configure, build

### Result

```
[proc] Executing command: "C:\Program Files\CMake\bin\cmake.exe" --build c:/njs/lux/depthai-core/build --config Debug --target all -- -j 15
[build] [3/43   2% :: 0.027] Generating depthai-resources resource loader
[build] [3/43   4% :: 0.846] Generating intermediate file for C:/njs/lux/depthai-core/build/resources/depthai-bootloader-0.0.11.cmd
[build] [3/43   6% :: 4.622] Generating intermediate file for C:/njs/lux/depthai-core/build/resources/depthai-device-fwp-5e4974e930f4ecbe1bf718463f6a5da664e68788.tar.xz
[build] [18/43   9% :: 7.446] Building CXX object CMakeFiles\depthai-core.dir\src\device\CallbackHandler.cpp.obj
[build] ..\src\device\CallbackHandler.cpp(40): warning C4101: 'ex': unreferenced local variable
[build] [19/43  11% :: 7.574] Building CXX object CMakeFiles\depthai-core.dir\src\pipeline\AssetManager.cpp.obj
[build] ..\src\pipeline\AssetManager.cpp(84): warning C4267: 'initializing': conversion from 'size_t' to 'uint32_t', possible loss of data
[build] ..\src\pipeline\AssetManager.cpp(93): warning C4267: 'argument': conversion from 'size_t' to 'uint32_t', possible loss of data
[build] [20/43  13% :: 7.579] Building CXX object CMakeFiles\depthai-core.dir\src\pipeline\node\XLinkIn.cpp.obj
[build] [21/43  16% :: 7.631] Building CXX object CMakeFiles\depthai-core.dir\src\pipeline\node\MyProducer.cpp.obj
[build] [22/43  18% :: 7.642] Building CXX object CMakeFiles\depthai-core.dir\src\pipeline\node\XLinkOut.cpp.obj
[...]

Issue #79, opened 2021-03-08
mutex violations in Device: causing race conditions and failures


diablodale:
Hi. While working on issue https://github.com/luxonis/depthai-core/issues/71, I found errant code in Device.cpp

https://github.com/luxonis/depthai-core/blob/9f5d0861c07fea580c652f435c7e1422473ae079/src/device/Device.cpp#L401

Notice there is no variable name for the lock. Therefore the lock is created and instantly destroyed. It does not protect the scope.
I have the fix (easy...just give the var a name) in my PR for issue #71. However, this fix then exposes failures elsewhere.

### Setup
* Using develop branch 9a0474522ba1c961b2ae087a4613b336dd8f648d
* Microsoft Windows [Version 10.0.19042.804]
* VS Studio 2019 v16.9.0  (however, the code is errant in all platforms/compilers)

### Repro

Do code review. See the variable has no name.
Then correct the code like...

```c++
std::unique_lock<std::mutex> lock(this->rpcMutex);
```

and then run all the unit tests

### Result

```
[proc] Executing command: ctest -j1 -C Debug -T test --output-on-failure
[ctest]    Site: DALE-XMG
[ctest]    Build name: Win32-ninja
[ctest] Test project C:/njs/depthai-core/build
[ctest]       Start  1: color_camera_node_test
[...]

PR #80, opened 2021-03-08
fix narrowing compile warnings; mutex lock

diablodale:
Partially fixes https://github.com/luxonis/depthai-core/issues/71

The PR does not include the unsafe function usage (e.g. strncpy) that @themarpe wrote they want to resolve. I see opportunities to move away from unsafe c-strings. Alternatives are string spans, zstring, std::string, std::array, etc. I highly recommend moving away from c-strings; buffer/string handling is one of the most common security vulnerabilities and I believe the current thinking is any performance gain from raw c-strings are no longer relevant. http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#slstr-string

The changes in this PR are usually the smallest changes needed to compile without warnings using this project's cmake. There is unsafe code and I do not address that in this PR. For example, in bspatch.c lines 129,134 -or- DeviceBootloader.cpp lines 101,102,107,... are warnings of narrowing integers (int64_t->unsigned int, size_t->uint32_t, etc.) This are all possible loss/corrupt of data, loss of precision, etc. The changes I made remove the compile warnings. However, the problems causing the warning still exist. It is still possible to overflow, wraparound, loose precision, etc. To fix the underlying problem requires work that cascades wider. I'm confident these underlying issues are fixable, but outside the scope of this PR without further discussion. This means that the code is often just as "unsafe" as it was before. If you want to fix all these, we should discuss enough examples so that I understand your project's philosophy and can apply wider fixes.

I did not add/use GSL. I think it might be useful in your project, but I chose to use other methods to fix for now.

I did not fix warnings in the hunter managed FP16 project.,

I have a question. I have not fixed the cmake warning in CMakeLists.txt line 30 `set(DEPTHAI_VERSION ${PROJECT_VERSION} PARENT_SCOPE)`. The warning is `CMake Warning (dev) at CMakeLists.txt:30 (set):Cannot set "DEPTHAI_VERSION": current scope has no parent.` My question is...what is the goal with that set(DEPTHAI_VERSION)? If it is for this project to declare its version, then this is not the cmake way to do it. Instead, creating a version config file is relatively easing using the helper functions https://cmake.org/cmake/help/latest/module/CMakePackageConfigHelpers.html and that version config file will set variables that can be read in a consistent way, uses in package version checks, etc. What is your project's intention here?

Device.cpp line 401. I added a variable name for the lock. This means...the previous code *did not* lock the mutex. Technically it locked it for only line 401 as the lock was created and destroyed on that one line. 🎯This looks like important critical path code so locking the mutex where it was never locked before is exposing bugs elsewhere as I document in issue #79.

I see a lot of int<->float conversions for framerate in VideoEncoder.cpp. Seems like something could be unified. For example: `VideoEncoder::setDefaultProfilePreset(float fps)`, calls `setFrameRate(int fps)`, which sets `float dai::VideoEncoderProperties::frameRate`. float->int->float has multiple layers of precision loss and unneeded cpu cycles.

### DIFF
--- a/include/depthai/pipeline/node/VideoEncoder.hpp
+++ b/include/depthai/pipeline/node/VideoEncoder.hpp
@@ -101,7 +101,7 @@ class VideoEncoder : public Node {
      * Sets expected frame rate
      * @param frameRate Frame rate in frames per second
      */
-    void setFrameRate(int frameRate);
+    void setFrameRate(float frameRate);
 
     /// Get rate control mode
     Properties::RateControlMode getRateControlMode() const;
@@ -123,7 +123,7 @@ class VideoEncoder : public Node {
     /// Get input height
     int getHeight() const;
     /// Get frame rate
-    int getFrameRate() const;
+    float getFrameRate() const;
 };
 
 }  // namespace node
--- a/src/bspatch/bspatch.c
+++ b/src/bspatch/bspatch.c
@@ -126,12 +126,12 @@ int bspatch_mem(const uint8_t* oldfile_bin, const int64_t oldfile_size, const ui
     uint8_t* p_decompressed_block_end[NUM_BLOCKS] = {NULL, NULL, NULL};
 
     for(int i = 0; i < NUM_BLOCKS; i++) {
-        unsigned int decompressed_size = newsize * 4;
+        unsigned int decompressed_size = (unsigned int)newsize * 4;
         p_decompressed_block_original[i] = malloc(decompressed_size);  // reserve enough memory
 
         int ret = 0;
         if((ret = BZ2_bzBuffToBuffDecompress(
-                (char*)p_decompressed_block_original[i], &decompressed_size, (char*)patchfile_bin + block_offset_bz2[i], block_size_bz2[i], 0, 0))
+                (char*)p_decompressed_block_original[i], &decompressed_size, (char*)patchfile_bin + block_offset_bz2[i], (unsigned int)block_size_bz2[i], 0, 0))
            != BZ_OK) {
             (void)ret;
             error = -1;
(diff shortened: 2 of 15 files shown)


themarpe:
Thanks for the PR.

The C-style strings are used only for compatibility with third party XLink library - which we decided not to modify too much for now.

 Most conversions from 64bit to 32bit arise from the fact that the target is a 32bit processor, so we settled with those datasizes.
 
 ```
 set(DEPTHAI_VERSION ${PROJECT_VERSION} PARENT_SCOPE)
```
Intention of the above is to present the version to the *parent* project (in our case, `depthai-python`), which adds it as a subdirectoy without the need to either parse the `CMakeLists.txt` or to install.

Thanks for catching that `unique_lock` issue. Lets hold off until further, to see the implications of it and to fix any subsequent issues that it triggers.

Can you modify the `setFrameRate` to take an float instead in the same PR? Just a mistake there.

Seems like `style` is failing. Did you run `cmake --build [dir] --target clangformat` to apply style?

diablodale:
I'll try and reconcile the framerates together.

Having difficulty with style. The logs in the CI don't tell me what failed. Just...bad. It would be great if the CI presented the output of the check so that I an told the specific problem, can learn from it, and can make specific changes. I'm on Windows and vscode, the setup of clangformat and such is foreign to that. So I'm trying to get something to tell me anything. So far, no luck.

Ehh, the add_subdirectory rather than the full config and version loading is a bit hacky but if it work for you...I'll leave it alone. It's your team's choice and I will generally stay far away from python. So the error in cmakefile.tx due to python I will not fix.

diablodale:
clangeformat updated...seems the tool, by default, doesn't tell of problems/errors...it just blindly makes changes. 🤷‍♀️

I made framerate round-trip a float.

diablodale:
squashed

diablodale:
@themarpe, can we merge in this PR? Its becoming out of sync due to delayed merge.

themarpe:
@diablodale thanks, merged.

Only exception was the mutex part - I've commented that out for now and marked it `TODO` until we figure out the Windows issue.

PR #85, opened 2021-03-13
Installation of dependencies and integration CI tests

themarpe:
EDITED: 
Added additional improvements: 
 - Build tests and examples
 - Capability to import by pointing to build directory instead of installing
 - Clangformat for tests and examples
 - Respecting `DEPTHAI_INSTALL` variable 

ORIGINAL:
Modified the installation of dependencies, fixed `depthai::opencv` target propagation and added CI to test various integration configurations:
 - Linux, MacOS & Windows
 - Installing library compiled as Shared or Static and in Debug or Release configuration
 - Test integration using `add_subdirectory` or `find_package` CMake calls
 - Compiles and tests running a sample program linked to either `depthai::core` or `depthai::opencv` target

Also addresses @saching13 latest push to `develop` regarding a partial fix of dependency search.


saching13:
LGTM. 
But just one question. is there a way to add path `depthai_DIR` to `CMAKE_PREFIX_PATH` or `MODULES` path before hand ?

themarpe:
> LGTM.
> But just one question. is there a way to add path `depthai_DIR` to `CMAKE_PREFIX_PATH` or `MODULES` path before hand ?

You mean instead of defining it using CMake `-D` to define it in script? Yes, Afaik that is possible, either using `CMAKE_PREFIX_PATH` or `depthai_DIR`. Eg: 
```
set(depthai_DIR [path]) # or set(CMAKE_PREFIX_PATH "[path]" ${CMAKE_PREFIX_PATH})
find_package(depthai ...)
```

saching13:
I know those ones. 
I meant like other packages we don't need to send that path at all. these are added to the system CMake's global registry I think. 
So wanted to see if we can add it to that registery.

themarpe:
Not aware of any global CMake registry. Only thing that CMake does is automatically checks some paths to search for this files, more information here: https://cmake.org/cmake/help/latest/command/find_package.html#search-procedure

That would mean that on Linux for instance, installing to eg: `/usr/local` will result in `find_package(depthai ...)` finding the dependency without setting any extra paths.

saching13:
https://gitlab.com/libeigen/eigen/-/blob/master/CMakeLists.txt#L634
This is one thing I found on that. 
And another way that I had seen was creating Find{ProjectName}.cmake and placing this in cmake/modules which finds the rest of the paths.

saching13:
But anyways I can use this. Thanks alot.

themarpe:
Interesting, didn't know about this aspect of CMake.
```
Changed in version 3.15: By default the export(PACKAGE) command does nothing (see policy CMP0090) 
because populating the user package registry has effects outside the source and build trees. 
Set the CMAKE_EXPORT_PACKAGE_REGISTRY variable to add build directories to the CMake user package registry.
```
It seems they are trying to push this off a bit.

I could create an export dependent on `CMAKE_EXPORT_PACKAGE_REGISTRY`, so if that is set an export will happen.
Thoughts?

saching13:
> I could create an export dependent on CMAKE_EXPORT_PACKAGE_REGISTRY, so if that is set an export will happen.
sounds good I think. Will give it a shot then ?

saching13:
> I could create an export dependent on CMAKE_EXPORT_PACKAGE_REGISTRY, so if that is set an export will happen.
sounds good I think. Will give it a shot then ?

szabi-luxonis (commented):
Can we also add building the examples to build process?

themarpe:
@saching13 - I added the "build directory export". Meaning that one can now set `depthai_DIR` (or `CMAKE_PREFIX_PATH`) to build directory as well and everything should work the same as if pointed to installation directory.

One thing that I didn't get to work is actually exporting to CMake User Registry. I tried a couple of things, but none of them worked (didn't create `~/.cmake` folder, nor my test project found `depthai` package, if maybe the user registry is located somewhere else that I didn't check)

I still left the `export(PACKAGE)` command under condition of `CMAKE_EXPORT_PACKAGE_REGISTRY` being set - can you give this a try on your end and maybe check how to properly use this?

themarpe:
@szabi-luxonis - I added both compilation of tests and examples to `build` matrix. As well as added OpenCV to build `depthai::opencv` target as well

saching13:
Thanks. I think we can merge this for now. I will try it out sometime and see if I can find something on that later.

saching13 (approved):

szabi-luxonis (approved):

PR #86, opened 2021-03-18
Gen2 spatial yolo example

szabi-luxonis:
Added spatial tiny yolo v3 and v4 examples.

### DIFF
--- a/examples/src/spatial_tiny_yolo_example.cpp
+++ b/examples/src/spatial_tiny_yolo_example.cpp
@@ -1,0 +1,209 @@
+#include <chrono>
+#include <cstdio>
+#include <iostream>
+
+#include "utility.hpp"
+
+// Inludes common necessary includes for development using depthai library
+#include "depthai/depthai.hpp"
+
+static const std::vector<std::string> labelMap = {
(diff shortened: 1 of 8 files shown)


Luxonis-Brandon:
Super excited for these thanks!

examples/src/spatial_tiny_yolo_example.cpp
```diff
@@ -0,0 +1,209 @@
...
+        auto currentTime = std::chrono::steady_clock::now();
+        auto elapsed = duration_cast<duration<float>>(currentTime - startTime);
+        if(elapsed > seconds(1)) {
+            fps = (float)counter / elapsed.count();
+            counter = 0;
+            startTime = currentTime;
+        }
+
+        if(imgFrame) {
+            frame = toMat(imgFrame->getData(), imgFrame->getWidth(), imgFrame->getHeight(), 3, 1);
```
themarpe, line 164: Prefer:
```
frame = imgFrame->getFrame();
```

themarpe (changes requested):

examples/src/spatial_tiny_yolo_example.cpp
szabi-luxonis: Added `getCvFrame` to all.

themarpe (approved):

PR #93, opened 2021-03-26
Openvino 2021.3 support

szabi-luxonis:
Added Openvino 2021.3 support.
Deprecated 2020.1, 2020.2

### DIFF
--- a/src/utility/Resources.cpp
+++ b/src/utility/Resources.cpp
@@ -3,6 +3,7 @@
 #include <array>
 #include <cassert>
 #include <condition_variable>
+#include <iostream>
 #include <thread>
 
 // libarchive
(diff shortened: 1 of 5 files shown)


src/utility/Resources.cpp
```diff
@@ -63,16 +70,18 @@ std::vector<std::uint8_t> Resources::getDeviceBinary(OpenVINO::Version version,
...
 
     switch(version) {
         case OpenVINO::VERSION_2020_1:
+            std::cerr << "OpenVino version 2020.1 is deprecated and will be removed in the next release!" << std::endl;
             depthaiBinary = resourceMap[DEPTHAI_CMD_OPENVINO_2020_1_PATH];
             depthaiUsb2Patch = resourceMap[DEPTHAI_CMD_OPENVINO_2020_1_USB2_PATCH_PATH];
             break;
 
         case OpenVINO::VERSION_2020_2:
+            std::cerr << "OpenVino version 2020.2 is deprecated and will be removed in the next release!" << std::endl;
```
themarpe, line 84: Should we rather use spdlog here? (warn)
Also maybe NN would be a better place to put this (how many times is this function called?)

themarpe (changes requested):

src/utility/Resources.cpp
szabi-luxonis: Changed to `warn`.
It's called only once, through `getDeviceFirmware`

PR #101, opened 2021-04-08
 Calibration read/write/load API

saching13:
Added 
- read/write api over xlink.
- Calibration Handler to load version 6 and 5 of EEPROM data along with manual setters and getters to get the camera intrinsic, extrinsic, distortion coeffs.

Next steps 
- Load EepromData Functionality (No flash)
- Stereo rectification getters
- Make stereo Node mode configurable by adding overloading rectification and other required stereo properties.


saching13:
> Make stereo Node mode configurable by adding overloading rectification and other required stereo properties.
On this will do it on a different PR after we finish this, or I can do it after merging this in stereo fixes PR of @alex-luxonis

src/device/CalibrationHandler.cpp
```diff
@@ -0,0 +1,534 @@
...
+    }
+    nlohmann::json jsonData = nlohmann::json::parse(jsonStream);
+    eepromData = jsonData;
+    // validateCameraArray();
+}
+
+CalibrationHandler::CalibrationHandler(std::string calibrationDataPath, std::string boardConfigPath) {
+    // std::ifstream jsonStream(boardConfigPath);
+    // nlohmann::json json_data = nlohmann::json::parse(jsonStream);
+    auto matrix_conv = [](std::vector<float>& src, int startIdx) {
```
themarpe, line 31: `matrixConv`

src/device/CalibrationHandler.cpp
```diff
@@ -0,0 +1,534 @@
...
+#include "depthai-shared/common/Point3f.hpp"
+#include "depthai/utility/matrixOps.hpp"
+#include "nlohmann/json.hpp"
+
+namespace dai {
+
+CalibrationHandler::CalibrationHandler(std::string eepromDataPath) {
+    std::ifstream jsonStream(eepromDataPath);
+    // TODO(sachin): Check if the file exists first.
+    if(!jsonStream.good() || jsonStream.bad()) {
```
themarpe, line 20: Doesn't `is_open` suffice?

src/device/CalibrationHandler.cpp
```diff
@@ -0,0 +1,534 @@
...
+        if(topLeftPixelId.y > eepromData.cameraData[cameraId].height || bottomRightPixelId.x > eepromData.cameraData[cameraId].width) {
+            std::runtime_error("Invalid Crop size. Crop width or height is more than the original resized height and width");
+        }
+    }
+
+    if(topLeftPixelId.x > bottomRightPixelId.x || topLeftPixelId.y < bottomRightPixelId.y) {
+        std::runtime_error("Invalid Crop ratio.");
+    }
+
+    std::cout << "Test here ------------" << std::endl;
```
themarpe, line 209: To remove

src/device/CalibrationHandler.cpp
```diff
@@ -0,0 +1,534 @@
...
+    for(auto row : intrinsicMatrix) {
+        for(auto val : row) std::cout << val << "  ";
+        std::cout << std::endl;
+    }
+    std::cout << topLeftPixelId.x << " - " << bottomRightPixelId.y << std::endl;
```
themarpe, line 214: To remove

themarpe (changes requested):
Initial review

examples/src/calibration_storer.cpp
```diff
@@ -0,0 +1,99 @@
+
```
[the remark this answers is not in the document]
saching13: I wil change it to calibration_flash actually to match with function namings

include/depthai/device/CalibrationHandler.hpp
```diff
@@ -0,0 +1,263 @@
...
+
+    /**
+     * @brief Set the Board Info object
+     *
+     * @param version - Sets the version of the Calibration data(Current version is 6)
+     * @param swapLeftRightCam - This bool swaps the connection of left and right camera
+     * @param boardName - Sets your board name.
+     * @param boardRev - set your board revision id.
+     */
+    void setBoardInfo(uint32_t version, bool swapLeftRightCam, std::string boardName, std::string boardRev);
```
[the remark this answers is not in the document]
saching13: Agree. Makes sense. Will remove it.
I was skeptical about adding it in the first place. And thought of deciding later.

src/device/CalibrationHandler.cpp
saching13: if the user copies a corrupted file then it can still be opened and not have acceptable data. With that in mind I hadded this check.

include/depthai/device/CalibrationHandler.hpp
```diff
@@ -0,0 +1,263 @@
...
+namespace dai {
+/**
+ * CalibrationHandler is an interface to read/load/write structured calibration and device data.
+ */
+class CalibrationHandler {
+   public:
+    CalibrationHandler() = default;
+
+    /**
+     * @brief Construct a new Calibration Handler object using the
```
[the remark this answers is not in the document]
saching13: Not sure. I know doxygen does.
Will check and sphinx fails will change it.

include/depthai/device/CalibrationHandler.hpp
```diff
@@ -0,0 +1,263 @@
...
+     * @param width - repesents the width of the image at which intrinsics are calculated.
+     * @param height - repesents the height of the image at which intrinsics are calculated.
+     *
+     * intrinsic matrix = \begin{bmatrix}
+     *                              f_x &  0  & c_x \\
+     *                              0   & f_y & c_y \\
+     *                              0   &  0  &  1
+     *                          \end{bmatrix}
+     */
+    void setCameraIntrinsics(CameraBoardSocket cameraId, std::vector<std::vector<float>> intrinsics, int width, int height);
```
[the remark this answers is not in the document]
saching13: will use Size2f instead

src/device/Device.cpp
```diff
@@ -870,6 +870,19 @@ float Device::getSystemInformationLoggingRate() {
     return client->call("getSystemInformationLoggingrate").as<float>();
 }
 
+bool Device::flashCalibration(CalibrationHandler calibrationDataHandler) {
+    int res = client->call("storeToEeprom", calibrationDataHandler.getEepromData());
+    if(res == 0)
+        return true;
+    else
+        return false;
```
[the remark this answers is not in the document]
saching13:
```
if(res == 0) return true;
else return false;
```
I had done this. formatting changed it back to next line

include/depthai/device/CalibrationHandler.hpp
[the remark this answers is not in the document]
themarpe: The reason for tuple is that we use it in other parts of codebase as well (VideoEncoder, ColorCamera, ...) and in bindings, where you can simply pass a tuple to the function call from bindings.

But you can also do a Size2f along that IMO

src/device/Device.cpp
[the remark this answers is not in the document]
themarpe: Yeah - a bit unfortunate from clang formatting perspective - I haven't found a setting yet which would prevent this till now.

Then go with above rather

include/depthai/device/Device.hpp
```diff
@@ -174,6 +174,13 @@ class Device {
      */
     LogLevel getLogLevel();
 
+    /**
+     * Get the Device Info object o the device which is currently running
+     *
+     * @return DeviceInfo - DeviceInfo of the current device in execution
+     */
+    DeviceInfo getCurrentDeviceInfo();
```
themarpe, line 182: Const and rename IMO - as the current context is implied from the object on which the function is called
```
DeviceInfo getDeviceInfo() const;
```

examples/CMakeLists.txt
```diff
@@ -67,8 +67,9 @@ dai_add_example(camera_video src/camera_video_example.cpp)
 # Mono camera video output example
 dai_add_example(mono_camera src/mono_camera_example.cpp)
 
-dai_add_example(calibration_storer src/calibration_storer.cpp)
-dai_add_example(calibration_version_5 src/calibration_storer_ver_5.cpp)
+dai_add_example(calibration_flash src/calibration_flash.cpp)
+dai_add_example(calibration_flash_version5 src/calibration_flash_v5.cpp)
+dai_add_example(calibration_load src/calibration_load.cpp)
```
themarpe, line 72: These should preferably be optional (or off by default) in the test suite though.
As otherwise everytime the tests are ran, the device will be reflashed.

themarpe: As well as other intrusive examples:
```
dai_add_example(bootloader_example src/bootloader_example.cpp)
...
```

include/depthai/device/CalibrationHandler.hpp
```diff
@@ -49,16 +50,48 @@ class CalibrationHandler {
...
+    /**
+     * @brief Get the Camera Intrinsics object
+     *
+     * @param cameraId - Uses the cameraId to identify which camera intrinsics to return
+     * @param destShape - resized width and height of the image for which intrinsics is requested.
+     * @param topLeftPixelId - (x, y) point represents the top left corner coordinates of the cropped image which is used to modify the intrinsics for the
+     * respective cropped image
+     * @param bottomRightPixelId - (x, y) point represents the bottom right corner coordinates of the cropped image which is used to modify the intrinsics for
+     * the respective cropped image
+     * @return std::vector<std::vector<float>> - repesents the 3x3 intrinsics matrix of the respective camera at the requested size and crop dimensions.
```
themarpe, line 73: Check that these are actually parsed correctly by Python side as well. I'd suggest just making it in same manner as other code is (also for sake of consistency).
```
@returns Represents the 3x3 intrinsic matrix ...
```

Example:
```
    /**
     * Add a callback for device logging. The callback will be called from a separate thread with the LogMessage being passed.
     *
     * @param callback - Callback to call whenever a log message arrives
     * @return Id which can be used to later remove the callback
     */
    int addLogCallback(std::function<void(LogMessage)> callback);
```

themarpe: ^ this applies to the whole PR

themarpe (changes requested):

include/depthai/device/CalibrationHandler.hpp
saching13: But C++ users need to know the type right. Where users don't want to use the `auto`.

themarpe: The type is in the function signature - not sure how doxygen displays that, but its still there.

saching13: Removed them

include/depthai/device/Device.hpp
```diff
@@ -384,6 +392,22 @@ class Device {
...
+     * @return Returns true on successful flash, false on failure
+     */
+    bool flashCalibration(CalibrationHandler calibrationDataHandler);
+
+    /**
+     * Fetches the EEPROM data from the device and loads it into CalibrationHandler object
+     *
+     * @return The CalibrationHandler object containing the calibration currently flashed on device EEPROM
+     */
+    CalibrationHandler getCalibration();
```
themarpe, line 409: I'd rename this to `readCalibration` instead - I think it's more of a counterpart to `flashCalibration`

include/depthai/pipeline/Pipeline.hpp
```diff
@@ -43,7 +44,7 @@ class PipelineImpl {
     std::vector<Node::Connection> getConnections() const;
     void link(const Node::Output& out, const Node::Input& in);
     void unlink(const Node::Output& out, const Node::Input& in);
-
+    void setCalibrationData(CalibrationHandler calib);
```
themarpe, line 47: Add `CalibrationHandler getCalibrationData() const`

include/depthai/utility/matrixOps.hpp
```diff
@@ -0,0 +1,153 @@
...
+ *
+ * @param firstMatrix - vector of vector of float of shape (m x n)
+ * @param secondMatrix - vector of vector of float of shape (n x p)
+ * @return std::vector<std::vector<float>> - product of firstMatrix * secondMatrix, a vector of vector of float of shape (m x p)
+ */
+std::vector<std::vector<float>> matMul(std::vector<std::vector<float>>& firstMatrix, std::vector<std::vector<float>>& secondMatrix) {
+    std::vector<std::vector<float>> res;
+
+    if(firstMatrix[0].size() != secondMatrix.size()) {
+        std::cerr << "Number of column of the first matrix should match with the number of rows of the second matrix " << std::endl;
```
themarpe, line 19: IMO - Throw instead (or spdlog - we don't use plain stdout/stderr in other cases)

include/depthai/utility/matrixOps.hpp
```diff
@@ -0,0 +1,153 @@
...
+            }
+        }
+    }
+}
+
+/* Recursive function for finding determinant of matrix.
+   n is current dimension of A[][]. */
+static float determinant(std::vector<std::vector<float>>& A, int n)
+
+{
```
themarpe, line 72: Was Clangformat applied in this file?

include/depthai/utility/matrixOps.hpp
```diff
@@ -0,0 +1,153 @@
+#pragma once
```
themarpe, line 1: As this header is not public - move to `src/utility` instead (See: BootloaderHelper.cpp, etc...)

src/device/CalibrationHandler.cpp
```diff
@@ -0,0 +1,589 @@
...
+        throw std::runtime_error("Calibration data file not found or corrupted");
+    }
+
+    if(file.is_open()) {
+        std::ifstream boardConfigStream(boardConfigPath);
+        if(!boardConfigStream.good() || boardConfigStream.bad()) {
+            throw std::runtime_error("BoardConfig file not found or corrupted");
+        }
+
+        if(boardConfigStream.is_open()) {
```
themarpe, line 62: Combine is_open with good || bad checks.
```
if(!file.is_open() || !file.good() || file.bad()){
    throw std::runtime_error("Calibration data file not found or corrupted");
}

// No need to do if again
std::ifstream boardConfigStream(boardConfigPath);

if(!boardConfigStream.is_open() || !boardConfigStream.good() || boardConfigStream.bad()) {
    throw std::runtime_error("BoardConfig file not found or corrupted");
}

// No need to do if again
nlohmann::json boardConfigData = nlohmann::json::parse(boardConfigStream);
...
```

src/device/CalibrationHandler.cpp
```diff
@@ -0,0 +1,589 @@
...
+
+        file.seekg(0, file.end);
+        unsigned fSize = file.tellg();
+        file.seekg(0, file.beg);
+
+        if(fSize != versionSize) {
+            throw std::runtime_error("The calib file version is less than version 5. which has been deprecated. Please Recalibrate with the new version.");
+        }
+
+        std::vector<float> calibration_buff(versionSize / sizeof(float));
```
themarpe, line 98: `calibrationBuff`

src/device/CalibrationHandler.cpp
```diff
@@ -0,0 +1,589 @@
...
+        file.read(reinterpret_cast<char*>(calibration_buff.data()), fSize);
```
themarpe, line 99: ^ 
To be on the safe side
```
std::vector<float> calibrationBuff(std::istreambuf_iterator<char>(file), {});
```

src/device/CalibrationHandler.cpp
```diff
@@ -0,0 +1,589 @@
...
+    }
+    return;
+}
+
+bool CalibrationHandler::eepromToJsonFile(std::string destPath) {
+    nlohmann::json j = eepromData;
+    std::ofstream ob(destPath);
+    ob << std::setw(4) << j << std::endl;
+    return true;
+}
```
themarpe, line 587:
```
void CalibrationHandler::eepromToJsonFile(const std::string& destPath) const;
```

themarpe (changes requested):
Remove unneeded commented code and address some left over TODOs.

Also not sure how sphinx renders:
```
@param cameraId - CameraId of the camera for which cameraModel Type is being updated.
```

The minus(-) between the parameter and description isn't used in other cases.

Same for @brief - usually not used.

Check https://developer.lsst.io/cpp/api-docs.html for reference

include/depthai/pipeline/Pipeline.hpp
saching13: Do we need to ?
User will be setting a `CalibrationHandler` object in the same script so would they be calling it back again from pipeline ?

themarpe: Might be needed in future, could be that a given pipeline is created somewhere else, etc...

saching13: got it. 
For scenarios where it is not available should I throw an error or return an empty constructor of CalibrationHandler ?

themarpe: If default constructed CalibrationHandler makes sense in terms of contents / being able to detect its empty somehow, ..., then I'd go in that direction.

include/depthai/utility/matrixOps.hpp
saching13: I did and it didn't fix the space. 
Fixed manually now

examples/src/calibration_flash.cpp
```diff
@@ -0,0 +1,98 @@
...
+    inOneD = {0, 0, 0};
+    calibData.setImuExtrinsics(dai::CameraBoardSocket::RGB, inMatrix, inOneD);
+
+    inMatrix = {{0.999954, -0.001489, -0.009451}, {0.001514, 0.999995, 0.002682}, {0.009447, -0.002697, 0.999952}};
+    calibData.setStereoLeft(dai::CameraBoardSocket::LEFT, inMatrix);
+
+    inMatrix = {{0.999763, -0.012779, -0.017639}, {0.012732, 0.999915, -0.002802}, {0.017673, 0.002577, 0.999840}};
+    calibData.setStereoRight(dai::CameraBoardSocket::RIGHT, inMatrix);
+    calibData.setFov(dai::CameraBoardSocket::RIGHT, 63.55);
+    calibData.setFov(dai::CameraBoardSocket::LEFT, 80.55);
```
pawel-pieniazek, line 70: why FOV of left and right camera are different?

saching13: Calibration Data in the example is made up to show as an example.
Writing an actual calibration script in python which calibrates and writes it to the device.

alex-luxonis:
The tests using parameters (NN blob path) are failing:
```
      Start  8: camera_mobilenet
 8/30 Test  #8: camera_mobilenet ....................***Failed    0.50 sec
statusfound -P
-- arguments: /home/alex/depthai-python/depthai-core/build/examples/camera_mobilenet;ON;
terminate called after throwing an instance of 'std::logic_error'
  what():  BlobReader error: File does not seem to be a supported neural network blob
-- After process executed,  produced the following exit code: Subprocess aborted
```
`ON` gets passed as an argument to the program, and it overrides the blob path.

szabi-luxonis:
> The tests using parameters (NN blob path) are failing:
> 
>       Start  8: camera_mobilenet
>  8/30 Test  #8: camera_mobilenet ....................***Failed    0.50 sec
> statusfound -P
> -- arguments: /home/alex/depthai-python/depthai-core/build/examples/camera_mobilenet;ON;
> terminate called after throwing an instance of 'std::logic_error'
>   what():  BlobReader error: File does not seem to be a supported neural network blob
> -- After process executed,  produced the following exit code: Subprocess aborted
>  Save
> ON gets passed as an argument to the program, and it overrides the blob path.
> 
> Yeah. I have seen that. Didn't realize ON was the reason. Weird that it overrides in test but not in actual example. Any idea why ?

Myabe it should be skipped https://github.com/luxonis/depthai-core/pull/101/files#diff-940e5356fa1137e646ea135a6f68cacbfad4fe7124c3a69163468f588acf9283R24 here

saching13:
> > The tests using parameters (NN blob path) are failing:
> > ```
> >   Start  8: camera_mobilenet
> > ```
> > 
> >  Save
> >   
> >     
> >     
> > 
> >     
> >     
> > 
> >   
> > 
> > 
> > 8/30 Test  #8: camera_mobilenet ....................***Failed    0.50 sec
> > statusfound -P
> > -- arguments: /home/alex/depthai-python/depthai-core/build/examples/camera_mobilenet;ON;
> > terminate called after throwing an instance of 'std::logic_error'
> > what():  BlobReader error: File does not seem to be a supported neural network blob
> > -- After process executed,  produced the following exit code: Subprocess aborted
> > Save
> > ON gets passed as an argument to the program, and it overrides the blob path.
> > Yeah. I have seen that. Didn't realize ON was the reason. Weird that it overrides in test but not in actual example. Any idea why ?
> 
> Myabe it should be skipped [here](https://github.com/luxonis/depthai-core/pull/101/files#diff-940e5356fa1137e646ea135a6f68cacbfad4fe7124c3a69163468f588acf9283R24)

Thanks. Fixed it.

themarpe (approved):
LGTM

examples/src/bootloader_example.cpp
```diff
@@ -0,0 +1,21 @@
+#include <string>
+
+#include "depthai/depthai.hpp"
+
+int main(int argc, char** argv) {
+    bool res = false;
+    dai::DeviceInfo info;
+    std::tie(res, info) = dai::DeviceBootloader::getFirstAvailableDevice();
+    std::string path;
+    if(argc = 2) {
```
alex-luxonis, line 10: Typo here? `=` -> `==`

szabi-luxonis (changes requested):
Examples are inconsistent with the python ones (there are examples that are not in depthai-python)

saching13:
> Examples are inconsistent with the python ones (there are examples that are not in depthai-python)

On this. Should I add calibration_load and calibration_flash example in C++ and python ? 
`Cacalibration_flash` would be done by calibration script so I was thinking of removing that one. 
And integrate `calibration_load` option inside `stereo` mode example. 

Thoughts ? @szabi-luxonis

szabi-luxonis:
> > Examples are inconsistent with the python ones (there are examples that are not in depthai-python)
> 
> On this. Should I add calibration_load and calibration_flash example in C++ and python ?
> `Cacalibration_flash` would be done by calibration script so I was thinking of removing that one.
> And integrate `calibration_load` option inside `stereo` mode example.
> 
> Thoughts ? @szabi-luxonis

Yes, better remove calib_flash because it can mess up calibration data if the user runs it accidentally.
Calibration load you could integrate in this experiment : https://github.com/luxonis/depthai-experiments/tree/master/gen2-camera-demo
since pcl requires it.

saching13:
> > > Examples are inconsistent with the python ones (there are examples that are not in depthai-python)
> > 
> > 
> > On this. Should I add calibration_load and calibration_flash example in C++ and python ?
> > `Cacalibration_flash` would be done by calibration script so I was thinking of removing that one.
> > And integrate `calibration_load` option inside `stereo` mode example.
> > Thoughts ? @szabi-luxonis
> 
> Yes, better remove calib_flash because it can mess up calibration data if the user runs it accidentally.
> Calibration load you could integrate in this experiment : https://github.com/luxonis/depthai-experiments/tree/master/gen2-camera-demo
> since pcl requires it, but also add a standalone example.

Yes. Adding the read in `gen2-camera-demo` is good idea. I already have local changes on that. 
But `calibration_load` example takes calibration data as input to the pipeline and when this is done calibration data in EEPROM will be overloaded with the calibration data in the pipeline. 
Examples for this falls under stereo setup as of now.

szabi-luxonis:
> > > > Examples are inconsistent with the python ones (there are examples that are not in depthai-python)
> > > 
> > > 
> > > On this. Should I add calibration_load and calibration_flash example in C++ and python ?
> > > `Cacalibration_flash` would be done by calibration script so I was thinking of removing that one.
> > > And integrate `calibration_load` option inside `stereo` mode example.
> > > Thoughts ? @szabi-luxonis
> > 
> > 
> > Yes, better remove calib_flash because it can mess up calibration data if the user runs it accidentally.
> > Calibration load you could integrate in this experiment : https://github.com/luxonis/depthai-experiments/tree/master/gen2-camera-demo
> > since pcl requires it, but also add a standalone example.
> 
> Yes. Adding the read in `gen2-camera-demo` is good idea. I already have local changes on that.
> But `calibration_load` example takes calibration data as input to the pipeline and when this is done calibration data in EEPROM will be overloaded with the calibration data in the pipeline.
> Examples for this falls under stereo setup as of now.

Then what's the difference between load and flash?

saching13:
> > > > > Examples are inconsistent with the python ones (there are examples that are not in depthai-python)
> 
> > > > 
> 
> > > > 
> 
> > > > On this. Should I add calibration_load and calibration_flash example in C++ and python ?
> 
> > > > `Cacalibration_flash` would be done by calibration script so I was thinking of removing that one.
> 
> > > > And integrate `calibration_load` option inside `stereo` mode example.
> 
> > > > Thoughts ? @szabi-luxonis
> 
> > > 
> 
> > > 
> 
> > > Yes, better remove calib_flash because it can mess up calibration data if the user runs it accidentally.
> 
> > > Calibration load you could integrate in this experiment : https://github.com/luxonis/depthai-experiments/tree/master/gen2-camera-demo
> 
> > > since pcl requires it, but also add a standalone example.
> 
> > 
> 
> > Yes. Adding the read in `gen2-camera-demo` is good idea. I already have local changes on that.
> 
> > But `calibration_load` example takes calibration data as input to the pipeline and when this is done calibration data in EEPROM will be overloaded with the calibration data in the pipeline.
> 
> > Examples for this falls under stereo setup as of now.
> 
> 
> 
> Then what's the difference between load and flash?

Flash writes to eeprom.
Load in pipeline loads calibration for that execution instance. This is replacement for loadCalibration() in stereo. 
And the reason for removing this from stereo is because it loads complete calibration which can be used by other nodes. So it's not specific to stereo node. 
In stereo node in an another PR we should add separate function to overload only rectification patameters and intrinsics if users are doing it from host or swapping camera modules and testing etc.

szabi-luxonis:
> In stereo node in an another PR we should add separate function to overload only rectification patameters and intrinsics if users are doing it from host or swapping camera modules and testing etc.

There is already `loadCalibrationData` in stereo node.

saching13:
> > In stereo node in an another PR we should add separate function to overload only rectification patameters and intrinsics if users are doing it from host or swapping camera modules and testing etc.
> 
> There is already `loadCalibrationData` in stereo node.

We decided to deprecate that. The reason being that calibration data is for the whole device and not specific to the stereo node. 
Going forward there will be other nodes and more features under ImageManip I think which needs this data. 
And as of stereo node it doesn't need the complete calibration. if the user is sending overloaded calibration for stereo there should be no need to send the complete structure of calibration data. Having separate `overloadLeftRectification(std::vector<std::vector<float>) / overloadRightRectification(std::vector<std::vector<float>)` makes it more intuitive for users to try and play with it.

examples/src/calibration_flash.cpp
```diff
@@ -0,0 +1,24 @@
+
+#include <cstdio>
+#include <iostream>
+#include <string>
+
+// Inludes common necessary includes for development using depthai library
+#include "depthai/depthai.hpp"
+
+int main(int argc, char** argv) {
+    std::string calib_pth(CALIB_PATH);
```
themarpe, line 10: calibPath

examples/src/calibration_load.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+    monoLeft->out.link(stereo->left);
+    monoRight->out.link(stereo->right);
+    stereo->depth.link(xoutDepth->input);
+
+    return p;
+}
+
+int main(int argc, char** argv) {
+    std::string calib_pth(CALIB_PATH);
```
themarpe, line 36: calibPath

include/depthai/device/Device.hpp
```diff
@@ -22,6 +22,7 @@
 #include "depthai-shared/common/MemoryInfo.hpp"
 #include "depthai-shared/log/LogLevel.hpp"
 #include "depthai-shared/log/LogMessage.hpp"
+#include "depthai/device/CalibrationHandler.hpp"
```
themarpe, line 25: Move under `//project` group of includes

src/device/Device.cpp
```diff
@@ -910,6 +914,19 @@ float Device::getSystemInformationLoggingRate() {
     return client->call("getSystemInformationLoggingrate").as<float>();
 }
 
+bool Device::flashCalibration(CalibrationHandler calibrationDataHandler) {
+    // std::unique_lock<std::mutex> lock(this->rpcMutex);
```
themarpe, line 918: Remove this line

themarpe (changes requested):

saching13 (commented):
@szabi-luxonis  and @themarpe Can you cross-check and approve it if all good?

alex-luxonis (approved):

saching13:
Cannot merge this. Need approval from @szabi-luxonis and @themarpe

alex-luxonis:
It appears this firmware fixes the slight shifting issue between RGB and depth, with RGB-depth alignment enabled. :+1:

Luxonis-Brandon:
Sweet!

PR #108, opened 2021-04-13
Video encoder fixes

szabi-luxonis:
While debugging rpi crash came across these issues.
Bitrate is set incorrectly, which results in 11MB encoded packets at 4k,


themarpe (approved):
Thanks for catching this!

Luxonis-Brandon:
Yes, thanks, both!

PR #110, opened 2021-04-14
Synchronize cpp examples with python

csaba-luxonis:
This PR is WIP, will be updated once it's ready to review.


Luxonis-Brandon:
Thanks for making these.  These will be great to have!  Will be so nice to have full Python and C++ examples for folks.

Luxonis-Brandon:
Oh and once these are done @csaba-luxonis - we should do the switcher on our docs so folks can switch between the Python code sample and the C++ code sample on the same page.

csaba-luxonis:
Yeah, that would be amazing.

themarpe:
I think we should do this in other direction:
Remove example numbers from python examples and leave these here as is.

Documentation should mark examples (code samples) with numbers but just as an easier overview

Luxonis-Brandon:
Oh, yes.  Sorry I didn't meant to say that we should have the numbers on these.  I agree we should remove the numbers from the Python examples.  And just have the documentation designate the order of the code samples.

examples/src/04_rgb_encoding.cpp
```diff
@@ -0,0 +1,68 @@
+
+#include <csignal>
+#include <cstdio>
+#include <iostream>
+
+#include "utility.hpp"
```
themarpe, line 6: Unneeded

examples/src/04_rgb_encoding.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+// Inludes common necessary includes for development using depthai library
+#include "depthai/depthai.hpp"
+
+    // Keyboard interrupt (Ctrl + C) detected
```
themarpe, line 11: Remove leading whitespace

examples/src/04_rgb_encoding.cpp
```diff
@@ -0,0 +1,68 @@
...
+    // Output queue will be used to get the encoded data from the output defined above
+    auto q = d.getOutputQueue("h265", 30, true);
+
+    // The .h265 file is a raw stream file (not playable yet)
+    auto videoFile = std::fstream(h265Path, std::ios::out | std::ios::binary);
+    std::cout << "Press Ctrl+C to stop encoding..." << std::endl;
+
+    while(alive) {
+        auto h264Packet = q->get<dai::ImgFrame>();
+        videoFile.write((char*)h264Packet->getData().data(), h264Packet->getData().size());
```
themarpe, line 60:
```
videoFile.write(static_cast<char*>(h264Packet->getData().data()), h264Packet->getData().size());
```

examples/src/04_rgb_encoding.cpp
```diff
@@ -0,0 +1,68 @@
...
+    }
+    videoFile.close();
```
themarpe, line 62: Unneeded - RAII should take care of it

examples/src/04_rgb_encoding.cpp
```diff
@@ -0,0 +1,68 @@
...
+    // Pipeline is defined, now we can connect to the device
+    dai::Device d(p);
+    // Start pipeline
+    d.startPipeline();
+
+    // Output queue will be used to get the encoded data from the output defined above
+    auto q = d.getOutputQueue("h265", 30, true);
+
+    // The .h265 file is a raw stream file (not playable yet)
+    auto videoFile = std::fstream(h265Path, std::ios::out | std::ios::binary);
```
themarpe, line 55:
```
auto videoFile = std::ofstream(h265Path, std::ios::binary);
```

examples/src/04_rgb_encoding.cpp
```diff
@@ -0,0 +1,68 @@
...
+#include <cstdio>
+#include <iostream>
+
+#include "utility.hpp"
+
+// Inludes common necessary includes for development using depthai library
+#include "depthai/depthai.hpp"
+
+    // Keyboard interrupt (Ctrl + C) detected
+static bool alive = true;
```
themarpe, line 12:
```
static std::atomic<bool> alive{true};
```

examples/src/04_rgb_encoding.cpp
```diff
@@ -0,0 +1,68 @@
...
+static void sigintHandler(int signum) {
+    alive = false;
+}
+
+int main(int argc, char** argv) {
+    using namespace std;
+    using namespace std::chrono;
+    signal(SIGINT, &sigintHandler);
```
themarpe, line 20: Maybe a small comment on this + remove the using std and std::chrono, we don't use it here actually
```
std::signal(SIGINT, &sigintHandler);
```

examples/src/04_rgb_encoding.cpp
```diff
@@ -0,0 +1,68 @@
+
+#include <csignal>
+#include <cstdio>
+#include <iostream>
+
+#include "utility.hpp"
+
+// Inludes common necessary includes for development using depthai library
```
themarpe, line 8: Typo: `Inludes` -> `Includes`

examples/src/04_rgb_encoding.cpp
```diff
@@ -0,0 +1,68 @@
+
```
themarpe, line 1: Extra line

themarpe (changes requested):

examples/src/04_rgb_encoding.cpp
csaba-luxonis: static_cast from 'unsigned char *' to 'char *' is not allowed

themarpe: Good catch, `reinterpret_cast<char*>` should do

I guess we can leave C-style cast in this case

themarpe:
Also one thing I was thinking, if we allow passing the output file as parameter (as well as the default), to be able to set it to `/dev/null` / `nul` for testing case. (Same applies to Python side)

themarpe (changes requested):
Throughout examples: use ImgFrame::Type alias instead of Raw*
```
manip->initialConfig.setFrameType(dai::ImgFrame::Type::BGR888p);
```

themarpe (dismissed):
LGTM apart from:

Style - either same line or with curly braces (in a couple of cases)
```
if(syncNN) {
    ...
} else {
    ...
}
```

szabi-luxonis (approved):
LGTM

PR #112, opened 2021-04-16
DataInputQueue nullptr check

themarpe:
This checks if a nullptr was passed to `send` method of DataInputQueue and which also applies to `None` being passed from Python bindings

### DIFF
--- a/src/device/DataQueue.cpp
+++ b/src/device/DataQueue.cpp
@@ -248,6 +248,7 @@ std::string DataInputQueue::getName() const {
 
 void DataInputQueue::send(const std::shared_ptr<RawBuffer>& rawMsg) {
     if(!running) throw std::runtime_error(exceptionMessage.c_str());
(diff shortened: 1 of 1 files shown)


szabi-luxonis (approved):
LGTM.
Is there a context where this issue came up (out of curiosity) or just general fix?

Luxonis-Brandon:
I was just about to ask the same question actually.  Mainly because I'm overly-optimistically-hoping that this is the fix to Raspberry Pi issue.  But I'm guessing this is likely not the resolution to that but rather something discovered in that hunt.

themarpe:
Unfortunately doesn't addresses any burning issue - was a tad stale from a couple of days ago as I was working on a custom script and sent `None` type instead of actual message (returned by a "void" function in Python). Programming mistake on my end but looked quite serious at first glance.

Luxonis-Brandon:
Thanks.  Makes sense.

PR #113, opened 2021-04-22
Capability to specify a custom device binary

themarpe:
Added an environmental variable option `DEPTHAI_DEVICE_BINARY` which can point to a custom device side binary.
This eases debugging with custom binaries

Also suppressed an warning on GCC WRT unused `int r` variable

### DIFF
--- a/src/utility/Resources.cpp
+++ b/src/utility/Resources.cpp
@@ -3,6 +3,7 @@
 #include <array>
 #include <cassert>
 #include <condition_variable>
+#include <fstream>
 #include <iostream>
 #include <thread>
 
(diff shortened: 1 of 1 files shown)


src/utility/Resources.cpp
```diff
@@ -299,6 +315,9 @@ Resources::Resources() {
             }
         }
         r = archive_read_free(a);  // Note 3
+        assert(r == ARCHIVE_OK);
```
szabi-luxonis, line 318: As a note: on release versions assert is disabled, so it can go through silently.

szabi-luxonis (approved):

src/utility/Resources.cpp
themarpe: I think in this case that's okay as this assert will only fail on malfunctioned build setup, compared to runtime

PR #115, opened 2021-04-24
fixed getting size of video/still when setting isp scaling

Erol444:


alex-luxonis:
Thanks for testing and fixing! LGTM, will merge.
I'll also add the FW fix for still capture not working with scaling enabled. The capture flag was not forwarded on that path.

PR #124, opened 2021-04-29
Device API improvements

themarpe:
Major changes:
 - Modified behavior of Device(pipeline) - this now starts the pipeline implicitly.
 - Added (back) startPipeline(pipeline) which can start a pipeline if device was booted with Device() or Device(openvinoVersion) constructor
 - Deprecated startPipeline() function
 - Added getConnectedCameras() function to be able to query connected cameras and start different pipeline without reconnecting to the device.
 
Minor changes:
 - Refactored all examples to remove deprecated startPipeline()
 - Fixed missing Node clone methods on some classes
 - WIP: Pipeline cloning - missing asset cloning.
 - Fixed sanitizers on examples when not set as testing

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -3,6 +3,7 @@ option(DEPTHAI_TEST_EXAMPLES "Test examples - examples will be ran as a part of 
 
 # Dependencies
 find_package(OpenCV REQUIRED)
+find_package(Sanitizers)
 
 # Create utility library
 add_library(utility src/utility.cpp)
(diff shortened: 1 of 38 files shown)


examples/CMakeLists.txt
```diff
@@ -13,13 +14,16 @@ macro(dai_add_example example_name example_src)
     add_executable("${example_name}" "${example_src}")
     target_link_libraries("${example_name}" PRIVATE utility depthai::opencv ${OpenCV_LIBS})
 
+    # Add sanitizers for example
+    add_sanitizers(${example_name})
```
szabi-luxonis, line 18: This is probably more suitable for "tests" instead "examples" since users might not want to enable it, but it's a nice addition, so let's keep it.

szabi-luxonis (approved):

examples/CMakeLists.txt
themarpe: Its actually necessary not for debugging solely, but to get it to compile. Linker has to be aware of `-fsanitize=address/undefined/...` flag to link correctly.

But its a noop in case sanitizers are disabled

PR #131, opened 2021-05-06
Camera custom tuning

alex-luxonis:
Add an option to override the default camera IQ (Image Quality) tuning database.
This is used for all cameras in the system, and specified at pipeline level:
`pipeline.setCameraTuningBlobPath("path/custom_tuning.bin")`

Related PR: https://github.com/luxonis/depthai-shared/pull/30

Internally it is stored as a pipeline asset (like an NN blob), and loaded over XLink, or flashed together with the pipeline.

Note: building such a tuning blob need to go through a request to Luxonis, as the tools required are not public.


src/pipeline/Pipeline.cpp
```diff
@@ -252,6 +253,24 @@ OpenVINO::Version PipelineImpl::getPipelineOpenVINOVersion() const {
     return openvinoVersion;
 }
 
+void PipelineImpl::setCameraTuningBlobPath(const std::string& path) {
+    std::string assetKey = "camTuning";
+
+    std::ifstream blobStream(path, std::ios::in | std::ios::binary);
```
szabi-luxonis, line 259: `std::ios::in` is redundant, that's why it's `ifstream` == input `fstream`
https://en.cppreference.com/w/cpp/io/basic_ifstream/basic_ifstream

themarpe (approved):
Looks good - thanks!

PR #132, opened 2021-05-11
stereo node getMaxDisparity()

Erol444:
added a function in stereo node that returns the maxDisparity. Also changed stereo_example.cpp to use this new function

### DIFF
--- a/include/depthai/pipeline/node/StereoDepth.hpp
+++ b/include/depthai/pipeline/node/StereoDepth.hpp
@@ -177,9 +177,9 @@ class StereoDepth : public Node {
 
     /**
      * Useful for normalization of the disparity map.
-     * @returns Number of disparity levels
+     * @returns Maximum disparity value that the node can return
      */
-    float getMaxDisparity();
+    float getMaxDisparity() const;
 };
 
 }  // namespace node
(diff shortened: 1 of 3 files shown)


themarpe (changes requested):
Also run clangformat (style check is failing):
`cmake --build build/ --target clangformat`

include/depthai/pipeline/node/StereoDepth.hpp
```diff
@@ -174,6 +174,12 @@ class StereoDepth : public Node {
      * DEPRECATED. The output is auto-enabled if used
      */
     [[deprecated("Function call should be removed")]] void setOutputDepth(bool enable);
+
+    /**
+     * Useful for normalization of the disparity map.
+     * @returns Number of disparity levels
+     */
+    float getMaxDisparity();
```
alex-luxonis, line 182: Any specific need for `float` instead of `int`?

And on the return description, the number of disparity levels would be `+1` more (also disparity value 0).

Erol444: So when you use it in code it doesn't matter if it's `255.0` or `255`. Great, thanks for noticing that:)

themarpe (approved):

PR #134, opened 2021-05-16
Added getUsbSpeed

themarpe:
Related: https://github.com/luxonis/depthai-shared/pull/31

Added capability to query current devices USB link speed

### DIFF
--- a/cmake/Depthai/DepthaiDeviceSideConfig.cmake
+++ b/cmake/Depthai/DepthaiDeviceSideConfig.cmake
@@ -2,7 +2,7 @@
 set(DEPTHAI_DEVICE_SIDE_MATURITY "snapshot")
 
 # "full commit hash of device side binary"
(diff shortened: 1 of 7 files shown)


alex-luxonis (commented):
Should we add a check in the Device() constructor about the actual speed, and issue an automated warning about USB2 detected?
Or leave it to the app to implement it.

And for the example to be consistent with the python one, add a snippet for calling `getUsbSpeed` as for `getConnectedCameras` here:
https://github.com/luxonis/depthai-core/blob/7d275ef/examples/src/camera_preview_example.cpp#L35-L39

themarpe:
Not sure yet - I think we should leave it up to users / examples. 
Some might not require full bandwidth, and USB2 might not be an issue.

alex-luxonis (approved):
Some formatting/style checks are failing.

PR #136, opened 2021-05-17
Overloaded setInputResolution to accept int tuple 

Erol444:
Overloaded setInputResolution to accept int tuple, just like `setPreviewSize`, `setVideoSize`, `setStillSize`, `setIspScale` etc.


include/depthai/pipeline/datatype/ImageManipConfig.hpp
```diff
@@ -40,6 +40,12 @@ class ImageManipConfig : public Buffer {
      */
     void setCropRect(float xmin, float ymin, float xmax, float ymax);
 
+    /**
+     * Specifies crop with rectangle with normalized values (0..1)
+     * @param coordinates Coordinate of rectangle
+     */
+    void setCropRect(std::vector<float> coordinates);
```
themarpe, line 47: I think a tuple is more suitable here:
`std::tuple<float, float, float, float>`

src/pipeline/datatype/ImageManipConfig.cpp
```diff
@@ -29,6 +29,10 @@ void ImageManipConfig::setCropRect(float xmin, float ymin, float xmax, float yma
     cfg.cropConfig.cropRect.ymax = ymax;
 }
 
+void ImageManipConfig::setCropRect(std::vector<float> coordinates) {
```
themarpe, line 32: I think a tuple is more suitable here:
`std::tuple<float, float, float, float>`

src/pipeline/datatype/ImageManipConfig.cpp
```diff
@@ -120,6 +128,10 @@ void ImageManipConfig::setResizeThumbnail(int w, int h, int bgRed, int bgGreen,
     cfg.resizeConfig.bgBlue = bgBlue;
 }
 
+void ImageManipConfig::setResizeThumbnail(std::tuple<int, int> size, int bgRed = 0, int bgGreen = 0, int bgBlue = 0) {
```
themarpe, line 131: Default parameter values are only specified in definition (.hpp)

themarpe (changes requested):

PR #138, opened 2021-05-21
Update FW with fix for random crashes (kernel crash on RPI/jetson)

szabi-luxonis:


src/device/Device.cpp
```diff
@@ -468,8 +468,7 @@ void Device::init2(bool embeddedMvcmd, bool usb2Mode, const std::string& pathToM
 
     client = std::unique_ptr<nanorpc::core::client<nanorpc::packer::nlohmann_msgpack>>(
         new nanorpc::core::client<nanorpc::packer::nlohmann_msgpack>([this](nanorpc::core::type::buffer request) {
-            // TODO(TheMarpe) - causes issues on Windows
-            // std::unique_lock<std::mutex> lock(this->rpcMutex);
+            std::unique_lock<std::mutex> lock(this->rpcMutex);
```
themarpe, line 471: This caused issues on Windows.
Was that issue resolved with your changes as well?

szabi-luxonis: How did the issue manifest? Based on my testings I didn't observe issues on windows.
RPC calls are serialized anyways, so for the 2nd caller the mutex will be blocked in XlinkWrite, until the first one receives answer in Xlinkread, since there is 1 worker thread to serve RPC requests on FW side.

szabi-luxonis: This doesn't belong to the fix though, I re-enabled while debugging.

szabi-luxonis: Reverted for now.

PR #146, opened 2021-06-02
Calibration data bug fix

saching13:
Changed double to float in set/get fov.
Modified Docstring for matrix (C++ only for now)
FW bug fix in stereo when `rgb` camera calibration was not available

### DIFF
--- a/include/depthai/device/CalibrationHandler.hpp
+++ b/include/depthai/device/CalibrationHandler.hpp
@@ -176,14 +176,15 @@ class CalibrationHandler {
      *                               0   &  0   &  0   & 1
      *                          \end{bmatrix}
      * 
-     *.. math::
-     *
-     *   M = \begin{bmatrix}
-     *           1 & 4 & 7 \\
-     *           2 & 5 & 8 \\
-     *           3 & 6 & 9
-     *       \end{bmatrix}
-     * 
+    r"""
+    .. math::
+
+        M = \begin{bmatrix}
+                1 & 4 & 7 \\
+                2 & 5 & 8 \\
+                3 & 6 & 9
+            \end{bmatrix}
+    """
      */
     std::vector<std::vector<float>> getImuToCameraExtrinsics(CameraBoardSocket cameraId, bool useSpecTranslation = false);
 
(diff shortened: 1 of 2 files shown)


alex-luxonis (approved):

include/depthai/device/CalibrationHandler.hpp
```diff
@@ -135,11 +151,9 @@ class CalibrationHandler {
...
      *
      * Matrix representation of transformation matrix
-     * transformation matrix = \begin{bmatrix}
-     *                              r_00 & r_01 & r_02 & T_x \\
-     *                              r_10 & r_11 & r_12 & T_y \\
-     *                              r_20 & r_21 & r_22 & T_z
-     *                          \end{bmatrix}
+     * \f[ \text{Transformation Matrix} = \left [ \begin{matrix}   r_{00} & r_{01} & r_{02} & T_x \\  r_{10} & r_{11} & r_{12} & T_y \\ r_{20} & r_{21} & r_{22}
+     * & T_z \\ 0   &  0   &  0   & 1\end{matrix} \right ] \f]
+     *
```
alex-luxonis, line 156: Aside from the extra additions, it wasn't rendered properly with the matrix rows on separate lines?
It would be slightly more readable directly from the text editor.
Anyway, not a blocker for the merge, we can see about it later.

saching13: it is rendering properly on C++ API but not on Python since mkdoc is removing `/` from `/f` for python. 
We should be able to fix it with slight changes on `pybind11_mkdoc`

alex-luxonis: Yes, saw it in the RTD temporary PR docs, it looks nice :+1: 
image

Was meaning just for keeping the rows separate here in .hpp, if possible.

saching13: Oh yeah. You are right. let me try that.

themarpe: @saching13 on Python side you can either:
 - Add parsing to pybind11_mkdoc (https://github.com/luxonis/pybind11_mkdoc) if easy doable
 - Or add custom docstring to this function in `depthai-python` (not using the `DOC(dai, ...)` but `"My docstring ..."`)

The later option is simpler but function descriptions won't be automatically synced anymore

PR #150, opened 2021-06-08
Separate Queue handling from core API

kunaltyagi:


src/device/Device.cpp
```diff
@@ -333,13 +328,25 @@ void Device::close() {
     // Clear map
     callbackIdMap.clear();
 
-    // Close connection first (so queues unblock)
-    connection->close();
-    connection = nullptr;
+    // Is this expected? Should we clear the queues before closing the device?
```
kunaltyagi, line 331: Question 1

src/device/Device.cpp
```diff
@@ -1055,4 +1018,53 @@ bool Device::startPipeline(const Pipeline& pipeline) {
     return true;
 }
 
+bool Device::startPipeline(const Pipeline& pipeline) {
+    bool success = DeviceBase::startPipeline(pipeline);
+    if(!success) return success;
+
+    // Open queues upfront, let queues know about data sizes (input queues)
```
kunaltyagi, line 1025: Question 2: this comment is no longer valid. Is the code still valid though?

Doing what the comment says requires breaking the startPipeline into 2 parts: pre and post (setupComm and startPipeline for example)

themarpe: What if we just move `DeviceBase::startPipeline(pipeline)` after opening the queues?

src/device/Device.cpp
themarpe: The connection is closed beforehand as queues block on waiting for data coming to XLink streams. 
This is a limitation of XLink as these blocking functions cannot be terminated gracefully.

themarpe (changes requested):
A couple of things are needed in this case:
 - `getConnection` to get the underlying `XLinkConnection` if working with streams manually is desired.
 - StreamPacketParser will have to be exposed & preferably cleaned up a bit (otherwise manual handling isn't possible)
 - I'd move DeviceBase to its own .hpp & .cpp

src/device/Device.cpp
kunaltyagi: Again, the issue is that `close` has multiple steps, and the queue handling is interleaved with the actual XLink setup logic. Any ideas on cleaning up this code? You'd be a better judge than me.

kunaltyagi (commented):
1. I made connection protected. I assume you'd prefer instead to add a function?
2. Would it be as simple as exposing the header properly (and relevant CMake changes)?
3. Sure. Will do that soon

src/device/Device.cpp
themarpe: We could do the same check twice, but I agree, lets move the check:
```
checkClosed();

    // first check if pipeline is not already running
    if(isPipelineRunning()) {
        throw std::runtime_error("Pipeline is already running");
    }
```
outside and create a private function for "starting" the pipeline (`startPipelineImpl` or something like that), which doesn't check.
Then we can check if pipeline is running, then create queues and at the end `startPipelineImpl`

For base we do just check and `startPipelineImpl`

src/device/Device.cpp
themarpe: Ohh, yeah, clearing them in the end should suffice. 
So callbacks must be removed first, device connection closed and then the queues can be cleared.

As is right now looks good to me on first glance

themarpe:
> 1. I made connection protected. I assume you'd prefer instead to add a function?

The connection being protected is great, but we also need to expose it as part of public API alongside being protected for internal use. 
Otherwise not much can be done with device without having its underlying XLinkConnection object. (The connection is needed for opening XLinkStreams)

> 2. Would it be as simple as exposing the header properly (and relevant CMake changes)?

Yes. And one extra thing that will have to be done eitherway are bindings, but we'll think about that after this PR is finalized.

> 3. Sure. Will do that soon

Thanks!

Luxonis-Brandon:
And CC: @saching13 on all the above.  Really cool.  Thanks, both!

kunaltyagi:
No modifications to startPipeline yet

src/device/Device.cpp
kunaltyagi: I've added another check in derived class. Is this also sufficient? I removed the extra `checkClosed` (since it's already part of `isPipelineRunning`)

include/depthai/pipeline/datatype/StreamPacketParser.hpp
```diff
@@ -6,10 +6,10 @@
...
 
-//project
+// project
 #include "depthai/pipeline/datatype/ADatatype.hpp"
 
-//shared
+// shared
 #include "depthai-shared/datatype/RawBuffer.hpp"
 
 // StreamPacket structure ->  || imgframepixels... , serialized_object, object_type, serialized_object_size ||
```
themarpe, line 15: This file being exposed now needs a tad more love:
 - class StreamPacketParser with static methods
 - serializeData -> serializeMessage
 - serializeMessage 4 overloads (shared_ptr and reference to RawBuffer & ADatatype)
 ```
std::vector<std::uint8_t> serializeMessage(const std::shared_ptr<RawBuffer>& rawMsg);
std::vector<std::uint8_t> serializeMessage(const RawBuffer& rawMsg);
std::vector<std::uint8_t> serializeMessage(const std::shared_ptr<ADatatype>& msg);
std::vector<std::uint8_t> serializeMessage(const ADatatype& msg);
```

For reference check `DataInputQueue::send` functions

**Review by themarpe — CHANGES_REQUESTED** [review]

File: include/depthai/device/DeviceBase.hpp [hunk 4]
```diff
@@ -0,0 +1,449 @@
...
+   private:
+    // private static
+    void init(OpenVINO::Version version, bool embeddedMvcmd, bool usb2Mode, const std::string& pathToMvcmd);
+    void init(const Pipeline& pipeline, bool embeddedMvcmd, bool usb2Mode, const std::string& pathToMvcmd);
+    void init2(bool embeddedMvcmd, bool usb2Mode, const std::string& pathToMvcmd, tl::optional<const Pipeline&> pipeline);
+
+    std::unique_ptr<nanorpc::core::client<nanorpc::packer::nlohmann_msgpack>> client;
+    std::mutex rpcMutex;
+    std::vector<uint8_t> patchedCmd;
+    std::shared_ptr<XLinkConnection> connection;
```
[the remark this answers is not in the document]
kunaltyagi: What's the benefit given that `getConnection` has been added? Brewity?

include/depthai/pipeline/datatype/StreamPacketParser.hpp
kunaltyagi: Should the API be `const std::shared_ptr<const T>& m` since there are `const T& m` overloads?

kunaltyagi: PS: RawBuffer::serialize is not const. Should it be made const?

kunaltyagi: Compilation error expected since waiting on ^^^ PR

src/device/Device.cpp
```diff
@@ -999,60 +268,6 @@ bool Device::startPipeline(const Pipeline& pipeline) {
...
-    std::tie(success, errorMsg) = client->call("buildPipeline").as<std::tuple<bool, std::string>>();
-    if(success) {
-        client->call("startPipeline");
-    } else {
-        throw std::runtime_error(errorMsg);
-        return false;
-    }
-
-    return true;
+    return DeviceBase::startPipeline(pipeline);
```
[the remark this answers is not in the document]
kunaltyagi: My mistake. Fixed it.

kunaltyagi:
PTAl @themarpe

include/depthai/device/DeviceBase.hpp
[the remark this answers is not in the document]
themarpe: The derived class `Device` (and others potentially) can use it as is, as it is the main thing that derived classes will need access to (especially with the need to change it, etc..).

src/device/Device.cpp
```diff
@@ -789,182 +222,18 @@ std::string Device::getQueueEvent(std::chrono::microseconds timeout) {
...
-    pipeline.serialize(schema, assets, assetStorage, pipelineOpenvinoVersion);
-    if(openvinoVersion != pipelineOpenvinoVersion) {
-        throw std::runtime_error("Device booted with different OpenVINO version that pipeline requires");
-    }
-
+bool Device::startPipelineImpl(const Pipeline& pipeline) {
     // Open queues upfront, let queues know about data sizes (input queues)
     // Go through Pipeline and check for 'XLinkIn' and 'XLinkOut' nodes
     // and create corresponding default queues for them
+    const auto& connection = getConnection();
```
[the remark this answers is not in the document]
themarpe: Also taking a const reference to a returned temporary object, not sure if this is valid

include/depthai/pipeline/datatype/StreamPacketParser.hpp
themarpe: > Should the API be const std::shared_ptr<const T>& m since there are const T& m overloads?
By `const T` do you mean a template? It could be as we wouldn't be exposing any "private" dependencies, although in its current form I think its okay as is - moves the actual implementation to the .cpp

include/depthai/pipeline/datatype/StreamPacketParser.hpp
```diff
@@ -0,0 +1,28 @@
...
+#include "depthai-shared/datatype/RawBuffer.hpp"
+
+// StreamPacket structure ->  || imgframepixels... , serialized_object, object_type, serialized_object_size ||
+// object_type -> DataType(int), serialized_object_size -> int
+
+namespace dai {
+class StreamPacketParser {
+   public:
+    static std::shared_ptr<RawBuffer> parsePacket(streamPacketDesc_t* packet);
+    static std::shared_ptr<ADatatype> parsePacketToADatatype(streamPacketDesc_t* packet);
```
themarpe, line 22: Thoughts on naming these `parseMessage` and `parseRawMessage` instead?
Would go better with the corresponding `serializeMessage` functions.

themarpe (changes requested):
Also added some additional responses to the comments.

Otherwise looking good, will give it a test locally after these finishing touches.

Thanks for your contribution!

src/device/Device.cpp
```diff
@@ -20,590 +16,27 @@
...
     for(const auto& kv : callbackIdMap) {
         outputQueueMap[kv.first]->removeCallback(kv.second);
     }
     // Clear map
     callbackIdMap.clear();
 
-    // Close connection first (so queues unblock)
-    connection->close();
-    connection = nullptr;
+    // Close the device before clearning the queues
```
themarpe, line 34: Typo: clearning -> clearing

themarpe (changes requested):

include/depthai/pipeline/datatype/StreamPacketParser.hpp
kunaltyagi: Sure, no issues

kunaltyagi: It's streamPacketParser, so I'm making it streamMessageParser

include/depthai/pipeline/datatype/ADatatype.hpp
```diff
@@ -11,12 +11,13 @@ namespace dai {
 class ADatatype {
    protected:
     friend class DataInputQueue;
-    virtual std::shared_ptr<dai::RawBuffer> serialize() const = 0;
+    friend class StreamPacketParser;
```
themarpe, line 14: `friend class StreamMessageParser;`

kunaltyagi:
```suggestion
    friend class StreamMessageParser;
```

kunaltyagi:
@themarpe PTAL

include/depthai/pipeline/datatype/NNData.hpp
```diff
@@ -14,7 +14,7 @@ namespace dai {
  */
 class NNData : public Buffer {
     static constexpr int DATA_ALIGNMENT = 64;
-    std::shared_ptr<RawBuffer> serialize() const;
+    std::shared_ptr<RawBuffer> serialize() const override;
```
themarpe, line 17: I presume this was meant to be done for all or none of the messages?

themarpe (changes requested):
Started merging but stumbled upon:
```
./depthai-core/examples/rgb_preview 
Connected cameras: 0 RGB 1 LEFT 2 RIGHT 
Usb speed: SUPER
terminate called after throwing an instance of 'std::runtime_error'
  what():  Queue for stream name 'rgb' doesn't exist
Aborted (core dumped)
```

Same with your branch unchanged. Please fix beforehand and merge in / rebase to latest develop

kunaltyagi:
Working on a rebase

include/depthai/pipeline/datatype/NNData.hpp
kunaltyagi: Other messages already had them. Just this one didn't have that

kunaltyagi:
PTAL. Due to huge amounts of code movement, please check the behavior a bit more carefully. Let's get this merged because dealing with merge conflicts is a big pain (due to code being moved across files)

kunaltyagi:
@themarpe

themarpe:
@kunaltyagi 
Hopefully I checked out the correct latest of your changes - but I'm still getting the error on simple `rgb_preview` example.
Please fix this beforehand and run the tests locally to verify (preferably with `TEST_TIMEOUT=10 make test`). In CMake configuration specify both `-D DEPTHAI_BUILD_EXAMPLES=ON` and `-D DEPTHAI_BUILD_TESTS=ON`, build and then test with previously mentioned command

kunaltyagi:
I don't actually know how to reslve the errors...

It's getting harder to rebase and compare. I think I'll start over if you can help fast-track the new PR before other code is merged in Device cpp and hpp files. It'd be easier to start over, but not something I'd do again and again
cc: @Luxonis-Brandon

themarpe:
@kunaltyagi I presume the error happens because Device doesn't open the queues upfront (the underlying startPipeline doesnt seem to do the right thing in Device case)

You can either fix in your branch, and I'll handle the merging with latest, or better if you fix after merging to latest in your own branch, to not reintroduce any issues.

By helping to fast track, what did you have in mind? I guess if you give me access to your branch I can help merging in latest develop and resolving conflicts?

kunaltyagi:
I already have allowed edits by maintainers
image

> error happens because Device doesn't open the queues upfront 

I see. Let me take a look at see how I can re-shuffle the code

src/device/Device.cpp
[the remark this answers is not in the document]
kunaltyagi: Taking a const reference to temporary extends the lifetime.

See the example in rvalue reference here: http://en.cppreference.com/w/cpp/language/reference

src/device/DeviceBase.cpp
```diff
@@ -0,0 +1,880 @@
...
+    PipelineSchema schema;
+    Assets assets;
+    std::vector<std::uint8_t> assetStorage;
+
+    // Mark the OpenVINO version and serialize the pipeline
+    OpenVINO::Version pipelineOpenvinoVersion;
+    pipeline.serialize(schema, assets, assetStorage, pipelineOpenvinoVersion);
+    if(openvinoVersion != pipelineOpenvinoVersion) {
+        throw std::runtime_error("Device booted with different OpenVINO version that pipeline requires");
+    }
```
kunaltyagi, line 824: These are the only lines which have been inverted in startPipeline logic. Queues are being opened upfront otherwise.

Current flow:
startPipeline -> Device::startPipelineImpl -> DeviceBase::startPipelineImpl

src/device/Device.cpp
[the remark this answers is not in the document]
themarpe: Thanks for the link - didn't know:)

Still, preferably expose using:

> Can be removed after DeviceBase::connection is protected

src/device/DeviceBase.cpp
themarpe: @kunaltyagi it seems that using this:
This is the stack trace:
```
dai::DeviceBase::startPipelineImpl(dai::Pipeline const&) (src/device/DeviceBase.cpp:815)
dai::DeviceBase::startPipeline(dai::Pipeline const&) (src/device/DeviceBase.cpp:811)
dai::DeviceBase::init2(bool, bool, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, tl::optional<dai::Pipeline const&>) (src/device/DeviceBase.cpp:638)
dai::DeviceBase::init(dai::Pipeline const&, bool, bool, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) (src/device/DeviceBase.cpp:388)
dai::DeviceBase::DeviceBase(dai::Pipeline const&) (src/device/DeviceBase.cpp:233)
dai::Device::Device(dai::Pipeline const&) (include/depthai/device/Device.hpp:26)
main (examples/src/rgb_preview.cpp:28)
__libc_start_main (@__libc_start_main:60)
_start (@_start:14)
```

With added constructor in Device.

I think the following happens: https://stackoverflow.com/a/962148

kunaltyagi:
Tested and worked.

I needed to add the following to CMakeLists.txt for examples and tests to get it working:
```cmake
set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -lrt")
```

Any ideas why? Should I create a PR for the same?

kunaltyagi:
> Any ideas why? Should I create a PR for the same?

Apparently, adding it to the depthai-core library also does the trick. If it's ok, I can create a PR for the same. (1 line: `target_link_options(${TARGET_CORE_NAME} INTERFACE "-lrt")`)

themarpe:
@kunaltyagi 
2. things:
 - DeviceBase constructors taking a pipeline don't start it now. I think a call to startPipeline should remain there, which will call `DeviceBase::startPipeline` (or can be explicit `DeviceBase::startPipeline` in ctor, just for it to be clear)
 - The now overloaded Device ctor needs a try catch around and do a close otherwise ctor will fail but device won't gracefully close.
 
Regarding the `rt` library - never happend yet on any system:/ Can you test if that is required in latest `develop` as well and if so can you share your system details?
I presume that if this might be related to "Threading" library that CMake knows to apply underneath, the correct place would be setting it in XLink as public: https://github.com/luxonis/XLink/blob/master/CMakeLists.txt#L24-L27

Would have to check which codebase actually pulls in the requirement for `rt`.

Otherwise this should be applied to Linux only afaik so should have a check with it and no need to do a flag, can just be:
```
target_link_libraries(target PUBLIC rt)
```

kunaltyagi:
I added a close for the second point.

What is the actionable item for the first point? There's no need for DeviceBase to call `startPipeline`, right?

The `Device` class takes care of that from the user perspective, and the developers of derived classes of `DeviceBase` can device what kind of behavior is preferred for them.

themarpe:
I think it just mixes with current constructors - these constructors actually shouldn't call DeviceBase(pipeline, ...) but regular non pipeline starting constructors. 

In Device constructors, there are 2 major ones - Device(pipeline, ...) and Device(openvino, ...).
The first one always starts the pipeline by itself and no subsequent `startPipeline` is required (or possible), while the second doesn't start a pipeline but just boots the correct FW (specified by openvino) and then waits for any other RPC calls to be made and/or pipeline started.

If we make DeviceBase not handle queues automatically, but still expose same API for constructors, IMO same logic should apply. So DeviceBase(pipeline, ...) starts the pipeline by itself, while DeviceBase(openvino, ...) doesn't.

kunaltyagi:
Thoughts on removing the `DeviceBase(pipeline)` ctors?
* DeviceBase doesn't have enough information to start pipeline safely (due to your arguments combined with the virtual function issue)
* This can be a feature for the `Device` class solely: start pipeline because it can handle queues

themarpe:
The second option is compelling, although for the first one it can indeed start the pipeline normally - its just going to call it's own DeviceBase::startPipelineImpl, resulting in started pipeline without queues opened upfront.

Maybe best to just modify the current Device ctors to call `Device(pipeline,...) : DeviceBase(pipeline.getOpenVINOVersion(), ...) {}`
And for DeviceBase adding back the call but explicitly to its own impl.

kunaltyagi:
> its just going to call it's own DeviceBase::startPipelineImpl, resulting in started pipeline without queues opened upfront.

I agree with the second portion: `Device(pipeline)` forwards to `Devicebase(openVinoVersion)`, and I still feel that adding `DeviceBase(pipeline)` creates a more confusing (and faulty: `starting pipeline without queues opened upfront`) API

themarpe:
> and I still feel that adding DeviceBase(pipeline) creates a more confusing (and faulty: starting pipeline without queues opened upfront) API

This actually won't happen - when a users uses DeviceBase constructor, they create a DeviceBase object, on which they won't be able to call `getOutputQueue`. I'd still rather retain the mostly the same behavior between DeviceBase and Device with exception of queues (at least in current of debate context and path of the PR)

kunaltyagi:
How does this look?

themarpe:
@kunaltyagi Thanks, looking good.
Will clone locally and update with latest and merge. Most likely tomorrowish.

kunaltyagi:
Just a ping :)

themarpe:
@kunaltyagi updated with latest develop. Hopefully haven't missed anything. Also created a branch on depthai-python with bindings, but have to add StreamMessageParser to it as well.

I'm still observing issues running the examples/tests. Specifically the `image_manip_node_test`. AFAIK `close` is suffering from same issues as opening the pipeline did. Its called from destructor where after `Device` is destructed, `DeviceBase` doesn't call its `Device::closeImpl` but just its own. (Construction/destruction: `DeviceBase->Device ... Device->DeviceBase` <- issue)

kunaltyagi:
FYI: The windows CI failed before I pushed

I modified the dtor in a similar fashion and added a note for developers. The test runs without any issue now.

kunaltyagi:
Errors on Windows CI:
```
D:\a\depthai-core\depthai-core\tests\src\image_manip_node_test.cpp(43,40): error C2668: 'dai::Device::Device': ambiguous call to overloaded function [D:\a\depthai-core\depthai-core\build\tests\image_manip_node_test.vcxproj]
D:\a\depthai-core\depthai-core\tests\src\color_camera_node_test.cpp(41,40): error C2668: 'dai::Device::Device': ambiguous call to overloaded function [D:\a\depthai-core\depthai-core\build\tests\color_camera_node_test.vcxproj]
D:\a\depthai-core\depthai-core\include\depthai\device/Device.hpp(59,5): message : could be 'dai::Device::Device(const dai::Pipeline &,const dai::DeviceInfo &,bool)' [D:\a\depthai-core\depthai-core\build\tests\image_manip_node_test.vcxproj]
D:\a\depthai-core\depthai-core\include\depthai\device/Device.hpp(59,5): message : could be 'dai::Device::Device(const dai::Pipeline &,const dai::DeviceInfo &,bool)' [D:\a\depthai-core\depthai-core\build\tests\color_camera_node_test.vcxproj]
D:\a\depthai-core\depthai-core\tests\src\image_manip_node_test.cpp(43,40): message : while trying to match the argument list '(dai::Pipeline, dai::DeviceInfo)' [D:\a\depthai-core\depthai-core\build\tests\image_manip_node_test.vcxproj]
D:\a\depthai-core\depthai-core\tests\src\color_camera_node_test.cpp(41,40): message : while trying to match the argument list '(dai::Pipeline, dai::DeviceInfo)' [D:\a\depthai-core\depthai-core\build\tests\color_camera_node_test.vcxproj]
D:\a\depthai-core\depthai-core\tests\src\image_manip_node_test.cpp(43,40): message : note: qualification adjustment (const/volatile) may be causing the ambiguity [D:\a\depthai-core\depthai-core\build\tests\image_manip_node_test.vcxproj]
D:\a\depthai-core\depthai-core\tests\src\color_camera_node_test.cpp(41,40): message : note: qualification adjustment (const/volatile) may be causing the ambiguity [D:\a\depthai-core\depthai-core\build\tests\color_camera_node_test.vcxproj]
```

Missing default ctor in `Device` (definitely missing, required for calibration examples)

themarpe:
Seems to work now as expected - MSVC was nitpicking, removed default arguments.
Will be merging after a release.

kunaltyagi:
The default ctor errors have remained

kunaltyagi:
🎊 yay 🎉

kunaltyagi:
@themarpe Will this be included in the next release?

themarpe:
@kunaltyagi yes, I plan to do so.
Currently working on some bootloader changes, but after I move back to core, will merge this in. 
Might come to develop by the end of this week

PR #151, opened 2021-06-09
Add support for median filter for LR check depth mode

szabi-luxonis:
Other filters are WIP


szabi-luxonis:
Included in: https://github.com/luxonis/depthai-core/pull/159

PR #153, opened 2021-06-11
Add convenience headers in depthai/pipeline

kunaltyagi:


szabi-luxonis:
Please target develop w/ PRs.

themarpe:
Does this enable anything outside of this refactor? 
If not its IMO unnecessary, as it just moves it to a separate include.
Otherwise I'd lean more on lowercase naming.

themarpe:
I think it could come handy in `StreamPacketParser` from #150 . 
Then I'd go with lowercase: nodes.hpp & datatypes.hpp and we can reuse in #150 as well (`StreamPacketParser.cpp`)

themarpe (approved):

PR #154, opened 2021-06-11
Make `getParentPipeline` publically available

kunaltyagi:
In absence of this, and node based queries need the pipline as an argument as well.

Moreover, there's a need for an extra check for confiming if the node actually belongs to the pipeline provided as the argument


themarpe (approved):

themarpe:
@kunaltyagi can you make a corresponding PR for `depthai-python` as well? (Adding the bindings for it)

Issue #155, opened 2021-06-11
getImuToCameraExtrinsics Error


PRono8:
Hi,

Thank you for your work. 
I have seen in the documentation the getImuToCameraExtrinsics and getCameraToImuExtrinsics functions. I have tried to use them in a similar way to the getCameraExtrinsics functions in the example code https://github.com/luxonis/depthai-core/blob/main/examples/src/calibration_reader.cpp. But I get the following error: segmentation fault ('Core' dump).

If someone could help me I would really appreciate it. Thank you very much
Cheers,
Jose

Luxonis-Brandon:
Hi @PRono8 ,

So @saching13 can correct me if I'm wrong, but my understanding is that all of our existing product do not have IMU to camera calibration performed - and this is why these are not working.  

That said, they shouldn't seg fault, so CC: @saching13 on that.

Sachin is working to write the routines now that will allow calibrating the IMU to the cameras.  And so these functions were actually architected in there as part of that effort - but we currently do not have the capability (to my knowledge) to calibrate the IMU to the cameras.

@saching13 is wrapping up some other things with calibration, and when that is done, he's moving on to IMU calibration, which will involve Vicalib Integration.

Thanks,
Brandon

saching13:
Hi @PRono8  ,
Like Brandon mentioned our devices don't ship with IMU calibration. Will be adding that feature soon and provide a code to calibrate that by the users. 

And sorry for Seg Fault. I will check on that and replace it with an appropriate error.

chobitsfan:
Hi @Luxonis-Brandon and @saching13 
I am geting "IMU calibration data is not available on device yet." from getImuToCameraExtrinsics in a recently bought OAK-D Lite. Are you sitll working on this? Thank you very much

PR #156, opened 2021-06-11
Fixed Seg Fault in getImuToCameraExtrinsics

saching13:
Addresses #155

### DIFF
--- a/src/device/CalibrationHandler.cpp
+++ b/src/device/CalibrationHandler.cpp
@@ -266,7 +266,6 @@ uint8_t CalibrationHandler::getLensPosition(CameraBoardSocket cameraId) {
     return eepromData.cameraData[cameraId].lensPosition;
 }
 
-// TODO(sachin) : Add a loop checker to make sure lin found doesnt go into infinite loop
 std::vector<std::vector<float>> CalibrationHandler::getCameraExtrinsics(CameraBoardSocket srcCamera, CameraBoardSocket dstCamera, bool useSpecTranslation) {
     /**
      * 1. Check if both camera ID exists.
@@ -339,7 +338,7 @@ std::vector<std::vector<float>> CalibrationHandler::getImuToCameraExtrinsics(Cam
     if(eepromData.imuExtrinsics.rotationMatrix.size() == 0 || eepromData.imuExtrinsics.toCameraSocket == CameraBoardSocket::AUTO) {
         throw std::runtime_error("IMU calibration data is not available on device yet.");
     } else if(eepromData.cameraData.find(cameraId) == eepromData.cameraData.end()) {
-        throw std::runtime_error("There is no Camera data available corresponding to the the requested source cameraId");
+        throw std::runtime_error("There is no Camera data available corresponding to the requested source cameraId");
     }
 
     std::vector<std::vector<float>> transformationMatrix = eepromData.imuExtrinsics.rotationMatrix;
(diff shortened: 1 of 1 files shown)


src/device/CalibrationHandler.cpp
```diff
@@ -336,12 +336,14 @@ std::vector<std::vector<float>> CalibrationHandler::getCameraToImuExtrinsics(Cam
 }
 
 std::vector<std::vector<float>> CalibrationHandler::getImuToCameraExtrinsics(CameraBoardSocket cameraId, bool useSpecTranslation) {
-    if(eepromData.cameraData.find(cameraId) == eepromData.cameraData.end()) {
+    if(eepromData.imuExtrinsics.rotationMatrix.size() == 0 || eepromData.imuExtrinsics.toCameraSocket == CameraBoardSocket::AUTO) {
+        throw std::runtime_error("IMU calibration data is not available on device yet!!!");
```
themarpe, line 340: `IMU calibration data is not available on device yet.`

themarpe (changes requested):
I presume segfault happened becasue of line 347, accessing tranformationMatrix which is empty?

If thats the case, please place a check there as well, could be empty in different cases. 

Same applies to other places

saching13:
> I presume segfault happened becasue of line 347, accessing tranformationMatrix which is empty?
> 
> If thats the case, please place a check there as well, could be empty in different cases.
> 
> Same applies to other places

`std::vector<std::vector<float>> transformationMatrix = eepromData.imuExtrinsics.rotationMatrix;`
The transformation matrix is a copy of the rotation matrix. So it will be empty only when rotation matrix is empty. And rotation matrix will be either empty or have 3x3 always.

src/device/CalibrationHandler.cpp
```diff
@@ -336,12 +336,14 @@ std::vector<std::vector<float>> CalibrationHandler::getCameraToImuExtrinsics(Cam
...
 
 std::vector<std::vector<float>> CalibrationHandler::getImuToCameraExtrinsics(CameraBoardSocket cameraId, bool useSpecTranslation) {
-    if(eepromData.cameraData.find(cameraId) == eepromData.cameraData.end()) {
+    if(eepromData.imuExtrinsics.rotationMatrix.size() == 0 || eepromData.imuExtrinsics.toCameraSocket == CameraBoardSocket::AUTO) {
+        throw std::runtime_error("IMU calibration data is not available on device yet.");
+    } else if(eepromData.cameraData.find(cameraId) == eepromData.cameraData.end()) {
         throw std::runtime_error("There is no Camera data available corresponding to the the requested source cameraId");
     }
+
     std::vector<std::vector<float>> transformationMatrix = eepromData.imuExtrinsics.rotationMatrix;
```
themarpe, line 345: Is this line guarenteed to not be empty? No check required? If it can be then vector should be resized first

themarpe (changes requested):

src/device/CalibrationHandler.cpp
saching13: ` if(eepromData.imuExtrinsics.rotationMatrix.size() == 0 || eepromData.imuExtrinsics.toCameraSocket == CameraBoardSocket::AUTO) {       
 throw std::runtime_error("IMU calibration data is not available on device yet.");   
 }`

Since this above section throws an error  when `eepromData.imuExtrinsics.rotationMatrix.size()` is zero the code never reaches the part where it is assigned to the `transformationMatrix`. So I thought we don't need to add an another check for that. 
Is there something I am missing or not taking into consideration ?

themarpe: Misread the first check, all good in this case (just recheck other functions if similar situation could occur)

themarpe (approved):

PR #158, opened 2021-06-14
Adjust visiblity and constness of member functions `dai::[Node]::get{{In,Out}put,Name}`

kunaltyagi:
EDITED:

* Make getInput and getOutput const
* Put getName, getInput and getOutput in public visibility

### DIFF
--- a/include/depthai/pipeline/Node.hpp
+++ b/include/depthai/pipeline/Node.hpp
@@ -34,14 +34,18 @@ class Node {
     struct Connection;
 
    protected:
+    // fwd declare classes
+    class Input;
+    class Output;
+
+    Output *firstOutput = nullptr, *lastOutput = nullptr;
+    Input *firstInput = nullptr, *lastInput = nullptr;
+
     struct DatatypeHierarchy {
         DatatypeHierarchy(DatatypeEnum d, bool c) : datatype(d), descendants(c) {}
         DatatypeEnum datatype;
         bool descendants;
     };
-
-    // fwd declare Input class
-    class Input;
 
     class Output {
         Node& parent;
@@ -170,11 +174,40 @@ class Node {
     /// Retrieves nodes name
     virtual std::string getName() const = 0;
     /// Retrieves all nodes outputs
-    virtual std::vector<Output> getOutputs() const = 0;
+    std::vector<Output> getOutputs() {
+        std::vector<Output> result;
+        for (auto* x: getOutputRefs()) {
+            result.push_back(*x);
+        }
+        return result;
+    }
     /// Retrieves all nodes inputs
-    virtual std::vector<Input> getInputs() const = 0;
+    std::vector<Input> getInputs() {
+        std::vector<Input> result;
+        for (auto* x: getInputRefs()) {
+            result.push_back(*x);
+        }
+        return result;
+    }
     /// Retrieves all nodes assets
     virtual std::vector<std::shared_ptr<Asset>> getAssets();
+
+    std::vector<Output*> getOutputRefs() {
+        if(firstOutput == nullptr) return {};
+        return std::vector<Output*>{firstOutput, lastOutput};
+    }
+    std::vector<const Output*> getOutputRefs() const {
+        if(firstOutput == nullptr) return {};
+        return std::vector<const Output*>{firstOutput, lastOutput};
+    }
+    std::vector<Input*> getInputRefs() {
+        if(firstInput == nullptr) return {};
+        return std::vector<Input*>{firstInput, lastInput};
+    }
+    std::vector<const Input*> getInputRefs() const {
+        if(firstInput == nullptr) return {};
+        return std::vector<const Input*>{firstInput, lastInput};
+    }
 
     /// Connection between an Input and Output
     struct Connection {
--- a/include/depthai/pipeline/Node.hpp
+++ b/include/depthai/pipeline/Node.hpp
@@ -38,8 +38,8 @@ class Node {
     class Input;
     class Output;
 
-    Output *firstOutput = nullptr, *lastOutput = nullptr;
-    Input *firstInput = nullptr, *lastInput = nullptr;
+    std::vector<Output*> outputs;
+    std::vector<Input*> inputs;
 
     struct DatatypeHierarchy {
         DatatypeHierarchy(DatatypeEnum d, bool c) : datatype(d), descendants(c) {}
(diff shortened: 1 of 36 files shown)


themarpe (approved):
Thanks for the PR, makes sense:)

themarpe (changes requested):
Although I think an overload has to be done, Input & Output provide with the way of accessing the parent node as well as its own properties like queue sizes, etc... which in `const` case shouldn't be changed (especially the `Node` itself)
```
virtual std::vector<Output> getOutputs() = 0;
virtual std::vector<Input> getInputs() = 0;

virtual std::vector<const Output> getOutputs() const = 0;
virtual std::vector<const Input> getInputs() const = 0;
```

kunaltyagi:
getInput and getOutput return a freshly created vector in all current nodes. The const-ness of `value_type` of the vector will not affect the API in any case. Examples:
```cpp
// current, applies to Input as well
std::vector<Output> p = node->getOutput();
p[0].parent;  // is a reference can't be changed
p[0].<any other member>  // doesn't affect the node because it's a copy vector

Pipeline pipe = p[0].parent.getParentPipeline();  // do what we want

// suggested
std::vector<Output> p1 = node->getOutput();
std::vector<const Output> p1 = node->getOutput();  // compilation error
std::vector<Output> q1 = pipeline.getNode(0)->getOutput();  // compilation error
std::vector<const Output> q2 = pipeline.getNode(0)->getOutput();

q2[0].parent; // is a reference, couldn't have been changed anyways
q2[0].<any other member>  // saved from being changed

Pipeline pipe = [0].parent.getParentPipeline();  // still do what we want
```
[Godbolt link](https://godbolt.org/z/hhrE1bdTj)

IMO the proposed API will
* break generic code
* not provide any significant benefits

themarpe:
Thanks for the godbolt link - had an idea that const applies transitively to references/pointers as well. 

IMO Input/Output::parent should then be private and an overloaded function to access either const or non-const reference should be made. 
Also `getInputs` & `getOutputs` should return const & non-const references/pointers for better capabilities of accesing IOs and being able to modify them as well.

With both of above applied I think it would make the full circle.
Thoughts?

Also, what do you mean with: 
> break generic code

kunaltyagi:
I think I can manage the modification for parent in this PR. The change for IO handling is too wide and should be tackled separately.

The reason for that is the same as the ones for breaking generic code.

```cpp
std::vector<const Input> a = std::vector<Input>{};  // throws a compilation error

std::shared_ptr<const Input> b = std::shared_ptr<Input>{};  // works as expected
```
Here example 1 breaks generic code assumption by not treating the containers themselves as values. Const variables can always be assigned from a non-const and a const reference can be created for a non-const variable/reference.

Any templated code will use these assumptions and making it work when these assumptions are invalidated is quite irritating.

themarpe:
Lets do the following then and merge afterwards:
- include/utility/ArrayOf.hpp
```
#pragma once 

namespace dai {

// Creates std::array without explicitly needing to state the size
template <typename V, typename... T>
static constexpr auto array_of(T&&... t) -> std::array<V, sizeof...(T)> {
    return {{std::forward<T>(t)...}};
}

}
```

And the following for nodes:
```
struct NodeX {
    Output a, b, c;
    std::array<Output*, 3> outputsRef = array_of<Output*>(&a, &b, &c);

    std::vector<Output> getOutputs(){
        std::vector<Output> vec;
        for(auto* out : outputsRef){
            vec.push_back(*out);
        }
        return vec;
    }
    std::vector<Output*> getOutputsRef(){
        return std::vector<int*>{begin(outputsRef), end(outputsRef)};;
    }
    const std::vector<const Output*> getOutputsRef() const{
        return std::vector<Output int*>{begin(outputsRef), end(outputsRef)}; 
    } 

   // Same for inputs
   ...
};
```

So we leave the getOutputs & getInputs without const and add getOutputsRef with const that returns pointer to const.

Thoughts?

kunaltyagi:
Same issues of this not working for the following:
```cpp
std::vector<Output*> p1 = node->getOutputRef();  // v1
std::vector<const Output*> p1 = node->getOutputRef();  // v2, compilation error

std::vector<Output*> q1 = pipeline.getNode(0)->getOutputRef();  // v3, compilation error
std::vector<const Output*> q2 = pipeline.getNode(0)->getOutputRef();  // v4
```

Ideally, const overloading shouldn't cause compilation errors on removing the const. As in the error in v3 is perfect, expected. But not the v2 error.

Your suggestion solves the issue of not mutating the pipeline, but the core issue are quite deep and require significant more work.

themarpe:
@kunaltyagi unfortunatelly there isn't any automatic casting between vector with pointers to const and regular as is for shared_ptr, etc...

In this case you have to use the type as given. Apart from using a different container supporting this I don't think we'll be able to achieve the automatic conversion.

> Ideally, const overloading shouldn't cause compilation errors on removing the const. As in the error in v3 is perfect, expected. But not the v2 error.

You mean on adding const? I agree if the container supports. But the thing is that here we don't expect to cause any issues with this as you can still do everything as expected with the returned type. The compilation error is only because of wrong usage. 

Most importantly, the main way I'd use this is like:
```
auto outputs = node->getOutputsRef();
for(auto out : outputs){ ... }
// or
for(const auto out : outputs) { ... }
```
Where the const is going to be applied to `out` as necessary

kunaltyagi:
PTAL

include/depthai/pipeline/Node.hpp
```diff
@@ -34,18 +34,24 @@ class Node {
     struct Connection;
 
    protected:
+    // fwd declare classes
+    class Input;
+    class Output;
+
+    Output *firstOutput = nullptr, *lastOutput = nullptr;
+    Input *firstInput = nullptr, *lastInput = nullptr;
```
themarpe, line 42: This is dangerous as it requires correct order of members. I'd suggest an array as described in https://github.com/luxonis/depthai-core/pull/158#issuecomment-864986329 (Or some other way that does not enforce this limitation, which can quickly be misused).

themarpe (changes requested):

include/depthai/pipeline/Node.hpp
kunaltyagi: I used this because of 2 main reasons:
* it made it possible to have everything non-virtual
* minimal effort in adding across different and new nodes
Downside is that it was brittle.

Sadly, C++ doesn't really have good mechanism for reflection. Perhaps, instead of `std::array<T*>`, what are your thoughts on putting all implementation using `std::vector<T*>` in the Node class?

themarpe: Right - array will be different type for different sizes. Yeah std::vector sounds good as well.

include/depthai/pipeline/Node.hpp
```diff
@@ -154,12 +174,41 @@ class Node {
...
+        return outputs;
+    }
+    std::vector<const Output*> getOutputRefs() const {
+        if(outputs.empty()) return {};
+        return {outputs.begin(), outputs.end()};
+    }
+    std::vector<Input*> getInputRefs() {
+        if(inputs.empty()) return {};
+        return inputs;
+    }
```
themarpe, line 206: Non const can be just `return outputs` and `return inputs` afaik.

kunaltyagi: No, I get a compile error (can't convert from vector<T*> to vector<const T*>)

kunaltyagi:
@themarpe PTAL

themarpe:
Will check locally and merge now that 2.6.0 is out. Hopefully today. Thanks @kunaltyagi

PR #159, opened 2021-06-14
Add support for 5x5 bilateral filter in stereo depth; add runtime con…

szabi-luxonis:
…figurability for stereo depth confidence threshold/bilateral filter sigma value.

In the future we will add more config options like: switching depth mode at runtime, configurable stereo match parameters (census kernel settings) etc.

Related PR: https://github.com/luxonis/depthai-shared/pull/39

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -170,6 +170,7 @@ add_library(${TARGET_CORE_NAME}
     src/pipeline/datatype/SpatialLocationCalculatorConfig.cpp
     src/pipeline/datatype/Tracklets.cpp
     src/pipeline/datatype/IMUData.cpp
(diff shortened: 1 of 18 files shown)


themarpe (commented):
IMO we should combine the median filter enum to a common place.
Otherwise LGTM

themarpe (changes requested):
I think the new message is missing in hierarchy

szabi-luxonis:
> I think the new message is missing in hierarchy

It's added [here](https://github.com/luxonis/depthai-shared/pull/39/files#diff-2afeb2df4497e69c27e911e090d806c41bdb282c9978e2089809d031a15af569R35)

themarpe (approved):

PR #161, opened 2021-06-16
Add const to XLinkStream member functions

kunaltyagi:


themarpe:
Not sure about this one - const can mean different things, and in case where object controls underlying mutable things, I would say that constness should extend to that. 
This is also the reason why Device getters which do RPC calls aren't const.

kunaltyagi:
I was using const to allow a lambda capture to work. The other option would be to use a mutable lambda, but that affects all captures.

I'll figure something out in this case.

PR #163, opened 2021-06-25
Calib fix

saching13:
- Fixes Spatial node zero when baseline is inverted.
- Contains FW fix for Inverted specTranslation
- Extrinsic calculation fix
- V5 Update sign fix


szabi-luxonis (approved):
LGTM but open PR for python too

PR #173, opened 2021-07-14
Close queues when closing the device

themarpe:
When device is closed, the queues should be as well. The issues arises as same link id is used for next XLink connection and old `XLinkStream` handles close stream of the new connection


alex-luxonis (approved):
Looks great! After the last update, all XLink streams are closed by the time `Device::close()` ends.

PR #189, opened 2021-07-28
XLink Error 221 fix

themarpe:
Added an XLink 221 fix in FW and a default detection network confidence threshold.

Fixes: https://github.com/luxonis/depthai-python/issues/300

### DIFF
--- a/src/pipeline/node/DetectionNetwork.cpp
+++ b/src/pipeline/node/DetectionNetwork.cpp
@@ -14,6 +14,9 @@ namespace node {
 DetectionNetwork::DetectionNetwork(const std::shared_ptr<PipelineImpl>& par, int64_t nodeId) : NeuralNetwork(par, nodeId) {
     inputs = {&input};
     outputs = {&out, &passthrough};
+
+    // Default confidence threshold
+    getPropertiesRef().confidenceThreshold = 0.5;
 }
 
 std::string DetectionNetwork::getName() const {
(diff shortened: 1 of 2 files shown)


src/pipeline/node/DetectionNetwork.cpp
```diff
@@ -14,6 +14,9 @@ namespace node {
 DetectionNetwork::DetectionNetwork(const std::shared_ptr<PipelineImpl>& par, int64_t nodeId) : NeuralNetwork(par, nodeId) {
     inputs = {&input};
     outputs = {&out, &passthrough};
+
+    // Default confidence threshold
+    getPropertiesRef().confidenceThreshold = 0.5;
```
szabi-luxonis, line 19: IMO rather set in properties a default value, more transparent

szabi-luxonis (changes requested):

src/pipeline/node/DetectionNetwork.cpp
themarpe: For properties, as they aren't constructed by FW, I think its better to do it in core, as it doesn't require updating FW.
How about specifying in both places (Could still be changed only in core by eg. a PR)?

This reasoning is mostly targeted toward cases where more people would be adding to core as part of an open source project.

szabi-luxonis: Sounds good

PR #190, opened 2021-07-28
Bootloader updates

alex-luxonis:
- Fix issues with upgrading bootloader/application with a non-empty flash (due to unaligned erase).
Add back bootloader flash example (was removed previously due to issues).

- Add support for more NOR flash chip types used on the 1099/2099 SoMs.

Related PR: https://github.com/luxonis/depthai-bootloader-shared/pull/3


alex-luxonis:
TODO: when flashing the bootloader/application, if a bootloader already exists, don't use it to flash, but load first the new bootloader over USB. We may also need to enforce flashing the new bootloader.

themarpe:
Ready for initial re-review 
@alex-luxonis @szabi-luxonis

examples/src/bootloader_config.cpp
```diff
@@ -0,0 +1,68 @@
+#include "depthai/depthai.hpp"
+
+int main(int argc, char** argv) {
+    // Options
+    bool usage = false, read = true, clear = false;
+    std::string path = "";
+    if(argc >= 2) {
+        std::string op = argv[1];
```
szabi-luxonis, line 8: We can add argparser, looks like this is a copy of the python implementation and header only:
https://github.com/jamolnng/argparse

szabi-luxonis (approved):

examples/src/bootloader_config.cpp
themarpe: Will leave this as a side task to refactor one day. Good suggestions to match Python 👍

PR #191, opened 2021-08-04
ImgFrame: report camera settings (focus, exposure, iso)

alex-luxonis:
Add new API which can be used on `ImgFrame` packets received from camera:
- getExposureTime()
- getSensitivity()
- getLensPosition() - ~~(note: lens position may be reported one frame in advance at the moment)~~ should be fixed
- getColorTemperature()

FW updates:
- fix black image when initial manual exposure was set with other controls **-- was merged previously**
- fix switching from auto to manual exposure causing the lens to jump to zero 
  (note: the lens is directly set now with manual focus, bypassing 3A) **-- was merged previously**
  
~~WIP, currently only reports values for ColorCamera.~~ - Fixed

Related PR: https://github.com/luxonis/depthai-shared/pull/51


src/pipeline/datatype/ImgFrame.cpp
```diff
@@ -43,6 +43,15 @@ unsigned int ImgFrame::getHeight() const {
 RawImgFrame::Type ImgFrame::getType() const {
     return img.fb.type;
 }
+int ImgFrame::getExposureTime() const {
+    return img.cam.exposureTimeUs;
+}
```
themarpe, line 48: IMO duration would be more applicable in this case
```
std::chrono::microseconds ImgFrame::getExposureTime() const {
    return std::chrono::microseconds(img.cam.exposureTimeUs);
}
```

themarpe (changes requested):

themarpe (changes requested):
Seems like `video` frames don't get this information applied to it, but `preview` do. 
TODO - check how FW handles this.

Erol444:
This (newer) PR also includes features added in this PR: https://github.com/luxonis/depthai-core/pull/210

alex-luxonis:
> Seems like `video` frames don't get this information applied to it, but `preview` do.
`VideoEncoder` didn't forward the new ImgFrame metadata, was fixed.

PR #194, opened 2021-08-06
Release v2.9.0

themarpe:
## Known issues
- Bootloader interaction (flashing and bootloader upgrading) is known to not work. A followup bugfix release will address this issue 

## Features
- Script Node (On device scripting using Python 3.9)
- SPIIn Node (Receive messages through SPI Interface)
- Asset managment improvement

## Bugs
- XLink segfault fix
- Performance regression fix when using USB due to changed L2 cache behaviour from copy back to write through.
- Added missing IMU calibration field when reading EEPROM data

## Misc
- Dependencies without `git clone` for piwheels compatibility
- Code changes: https://github.com/luxonis/depthai-core/pull/194


examples/src/depth_crop_control.cpp
```diff
@@ -69,7 +69,8 @@ int main() {
         cv::imshow("depth", depthFrameColor);
 
         // Update screen (10ms pooling rate)
-        int key = cv::waitKey(10);
+        int key = cv::waitKey(9);
+        cv::waitKey(1);  // glitch workaround
```
szabi-luxonis, line 73: what is this for and why 10 ms?

include/depthai/pipeline/node/Script.hpp
```diff
@@ -0,0 +1,82 @@
...
+#include <fstream>
+
+// shared
+#include <depthai-shared/properties/ScriptProperties.hpp>
+
+namespace dai {
+namespace node {
+
+class Script : public Node {
+    dai::ScriptProperties properties;
```
szabi-luxonis, line 16: set alias `using Properties = dai::ScriptProperties` as in other nodes.

szabi-luxonis (commented):
Just some minor things I noticed.

szabi-luxonis:
Shall we mention the addition of IMU config to eeprom (bugfix/misc) and performance regression fix when using USB (fix or misc) due to cache behaviour in changelog?

Luxonis-Brandon:
> Shall we mention the addition of IMU config to eeprom (bugfix/misc) and performance regression fix when using USB (fix or misc) due to cache behaviour in changelog?

Yes good idea!

themarpe:
@szabi-luxonis sure - feel free to edit the release notes directly (in this PR)
I'll copy at the end to final release

examples/src/depth_crop_control.cpp
themarpe: 10ms was used before and the workaround is that when holding down any of the keys, the movement got stuck and frames didn't continue updating after releasing. With separate `waitKey` this doesn't happen anymore and actual behavior mimics that of Python example

szabi-luxonis:
> @szabi-luxonis sure - feel free to edit the release notes directly (in this PR)
> I'll copy at the end to final release

Added as bugs

examples/src/depth_crop_control.cpp
szabi-luxonis: Hmm, interesting. Could be because it's on the same thread, while python probably spawns different threads.
It's good to merge for now.

szabi-luxonis (approved):

alex-luxonis:
Missing version bump (to `2.9.0`) in CMakeLists.txt

themarpe:
Thanks - updated!
Will release in the morning after the checks and wheel CI pass

PR #198, opened 2021-08-16
Throw XLink specific errors for read/write errors

kunaltyagi:


kunaltyagi:
Created from main. Develop seems to be behind main

szabi-luxonis (changes requested):
LGTM, but need to add python bindings too in `depthai-python`

kunaltyagi:
Windows CI failing for some reason. Python bindings are added post merge in core, right?

include/depthai/xlink/XLinkStream.hpp
```diff
@@ -60,4 +60,14 @@ class XLinkStream {
     streamId_t getStreamId() const;
 };
 
+struct XLinkError : public std::runtime_error {
+    using std::runtime_error::runtime_error;
```
themarpe, line 64: Could add a field something along the lines of:
`XLinkError_t xlinkError` / `XLinkError_t status`
And populate with returned status for extra information to be passed down.

themarpe (changes requested):
I think this is the right direction for better exception handling. Thanks!

themarpe:
@kunaltyagi Python bindings can be done in same manner, separate branch, targeting develop, which complements changes in core branch. When core is merged, so is the branch in bindings and submodules are again updates as necessary. (bindings branch points to this branch commit before hand)

kunaltyagi:
image
There are no bindings for XLinkStream

include/depthai/xlink/XLinkStream.hpp
kunaltyagi: What would the API look like? Should the ctor change? Should the string message be moved to the what function?

themarpe: Hmm, I'd go with an additional ctor:
`XLinkError::XLinkError(XLinkError_t, std::string/const char* msg)`

otherwise the field can be accessed publicly for consumption (or for setting to another value, etc...).

themarpe: And default ctors would just leave the field as is - which could have default value of `X_LINK_ERROR`

kunaltyagi: Made some changes to the API. PTAL (No default ctor: doesn't make sense in the use-case right now)

kunaltyagi:
https://github.com/luxonis/depthai-python/pull/345

include/depthai/xlink/XLinkStream.hpp
```diff
@@ -60,4 +60,20 @@ class XLinkStream {
...
+
+    XLinkError(XLinkError_t status_, const std::string& streamName_, const std::string& message_)
+        : runtime_error(message_), status(status_), streamName(streamName_) {}
+};
+struct XLinkReadError : public XLinkError {
+    XLinkReadError(XLinkError_t status_, const std::string& streamName_);
+};
+struct XLinkWriteError : public XLinkError {
+    XLinkWriteError(XLinkError_t status_, const std::string& streamName_);
+};
```
themarpe, line 77: Retain option capability to set message - otherwise default is used.
(& naming convention for parameters, remove ending `_`)

themarpe (changes requested):

include/depthai/xlink/XLinkStream.hpp
kunaltyagi: Suggested names for arguments?

themarpe: Can just remove the trailing `_`. Should shadow correctly AFAIK. Otherwise any abbreviation (status->s, streamName->stream/name, message->msg)
```
struct XLinkError : public std::runtime_error {
    const XLinkError_t status = 0;
    const std::string streamName;

    using std::runtime_error::runtime_error;

    XLinkError(XLinkError_t status, const std::string& streamName, const std::string& message)
        : runtime_error(message), status(status), streamName(streamName) {}
};

struct XLinkReadError : public XLinkError {
    XLinkReadError(XLinkError_t status, const std::string& streamName);
};
struct XLinkWriteError : public XLinkError {
    XLinkWriteError(XLinkError_t status, const std::string& streamName);
};
```

kunaltyagi:
It should be ready for a merge

themarpe (approved):
Thanks!

kunaltyagi:
Can we get this merged?

themarpe:
@kunaltyagi 
I've fixed an incorrect set exception messages and added bindings in python.
Will merge after release, today/tomorrow.

Thanks!

PR #199, opened 2021-08-17
Configure XLink chunk size

alex-luxonis:
Adding new API:

    /**
     * Set chunk size for splitting device-sent XLink packets, in bytes. A larger value could
     * increase performance, with 0 disabling chunking. A negative value won't modify the
     * device defaults - configured per protocol, currently 64*1024 for both USB and Ethernet.
     */
    void Pipeline::setXlinkChunkSize(int sizeBytes);

The defaults on device side were not changed, as it might be possible (mainly for USB case) that increasing the chunk size won't work with certain hosts / OSes.

For USB, if supported by the host, best performance would be achieved with:  `pipeline.setXlinkChunkSize(0);`
This should increase the max available USB bandwidth and reduce the device CPU load (less overhead from splitting).

Related PR: https://github.com/luxonis/depthai-shared/pull/53

### DIFF
--- a/include/depthai/pipeline/Pipeline.hpp
+++ b/include/depthai/pipeline/Pipeline.hpp
@@ -36,6 +36,7 @@ class PipelineImpl {
     PipelineSchema getPipelineSchema() const;
     OpenVINO::Version getPipelineOpenVINOVersion() const;
     void setCameraTuningBlobPath(const std::string& path);
+    void setXlinkChunkSize(int sizeBytes);
 
     // Access to nodes
     std::vector<std::shared_ptr<const Node>> getAllNodes() const;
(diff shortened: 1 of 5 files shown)


include/depthai/pipeline/Pipeline.hpp
```diff
@@ -36,6 +36,7 @@ class PipelineImpl {
     PipelineSchema getPipelineSchema() const;
     OpenVINO::Version getPipelineOpenVINOVersion() const;
     void setCameraTuningBlobPath(const std::string& path);
+    void setXlinkChunkSize(int sizeBytes);
```
themarpe, line 39: `void setXLinkChunkSize(int sizeBytes);`

themarpe (approved):
Otherwise LGTM

include/depthai/pipeline/Pipeline.hpp
themarpe: ^ And related functions

themarpe:
Also, would it make sense adding an RPC call for this?

themarpe (approved):

PR #210, opened 2021-09-09
Multiple cameras support, any camera type on any socket/port

alex-luxonis:
Any camera type (from the supported list) can be used on any board socket/port: 
- RGB / CENTER / CAM_A
- LEFT / CAM_B
- RIGHT / CAM_C
- CAM_D

Supports for now:
- ColorCamera: IMX378, IMX477, OV9782, IMX214, AR0234.
- MonoCamera: OV9282/OV9281, OV7251.

WIP: support for more sensors.

Limitations: 
- for multiple cameras of the same type (also including OV9282/9281 and OV9782, which use the same sensor die, except the CFA - color filter array), they need to use the same config for now (e.g. resolution, fps, readout order). WIP on fixing this.
- if AR0234 cameras are used (one or multiple), cannot use other camera modules on the other ports, due to the structure of the camera tuning blob. WIP on fixing this.


Luxonis-Brandon:
SWEET!

slitcch:
I've been using this on Ubuntu 20.04 for a while - no problems so far!

oysstu (commented):
@alex-luxonis In what way is the image quality degraded if for example the AR0234 is combined with OV9282?

alex-luxonis:
That's correct, the color channels (R, G, B) could have their calibration not properly applied (with specific gains to compensate). But in general the debayer process would still be properly applied, so the basic R, G, B should still look good (not swapped).

oysstu:
Got a combination of AR0234 and OV9282 working. The colors were a bit off, but acceptable (perhaps a bit dark though). What is the ETA on the current limitations and getting this merged? Also, what are your intentions regarding sensor support (i.e. how much work is it for you guys to add support for new sensors)? I'm guessing the support needs to be added in firmware, and thus closed-source due to the Intel proprietary tooling. Will the upgrade to oak series 3 / keem bay enable you to open the firmware, or is that still proprietary?

alex-luxonis:
@oysstu For MyriadX, yes, support needs to be added by us in firmware, once we get the sensor documentation (datasheet, register reference manual, register config arrays / INI files) and the hardware to test, it's around 2 weeks of work. Note this will not include camera module (sensor, lens, VCM/focus) IQ (image quality) tuning, but using a tuning database of a different sensor might be acceptable enough.
For mainlining this PR, it's on my plate now, I'm fixing last issues (board-dependent camera enumeration order for faster application loading) and per-sensor independent config for multiple sensors of the same type.

For Keem Bay, it's possible this procedure could be simplified, as the user will have some access to the Linux/ARM side, where the sensor/CCM config driver (I2C, power, etc) resides, and 3A runs as well on Linux, but the tuning tools might still be closed-source (proprietary). MIPI / ISP will run on the closed VPU side, but we may be able to pass the MIPI config to it at runtime (number of lanes, expected lane speed for receiver/PHY tuning, image resolution to expect). We don't know for sure yet.

oysstu:
I see. More openness in Keem Bay would definitely increase the flexibility of the platform. We've been looking at a [camera with an IMX264 sensor](https://www.leopardimaging.com/wp-content/uploads/LI-IMX264-MIPI-M12_Datasheet.pdf), but unfortunately it's not supported yet.

Will the issue with the camera tuning blob, discussed earlier in thread, be resolved before it's merged (i.e. the AR0234 using the IMX378 tuning blob when paired with OV9282)? After more testing, the colors were more off than I thought previously, but it's somewhat compensated by adjusting the white balance.

alex-luxonis:
Yeah, IMX264 is a nice sensor: 5MP, large 3.45um pixels, global shutter color, only limitation in pairing with our platform is the (S)LVDS video interface instead of the supported MIPI D-PHY, but it can be solved with a converter FPGA as the LI module is using (link above).
We were looking at this sensor as well, but don't have a timeline for when it would be supported.

About the camera tuning blobs, this will be resolved later. More details on the issue: 
- a single tuning blob can be loaded at a time when the camera system is initialized (we store multiple of them in firmware, and auto-select one based on detected cameras)
- a tuning blob can contain data for multiple cameras, now we have blobs like:
 1. IMX378 + OV9282. Also applied for RGB sensors IMX477 / IMX577
 2. IMX214 + OV7251
 3. AR0234 only
  In case AR0234 camera is combined with others like OV9282, we have to switch from using blob `3` to blob `1`, to have a tuning that supports also mono cameras. A proper fix for this would be to create a single blob with all camera tunings merged into.

Manual white balance can help to fix the color issues, and we'll add more control there, instead of selecting a color temperature to compensate for, we'll add a control to set color gains per each channel (`R`, `Gr`, `Gb`, `B`).

oysstu:
Thanks for the detailed reply. The ability to set gains per channel sounds great. Would these gains be analog or digital (i.e. before/after ADC)? It would be nice to have issues to track for these improvements, since these are not a part of this PR (tuning blobs, per-channel-gains, IMX264 support).

rupa247:
Hello @alex-luxonis 
When are you going to merge this pull request? Will there be a release soon?

themarpe:
@rupa247 
We are planning on releasing this alongside 2.18.0 by the end of this week :)

themarpe (approved):
LGTM

oysstu:
Seeing as this is about to get merged, what is the status for the WIPs for the stated limitations? I've tried to use the AR0234 with the multi cam support in its current state, and it's unfortunately not providing the desired performance (too dark). At the moment I'm unsure if this is due to the camera tuning blob, or if the sensor/lens combination simply requires more light than we can provide. At any rate, the AR0234 is a nice global shutter sensor for many use cases if it's properly supported.

alex-luxonis:
@oysstu I added some improvements for AR0234 exposure with the latest. The tuning is still not updated, and it limits AE to about 16ms (note though it's limited to 10ms with the default anti-banding for 50Hz lighting where the exposure gets set in multiples of 10ms -- half-of-mains-period, so for higher AE anti-banding should be set to OFF). There was also a driver limitation of max exposure around 16ms, it's lifted now, so manual exposure can be set higher, limited by the configured FPS (for example, up to 100ms for 10fps).

You can check the actual exposure values (set by AE, or manual) from ImgFrame metadata which was added recently. See here an example: https://github.com/luxonis/depthai-python/commit/57c3cc28c52387fb0ce33bc868c9e32a36d84983 . Also can experiment with changing anti-banding setting with that example.

We'll try to get some updated tuning as well in the future.

PR #216, opened 2021-09-23
Fine tune stereo depth settings

szabi-luxonis:
Enabling census transform mean mode and alpha=0, beta=2 census transform parameters improves the overall disparity map by improving the descriptors generated by census transform.
Valid data, a.k.a. fill level (== % of values that are not invalid) of disparity map increases by ~20%.

400p:
Before:
stereo_400_current

After:
stereo_400_mean_0_2

800p:
Before:
stereo_800_current

After:
stereo_800_mean_0_2

There is more in census transform settings to squeeze out that I'm looking into.

CC: @saching13 for testing with your setup for depth measurement.


themarpe (changes requested):
CC: @saching13 
I think we should compare these changes to some close to ground by feeding the datasets through it like: http://www.cvlibs.net/datasets/kitti/, just to be able to see how this affects the accuracy, etc...

saching13:
> CC: @saching13
> I think we should compare these changes to some close to ground by feeding the datasets through it like: http://www.cvlibs.net/datasets/kitti/, just to be able to see how this affects the accuracy, etc...

actually that's a great idea. Will do it.

Luxonis-Brandon:
Is there a Python version of this I can run or should I get C++ up/running?  Sorry if dumb question.

szabi-luxonis:
> Is there a Python version of this I can run or should I get C++ up/running? Sorry if dumb question.

The images were taken actually with a python script, I just happened to add them to core PR, because core contains the proposed changes.
Install the following library:
```
python3 -m pip install --user --extra-index-url https://artifacts.luxonis.com/artifactory/luxonis-python-snapshot-local "depthai==2.10.0.0.dev+3fdf376f871d1144e6f677240050211a897d0ef5"
```
Use the following branch `temp_depth_quality_test`
https://github.com/luxonis/depthai-python/tree/temp_depth_quality_test/examples

They are temporary changes added to `spatial_mobilenet.py` + `spatial_location_calculator.py`, since we should keep them simple, adding all the configs to them is too much.

Luxonis-Brandon:
Thanks!  So @VanDavv - we should definitely get this slider functionality into some advanced UI in depthai_demo.py.  What would be ideal is to have an advanced checkbox that opens up a window with all these controlls.

Thoughts?

PR #232, opened 2021-10-20
Restructured README.md

themarpe:
Refactored README to expose common use cases first, and made integration steps a bit more clear.

Also replaced `-H` parameter with `-S` and added a warning if an user might need to use `-H` (old version)

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,6 +66,8 @@ cd build/examples
 
 ## Integration
 
+Under releases you may find prebuilt library for Windows, for use in either integration method. See [Releases](https://github.com/luxonis/depthai-core/releases)
+
 ### CMake
 
 Targets available to link to are:


Erol444 (approved):

README.md
```diff
@@ -21,6 +21,50 @@ MacOS: `brew install libusb`, optionally with `brew install opencv`
...
+```
+
+> ℹ️ To speed up build times, use `cmake --build build --parallel [num CPU cores]` (CMake >= 3.12).
+For older versions use: Linux/macOS: `cmake --build build -- -j[num CPU cores]`, MSVC: `cmake --build build -- /MP[num CPU cores]`
+
+> ⚠️ If any CMake commands below error with `CMake Error: The source directory "" does not exist.` replace argument `-S` with `-H`
+
+### Static library
+
+To build dynamic version of library configure with following option added
```
alex-luxonis, line 38: Typo: ~~dynamic~~ static

Issue #246, opened 2021-11-08
fails to cmake config+build on Windows using cmake and vscode (hunter pain)


diablodale:
depthai-core fails to config (and therefore can't build) on a clean repo and using VSCode.
Cascading failures in hunter.

### Setup

* Microsoft Windows [Version 10.0.19043.1288]
* VS2019 specifics below
* depthai-core main (currently it is c1b697afa432c8d620b097ba57e20a0ffd2ec52d)

### Repro

1. install VS2019 community 16.11.5 with: Desktop Development c++, linux dev with c++. In installation details for Desktop Development include C++ CMake tools for Windows.
2. install vscode x64 for windows
3. install git for windows
4. `git clone https://github.com/luxonis/depthai-core` into an empty folder.
5. `git checkout c1b697afa432c8d620b097ba57e20a0ffd2ec52d` so to have less variations in the setup.
6. run VSCode
7. Add the extensions for: C/C++, CMake Tools. Both of these are made by Microsoft. Restart vscode if prompted.
8. File/open the `depthai-core` folder. Select `yes` to trust.
9. At bottom right, select `NOT NOW` and `NOT THIS WORKSPACE` at the configure project popup.
10. At bottom center, you should see "No Kit Selected". Click that and select "Visual Studio Community 2019 Release - amd64"
11. Bottom center should update with no errors to show "Visual Studio Community 2019 Release - amd64"
12. Bottom left should show the checked out commit c1b697af
13. Note, since `BUILD_SHARED_LIBS` is not declared anywhere, it should be a static library build
14. in the left nav, click the CMake icon. At the top of the left nav beside the "CMAKE: PROJECT OUTLINE" click the first icon for Configure All Projects.

### Result

```
[main] Configuring folder: depthai-core 
[proc] Executing command: "C:\Program Files\CMake\bin\cmake.exe" --no-warn-unused-cli -DCMAKE_EXPORT_COMPILE_COMMANDS:BOOL=TRUE -DCMAKE_BUILD_TYPE:STRING=Debug -Hc:/repos-nobackup/depthai-core -Bc:/repos-nobackup/depthai-core/build -G Ninja
[cmake] Not searching for unused variables given on the command line.
[cmake] -- The CXX compiler identification is MSVC 19.29.30136.0
[cmake] -- The C compiler identification is MSVC 19.29.30136.0
[cmake] -- Detecting CXX compiler ABI info
[cmake] -- Detecting CXX compiler ABI info - done
[...]

PR #258, opened 2021-11-11
enable build in vscode, custom toolchain+include

diablodale:
fixes luxonis/depthai-core#246

full details on how this fix can be used in vscode is in the above issue.

### DIFF
--- a/cmake/depthaiDependencies.cmake
+++ b/cmake/depthaiDependencies.cmake
@@ -63,6 +63,11 @@ endif()
 # OpenCV 4 - (optional, quiet always)
 find_package(OpenCV 4 QUIET CONFIG)
 
+# include optional dependency cmake
+if(DEFINED DEPTHAI_DEPENDENCY_INCLUDE)
+    include(${DEPTHAI_DEPENDENCY_INCLUDE} OPTIONAL)
+endif()
+
 # Cleanup
 if(CONFIG_MODE)
     set(CMAKE_PREFIX_PATH ${_CMAKE_PREFIX_PATH_ORIGINAL})
--- a/cmake/depthaiDependencies.cmake
+++ b/cmake/depthaiDependencies.cmake
@@ -64,7 +64,7 @@ endif()
 find_package(OpenCV 4 QUIET CONFIG)
 
 # include optional dependency cmake
-if(DEFINED DEPTHAI_DEPENDENCY_INCLUDE)
+if(DEPTHAI_DEPENDENCY_INCLUDE)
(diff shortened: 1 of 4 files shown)


cmake/depthaiDependencies.cmake
```diff
@@ -63,6 +63,11 @@ endif()
 # OpenCV 4 - (optional, quiet always)
 find_package(OpenCV 4 QUIET CONFIG)
 
+# include optional dependency cmake
+if(DEFINED DEPTHAI_DEPENDENCY_INCLUDE)
+    include(${DEPTHAI_DEPENDENCY_INCLUDE} OPTIONAL)
+endif()
```
themarpe, line 69: `depthaiDependencies.cmake` are installed as well to specify correctly all dependencies.
Any requirement to carry over the specified include over to be installed as well?

Otherwise as its guarded LGTM for non-installed use case

themarpe (approved):
Thanks for the PR!

cmake/depthaiDependencies.cmake
themarpe: Worded it oddly - meant to ask if the `DEPTHAI_DEPENDENCY_INCLUDE` file should be installed as well, so that consumers get this information as well, or should this be for local use only?

Yeah, I think it'd be good for discoverability. One a path and one a string type.

diablodale: I slightly lean towards not installing the file `DEPTHAI_DEPENDENCY_INCLUDE`.

* in my specific usage, it is only extra stuff needed to compile depthai and its test/examples. For example, I use TBB, IPP, and Eigen with my private OpenCV. So I need to find_package() those so that later cmake can correctly find includes and link EXEs.
* I tend to think that the devs that need this (a minority) will privately manage their `DEPTHAI_DEPENDENCY_INCLUDE` file for depthai itself -and- also for their final EXE. In my case, I have a cmake for my final EXE which manages its own find_package(), points to the depthai cmake, etc. For example, I need OpenCV directly in my EXE so I find_package() for TBB, IPP, and Eigen in the app's cmake. So I tend to think a dev needing this will self-manage and that they only need a way to adapt depthai without having to make private commits to it.

I'll add options for the two and force push after this.

themarpe: Sounds good - lets leave it for local only

diablodale: pushed, ci is running

diablodale:
I don't think this is a me/code issue. It complains of 
> Couldn't check if depthai-shared codebase matches between device and host

PR #262, opened 2021-11-15
Support for OAK-D-Lite

alex-luxonis:
- Firmware support and handling for the new resolutions of the cameras:

| Color Resolution | IMX378 (OAK-D) | IMX214 (OAK-D-Lite) |
| --- | ----------- | -- |
| `THE_1080_P` (1920×1080) | ✔️ binning+cropping | ✔️ binning+cropping |
| `THE_4_K` (3840×2160) | ✔️ cropping | ✔️ cropping |
| `THE_12_MP` (4056×3040) | ✔️ full | ✔️ cropping |
| `THE_13_MP` (4208×3120) | ❌ Not supported | ✔️ full |

| Mono Resolution | OV9282 (OAK-D) | OV7251 (OAK-D-Lite) |
| --- | ----------- | -- |
| `THE_400_P` (640×400) | ✔️ binning | ✔️ cropping |
| `THE_480_P` (640×480) | ❌ Not supported | ✔️ full |
| `THE_720_P` (1280×720) | ✔️ cropping | ❌ Not supported |
| `THE_800_P` (1280×800) | ✔️ full | ❌ Not supported |

If an unsupported resolution is configured, the firmware will log an error and default to a working one, but the host might be unaware of this, for example if using API when the pipeline was defined, like `ColorCamera::getResolution()`.
As a solution for that, a new function `device.getCameraSensorNames()` was added, which can be called after the device was created, but before sending down the pipeline. So the pipeline can be customized accordingly. It will usually return:
```
OAK-D:      RGB: "IMX378", LEFT: "OV9282", RIGHT: "OV9282"
OAK-D-Lite: RGB: "IMX214", LEFT: "OV7251", RIGHT: "OV7251"
```
Related PR: https://github.com/luxonis/depthai-shared/pull/63

### DIFF
--- a/cmake/Depthai/DepthaiDeviceSideConfig.cmake
+++ b/cmake/Depthai/DepthaiDeviceSideConfig.cmake
@@ -2,7 +2,7 @@
 set(DEPTHAI_DEVICE_SIDE_MATURITY "snapshot")
 
 # "full commit hash of device side binary"
-set(DEPTHAI_DEVICE_SIDE_COMMIT "3bb469fa08b7b0abc4353cd27219521dfd763a9d")
+set(DEPTHAI_DEVICE_SIDE_COMMIT "4354b366954d3c743e387b058e4bc369a6d0e758")
 
 # "version if applicable"
 set(DEPTHAI_DEVICE_SIDE_VERSION "")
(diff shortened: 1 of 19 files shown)


szabi-luxonis (approved):

themarpe (approved):
LGTM

PR #263, opened 2021-11-16
Adds a timeout for closing an XLink connection

themarpe:
Adds a timeout to XLinkConnection `close` as this can block main thread of execution if a device doesn't properly let host know that the link has fallen down (PoE devices specific)


alex-luxonis (approved):
Change target to `develop`.

Issue #248, opened 2021-11-09
truncation of struct initializers results in compiler warnings


diablodale:
The codebase (including the precompiled Windows distribution headers) has initializers that are truncating their values. This results in compiler warnings. This and other such errors I've found https://github.com/luxonis/depthai-core/issues/71 and https://github.com/luxonis/depthai-core/issues/247 and https://github.com/luxonis/depthai-core/issues/79 suggest the codebase has not been compiled with analyzers and warnings to surface easily found bugs. 

Three incidents I found in my compile are `double` truncated to `float`. As they are initializers, I would hope (but it is not promised across all compilers) that the executable would only store a float in the static init data. But if the compiler does not, then the size is doubled (4 -> 8 bytes) and additional bytecode is needed to convert the double -> float at runtime.

Naturally, none of that is desired and the fix is easy. Put an `f` after the number literal.

```c++
// errant
float epsilon = 0.01;

// correct
float epsilon = 0.01f;
```

### Setup

* all compilers and OS
* throughout codebase, yet can also be seen in the headers of precompiled Windows distrib:  `depthai-core-v2.11.1-win64`

### Repro

1. Enable warnings and errors in compiler. e.g. in MSVC it is with `/W4`
2. Compile an app that uses the `depthai-core-v2.11.1-win64` and simple code that cascades to headers like `RawFeatureTrackerConfig.hpp`. Most any code, even just a dai device, will do this.

### Result

```
...
[build] C:\repos-nobackup\depthai-core-v2.11.1-win64\include\depthai-shared/datatype/RawFeatureTrackerConfig.hpp(99): warning C4305: 'initializing': truncation from 'double' to 'float'
[build] C:\repos-nobackup\depthai-core-v2.11.1-win64\include\depthai-shared/datatype/RawFeatureTrackerConfig.hpp(105): warning C4305: 'initializing': truncation from 'double' to 'float'
[build] C:\repos-nobackup\depthai-core-v2.11.1-win64\include\depthai-shared/datatype/RawFeatureTrackerConfig.hpp(174): warning C4305: 'initializing': truncation from 'double' to 'float'
...
```

### Expected

No warnings or errors. No increase in exe size, no runtime convert double->float.

Luxonis-Brandon:
Thanks for the thorough report.  Brought up internally.  :-)

diablodale:
Also found truncation in depthai SDK itself (I'm getting closer to a full depthai sdk compile)
[...]

PR #268, opened 2021-11-17
correct float literals, 32/64 trunc, unref vars

diablodale:
The depthai-core portion of fixes for luxonis/depthai-core#248

Together with the already merged depthai-shared portion, they eliminate hundreds of warnings to lessen noise.

### DIFF
--- a/src/utility/matrixOps.hpp
+++ b/src/utility/matrixOps.hpp
@@ -43,12 +43,12 @@ std::vector<std::vector<float>> matMul(std::vector<std::vector<float>>& firstMat
  * Function to get cofactor of A[p][q] in temp. n is current
  * dimension of part of A that we are calculating cofactor for.
  */
-static void getCofactor(std::vector<std::vector<float>>& A, std::vector<std::vector<float>>& temp, int p, int q, int n) {
-    int i = 0, j = 0;
+static void getCofactor(std::vector<std::vector<float>>& A, std::vector<std::vector<float>>& temp, size_t p, size_t q, size_t n) {
+    size_t i = 0, j = 0;
 
     // Looping for each element of the matrix
-    for(int row = 0; row < n; row++) {
-        for(int col = 0; col < n; col++) {
+    for(size_t row = 0; row < n; ++row) {
+        for(size_t col = 0; col < n; ++col) {
             //  Copying into temporary matrix only those element
             //  which are not in given row and column
             if(row != p && col != q) {
@@ -58,7 +58,7 @@ static void getCofactor(std::vector<std::vector<float>>& A, std::vector<std::vec
                 // reset col index
                 if(j == n - 1) {
                     j = 0;
-                    i++;
+                    ++i;
                 }
             }
         }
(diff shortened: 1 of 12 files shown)


szabi-luxonis (approved):

szabi-luxonis (changes requested):
CI is failing due to style checks.
`clangformat` target should solve it, but I assume you don't have it installed on windows?

diablodale:
It is enabled in cmake, but such isn't available in typical Windows installations so cmake it likely quietly ignoring that target.

Alternatively, I used the VSCode c++ extension from Microsoft and have enabled. It does apply clang-format when I type a semicolon. But in many of these, I never typed a semi...I just inserted the casts inside the line. So it never formatted.
I del/add a semi on one of the lines lang-format reported and it reformatted it in the same way as the CI report.

I've turned on the additional option to clang-format on save of the file. Makes me a bit paranoid because I'll likely not notice the reformatting, but I get your team's need for this and I can work with that.
I'll resave the two files that the CI reported wrong and force push.

That all ok with u?

szabi-luxonis:
> It is enabled in cmake, but such isn't available in typical Windows installations so cmake it likely quietly ignoring that target.
> 
> Alternatively, I used the VSCode c++ extension from Microsoft and have enabled. It does apply clang-format when I type a semicolon. But in many of these, I never typed a semi...I just inserted the casts inside the line. So it never formatted. I del/add a semi on one of the lines lang-format reported and it reformatted it in the same way as the CI report.
> 
> I've turned on the additional option to clang-format on save of the file. Makes me a bit paranoid because I'll likely not notice the reformatting, but I get your team's need for this and I can work with that. I'll resave the two files that the CI reported wrong and force push.
> 
> That all ok with u?

Yes, that's fine.

diablodale:
🤔 the CI here isn't getting the forcepush I did. Its available at https://github.com/diablodale/depthai-core/tree/fix248-trunc-2 but it isn't showing up on the tabs here (and therefore not reprocessing it)

themarpe:
Reran the checks - will merge afterwards :+1:

src/utility/matrixOps.hpp
```diff
@@ -136,8 +136,8 @@ bool matInv(std::vector<std::vector<float>>& A, std::vector<std::vector<float>>&
 
     std::vector<float> temp;
     // Find Inverse using formula "inverse(A) = adj(A)/det(A)"
-    for(size_t i = 0; i < A.size(); i++) {
-        for(size_t j = 0; j < A.size(); j++) {
+    for(size_t i = 0; i < A.size(); ++i) {
```
alex-luxonis, line 139: Just curious on the `i++` -> `++i` changes, were there any warnings triggered by the previous version, or it's more like a style preference?

szabi-luxonis: Porbably style, pre-increment is faster than post-increment, theoretically, but not for POD.

diablodale: pre is my preferred for best opportunity for optimization, see notes https://en.cppreference.com/w/cpp/language/operator_incdec
No guarantee compilers old/new on all platforms won't use the slow path on POD. So I prefer pre.

I did not do it everywhere i that matrix code as some places need the post behavior.

FYI, the Eigen library is fantastically excellent and all header. Even for light use like cali processing. They've been testing  and optimizing across years and more platforms than most anyone could redo.

PR #274, opened 2021-11-23
Added stereo post processing filters

szabi-luxonis:
Implemented filters:
- Spatial
- Temporal
- Speckle
- Decimation
- Threshold

Added runtime config for `disparity/depth aligment`.

Improved on-device resource allocation logic.

Added presets for stereo settings: `HIGH_ACCURACY` or `HIGH_DENSITY`.
`HIGH_ACCURACY` is suitable for point cloud, while `HIGH_DENSITY` for visualization.

Implemented disparity/depth alignment to `RECTIFIED_LEFT` perspective without additional performance penalty.

Related PR:
https://github.com/luxonis/depthai-shared/pull/70


src/pipeline/node/StereoDepth.cpp
```diff
@@ -171,5 +171,23 @@ float StereoDepth::getMaxDisparity() const {
...
+        case PresetMode::HIGH_ACCURACY: {
+            initialConfig.setConfidenceThreshold(200);
+            initialConfig.setLeftRightCheck(5);
+        } break;
+        case PresetMode::HIGH_DENSITY: {
+            initialConfig.setConfidenceThreshold(245);
+            initialConfig.setLeftRightCheck(10);
+        } break;
+    }
+}
```
themarpe, line 190: Nice - lets call this in constructor as well? And having some default PresetMode that gets applied (if it doesn't otherwise)

themarpe (approved):
LGTM

PR #289, opened 2021-12-01
Added clock

themarpe:
Might be better to name it more specifically and to use it everywhere else as well.

Main intention is to expose it for Python side for now


szabi-luxonis (approved):
Style check is failing, otherwise LGTM.

Issue #264, opened 2021-11-16
feature request: `DEPTHAI_SHARED_LOCAL` `DEPTHAI_BOOTLOADER_SHARED_LOCAL`


diablodale:
I request a feature to allow easy dev/debug of the two submodules `depthai-shared` and `depthai-bootloader-shared`

Benefit is I can store separate git repos for `depthai-shared` on my drive, point depthai-core to it, make commits to that xxx-shared repo, cmake build, push to github, etc.  All without having to deal with the git submodule configuration.

Perhaps using a similar approach to the already existing `DEPTHAI_XLINK_LOCAL`. That works very well for me.
🤔 Is it a trivial change of the main cmakefiles.txt to have two option(), and then check those cmake vars at the top of the two `shared/submodule.cmake` files to set each their `xxx_SHARED_FOLDER` vars?

themarpe:
@diablodale should mostly be only these changes, although make sure to check git commit code, which makes sure that FW & shared match. I assume that for dev use, it'll just pass even if its pointed to a non git folder, but something to check still.

And installation takes these into account as well

PR #296, opened 2021-12-07
add cmake vars for local depthai-bootloader/shared

diablodale:
Fixes luxonis/depthai-core#264 by adding two optional cmake vars `DEPTHAI_SHARED_LOCAL` and `DEPTHAI_BOOTLOADER_SHARED_LOCAL`
When either or both are set, they will cause cmake to use the given directories instead of the submodules

Allows for easy coding/debugging of those two components/repos by optional separation


themarpe (approved):
LGTM - Thanks!

Issue #272, opened 2021-11-22
XLink not installed with `cmake install` (only static issue?)


diablodale:
For a static build of depthai, using `DEPTHAI_XLINK_LOCAL`... a cmake install
* does not install the XLink cmake files 😕
* does not install the XLink static libraries 😕
* does install the xlink includes 🙂

### Setup

* Microsoft Windows [Version 10.0.19044.1348]
* VS2019 Community v16.11.7
* cmake v3.19.1
* depthai-core v2.12.1 + my local fixes

### Repro

1. set depthai for a static build, x64, and opencv support
2. set depthai to use a local xlink with `DEPTHAI_XLINK_LOCAL`. Use the same commit as hunter would.
3. set depthai to install to a known location e.g. `CMAKE_INSTALL_PREFIX=c:/myinstalldir`
4. cmake configure, build, and then install

### Result

Either try to find_package() and then compile an app that would use this depthai install
-or-
search `c:/myinstalldir` for the xlink cmake file `XLinkConfig.cmake` and its static libraries `XLink.lib`/`XLinkd.lib`
You will find none of these files.

### Expected

Your app finds depthai and all its dependencies (like XLink) and cleanly builds.
-or-
You find XLink's cmake files and static libraries in the given install location `c:/myinstalldir`

### Notes

I see all the needed files in the depthai build directory `...\depthai-core\build\XLink`. This suggests their might be only a small change needed for cmake install process to use those and then do the install.

diablodale:
Hacking some on this in depthaiDependencies.cmake (probably wrong approach)...
```cmake
# XLink
if(NOT CONFIG_MODE)
    if(DEPTHAI_XLINK_LOCAL)
        add_subdirectory("${DEPTHAI_XLINK_LOCAL}" ${CMAKE_CURRENT_BINARY_DIR}/XLink EXCLUDE_FROM_ALL)
    else()
        find_package(XLink ${_QUIET} CONFIG REQUIRED)
    endif()
    if(NOT BUILD_SHARED_LIBS)
        list(APPEND targets_to_export XLink)
    endif()
endif()
```

I've got a few steps further, but app compile fails with depthai because it can't find `XLink/XLinkPublicDefines.h`
That header was not cmake-installed by depthai yet both `XLinkConnection.hpp` and `XLinkStream.hpp` need that include themselves.

diablodale:
I've got a local hack almost working. I think are three things in play...

[...]

PR #298, opened 2021-12-08
fix xlink cmake install for local, shared, and static

diablodale:
fixes luxonis/depthai-core#272

The default behavior of depthai-core is to use xlink via hunter. And that always creates an xlink static library -- it ignores the setting for `BUILD_SHARED_LIBS` for depthai-core.xxx itself.

Therefore, when a local xlink is used, I cause the same behavior by persisting that setting to `BUILD_SHARED_LIBS_SAVED` before adding the local xlink, and restoring the setting after.


cmake/depthaiDependencies.cmake
```diff
@@ -54,10 +54,18 @@ find_package(Threads ${_QUIET} REQUIRED)
...
-if(DEPTHAI_XLINK_LOCAL)
-    add_subdirectory("${DEPTHAI_XLINK_LOCAL}" ${CMAKE_CURRENT_BINARY_DIR}/XLink EXCLUDE_FROM_ALL)
-else()
-    find_package(XLink ${_QUIET} CONFIG REQUIRED)
+if(NOT CONFIG_MODE)
+    if(DEPTHAI_XLINK_LOCAL)
+        set(BUILD_SHARED_LIBS_SAVED "${BUILD_SHARED_LIBS}")
+        set(BUILD_SHARED_LIBS OFF)
+        add_subdirectory("${DEPTHAI_XLINK_LOCAL}" ${CMAKE_CURRENT_BINARY_DIR}/XLink)
+        set(BUILD_SHARED_LIBS "${BUILD_SHARED_LIBS_SAVED}")
```
themarpe, line 62: Just a 2c - `_BUILD_SHARED_LIBS_SAVED` and after setting it back, remove it: `set(_BUILD_SHARED_LIBS_SAVED)`

themarpe (changes requested):
Otherwise LGTM - Thanks

cmake/depthaiDependencies.cmake
diablodale: So I understand...this is a style thing. That underscore means "temporary use"?
And I can use `unset(_BUILD_SHARED_LIBS_SAVED)`?

themarpe: "Temporary/private" in our case - not actually sure about "in general" in CMake
Oh, yeah, `unset` will be a bit more explicit hah.

diablodale: Got it, thanks for guidance. Force pushing update...

cmake/depthaiDependencies.cmake
```diff
@@ -54,10 +54,19 @@ find_package(Threads ${_QUIET} REQUIRED)
...
+        set(BUILD_SHARED_LIBS OFF)
+        add_subdirectory("${DEPTHAI_XLINK_LOCAL}" ${CMAKE_CURRENT_BINARY_DIR}/XLink)
+        set(BUILD_SHARED_LIBS "${_BUILD_SHARED_LIBS_SAVED}")
+        unset(_BUILD_SHARED_LIBS_SAVED)
+    else()
+        find_package(XLink ${_QUIET} CONFIG REQUIRED)
+    endif()
+    if(NOT BUILD_SHARED_LIBS)
+        list(APPEND targets_to_export XLink)
+    endif()
```
themarpe, line 69: Merged a tad too soon...
This had to go in the `if(DEPTHAI_XLINK_LOCAL)` as it cannot be exported if wasn't built by the current project.
Pushed a fix to `develop`

diablodale: Have you tested that change with a `cmake install`?
Trying to remember...I think it is where I put it so that cmake install works correctly. The main cmakefile.txt uses the export list as the list to install. So even in the case where xlink is NOT local, it still needs to install all the xlink files.

diablodale: https://github.com/luxonis/depthai-core/blob/ab89e8b0cbd372e79a7b7f542af15bd8473e7d3c/CMakeLists.txt#L544-L546

themarpe: Reran the CI, there still seems to be some issues.
(Seems like related to the `NOT CONFIG_MODE` check - in that case find_package isn't called...)

> So even in the case where xlink is NOT local, it still needs to install all the xlink files.

If XLink isn't local, the files are installed when installing "Hunter dependencies"
```
# Install Hunter dependencies
install(DIRECTORY "${HUNTER_INSTALL_PREFIX}/" DESTINATION "${CMAKE_INSTALL_LIBDIR}/cmake/${PROJECT_NAME}/dependencies")
```

If it is local, then adding it to the export list will make it findable by consumers of core I assume.

themarpe: Was the
```
if(NOT CONFIG_MODE)
```
Intentionally added to resolve some issue case?

themarpe: @diablodale can you recheck with latest `develop` if it still works for you as intended?
Fixed the issue in config mode

diablodale: I'm 95% confident your current code will fail.

`if(NOT CONFIG_MODE)` is needed when the main app (the user of depthai-core) does ITS cmake config.
Scroll up to line 1 of the same file. You have to skip all the xlink add_sub/find_package when the main app does its cmake config.

diablodale: meanwhile, what is the specific setup scenario that causes the failure with my original PR code?

themarpe: > meanwhile, what is the specific setup scenario that causes the failure with my original PR code?

Regular compilation without specifying `DEPTHAI_XLINK_LOCAL` failed, because it tried exporting non project built XLink. https://github.com/luxonis/depthai-core/runs/4507551249?check_suite_focus=true

> You have to skip all the xlink add_sub/find_package when the main app does its cmake config.

When a main app calls `find_package(depthai CONFIG)` to consume it, the `find_package(XLink)` among others will be called which is correct as its needed. Not sure what you have in mind actually. It works in major usecases: https://github.com/luxonis/depthai-core/actions/runs/1574320534
 - Consumed by example from same project (just target)
 - Consumed by `add_subdirectory`
 - Consumed by `find_package`

But not sure about your use - I think it should. Let me know if it causes issues

diablodale: Unfortunately, I was correct...your code at `develop` fails when an app runs its cmake config.
```
[cmake] -- Found Threads: TRUE  
[cmake] -- Could NOT find libdw (missing: LIBDW_LIBRARY LIBDW_INCLUDE_DIR) 
[cmake] -- Could NOT find libbfd (missing: LIBBFD_LIBRARY LIBDL_LIBRARY LIBDL_INCLUDE_DIR) 
[cmake] -- Could NOT find libdwarf (missing: LIBDWARF_LIBRARY LIBDWARF_INCLUDE_DIR LIBELF_LIBRARY LIBELF_INCLUDE_DIR LIBDL_LIBRARY LIBDL_INCLUDE_DIR) 
[cmake] CMake Error at C:/njs/vcpkg/scripts/buildsystems/vcpkg.cmake:788 (_find_package):
[cmake]   Could not find a package configuration file provided by "XLink" with any of
[cmake]   the following names:
[cmake] 
[cmake]     XLinkConfig.cmake
[cmake]     xlink-config.cmake
[cmake] 
[cmake]   Add the installation prefix of "XLink" to CMAKE_PREFIX_PATH or set
[cmake]   "XLink_DIR" to a directory containing one of the above files.  If "XLink"
[cmake]   provides a separate development package or SDK, be sure it has been
[cmake]   installed.
[cmake] Call Stack (most recent call first):
[cmake]   C:/njs/depthai-core/.install/Debug/lib/cmake/depthai/depthaiDependencies.cmake:67 (find_package)
[cmake]   C:/njs/depthai-core/.install/Debug/lib/cmake/depthai/depthaiConfig.cmake:14 (include)
[cmake]   C:/njs/vcpkg/scripts/buildsystems/vcpkg.cmake:788 (_find_package)
[cmake]   cmake/find-oak.cmake:9 (find_package)
[cmake]   CMakeLists.txt:78 (include)
[cmake] 
[cmake] 
[cmake] -- Configuring incomplete, errors occurred!
```

You must skip that section during config mode...just like it is skipped on line 1.

Tomorrow, I'll investigate the scenario you describe above using my original PR code and come back with my investigation results.

themarpe: I'm not sure what you mean with same as on line 1 - also, does it not fail for you to find `nlohmann_json` before that? (I assume as you use vcpkg, that you have that one downloaded already, which is why it works?)

XLink should be found by `CMAKE_PREFIX_PATH` being appended with the path of `dependencies` which are installed alongside. If that is ignored, then XLink won't be able to be found.

Also on the topic of custom toolchains, there is `xlink_device_search_improvements` branch (`custom.toolchain.in` and related changes in main `CMakeLists.txt`), which you could checkout regarding custom toolchain usage (takes any toolchain and generates an intermediate one combined), was required for Android, but seems to work pretty well for other cases as also (YMMV).

diablodale:
@themarpe , hi. I have a fix ready. But this PR has been closed. Should I forcepush here? Or open a new PR?

themarpe:
Great @diablodale - please open a new PR, I think will be better from CI standpoint

PR #305, opened 2021-12-17
NNData serialize fix

themarpe:
Adds required metadata about tensors.

TODO later, add a proper way of setting tensors, that retain dimensions, striding, etc...


szabi-luxonis (approved):
We really need to go through stale branches and clean them up.
LGTM

PR #322, opened 2022-01-07
Add spatialLocationCalculator output message to spatial detection network

szabi-luxonis:
Added `spatialLocationCalculatorOutput` message, which carries the underlying `SpatialLocationCalculator` node output for `SpatialDetectionNetwork`


themarpe:
Would it be more suitable/easier to consume to perhaps add this infromation to SpatialImgDetections?

szabi-luxonis:
> Would it be more suitable/easier to consume to perhaps add this infromation to SpatialImgDetections?

Easier yes, but not more suitable, since they don't belong semantically to SpatialImgDetections

themarpe (approved):

Issue #334, opened 2022-01-16
incompatible BUILD_SHARED_LIBS + WINDOWS_EXPORT_ALL_SYMBOLS + INTERPROCEDURAL_OPTIMIZATION


diablodale:
On Windows, all Release builds fail when `BUILD_SHARED_LIBS=ON` and `CMAKE_INTERPROCEDURAL_OPTIMIZATION=ON`.

I think because CMakeLists.txt line 15 sets `CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS=ON`.
And that `...EXPORT_ALL_SYMBOLS=ON` feature of cmake is incompatible with `CMAKE_INTERPROCEDURAL_OPTIMIZATION=ON`.

When both are `ON` and `BUILD_SHARED_LIBS=ON` then the `exports.def file` created by the `CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS` feature is empty. With no def file, nothing is exported from the DLL and the linker fails.

Because depthai-core uses `CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS` instead of exporting via Windows standard approachs (e.g __declspec), it appears (from what I know so far) that the `INTERPROCEDURAL_OPTIMIZATION=ON` setting can not be used on Windows with the MSVC toolchain. FYI, `INTERPROCEDURAL_OPTIMIZATION=ON` cmake setting enables link-time code generation on MSVC with `/LTCG`.

### Setup

* Microsoft Windows [Version 10.0.19044.1466]
* VS2019 v16.11.9
* depthai-core `develop` 3d5dc2ee77b78ace7c20c49ec1ad2b13545aa4e6

### Repro

1. set `BUILD_SHARED_LIBS=ON`, `CMAKE_INTERPROCEDURAL_OPTIMIZATION=ON`, and `DEPTHAI_ENABLE_BACKWARD=ON`
2. cmake config for Release, x64, shared lib
3. cmake build

### Result

```
[...]

PR #337, opened 2022-01-17
fix Win MSVC cmake INTERPROCEDURAL_OPTIMIZATION

diablodale:
Workaround for Windows MSVC incompatibility with BUILD_SHARED_LIBS + WINDOWS_EXPORT_ALL_SYMBOLS + INTERPROCEDURAL_OPTIMIZATION
Fixes luxonis/depthai-core#334

Approach:  used the existing `add_default_flags()` function as the issue is about flags and that function was used in all the needed places.

I'm interested to see the CI test it across the matrix. 🤔
Passes all tests and examples on my Windows MSVC x64 build, debug+release, with settings...
```cmake
set(CMAKE_INTERPROCEDURAL_OPTIMIZATION ON CACHE BOOL "")
set(CMAKE_INTERPROCEDURAL_OPTIMIZATION_DEBUG OFF CACHE BOOL "")
```

### DIFF
--- a/cmake/Flags.cmake
+++ b/cmake/Flags.cmake
@@ -1,6 +1,8 @@
 ## setup compilation flags
-# conditionally applies flag. If flag is supported by current compiler, it will be added to compile options.
+include(CMakeParseArguments)
 include(CheckCXXCompilerFlag)
+
+# conditionally applies flag. If flag is supported by current compiler, it will be added to compile options.
 function(add_flag target flag)
     check_cxx_compiler_flag(${flag} FLAG_${flag})
     if (FLAG_${flag} EQUAL 1)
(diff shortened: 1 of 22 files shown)


diablodale:
These seem to be valid bugs surfaced in the tests and examples. The comprehensive fix is to fix the code in the tests+examples.
A workaround is to ignore those bugs and not apply the default flags to them which are currently surfacing valid issues.

Opinions? Thoughts?

themarpe:
@diablodale 
```
/home/runner/work/depthai-core/depthai-core/tests/src/bootloader_config_test.cpp:23:22: error: comparison of integer expressions of different signedness: ‘int’ and ‘std::array<unsigned char, 4>::size_type’ {aka ‘long unsigned int’} [-Werror=sign-compare]
   23 |     for(int i = 0; i < ipv4InMemory.size(); i++) {
      |                    ~~^~~~~~~~~~~~~~~~~~~~~
```

Can you fix the surfaced issues in the scope of this PR as well? Not sure if this is the only one, but seems minor.

diablodale:
> Can you fix the surfaced issues in the scope of this PR as well? Not sure if this is the only one, but seems minor.

I have the knowledge, though want to be careful with my time the next two weeks as I am in release crunch.
I do not have a Linux build harness setup. And I don't know if those Linux flags are/will report errors in depthai-core itself (not tests/examples)...how deep goes the rabbit hole🐇😆?

Does anyone on your team have a ready-made Dockerfile for building on Ubuntu? That can help speed my progress. If not, I'll need to make one today.

themarpe:
Thanks - as your time permits

> Does anyone on your team have a ready-made Dockerfile for building on Ubuntu? That can help speed my progress. If not, I'll need to make one today.

We have from `depthai-python`, but rechecked, it doesn't include native opencv deps which would be needed for examples...

I'll test locally if that change is the only one and circle back.

themarpe:
Using default flags on examples pops up way to many warnings & errors. 

I think it would be better for sake of simplicity of examples to not have those flags set to such high levels.

diablodale:
I suspected 🐛🦶.
Are you seeing the error/warns from the SDK code, or are they only errors/warns in the cpp files of the tests+examples?
If the former...yes...there is sdk cleanup work to do ;-)
If the latter...it couldn't be more than a long day work; two max? There are only ~65 cpp files for tests+examples.

If not fix the errant code, then I could move this PR's fix out of `add_default_flags()` to something separate. Or add a param like `add_default_flags(BUGS_ONLY)` or `add_default_flags(FIXES_ONLY)`. Or...?

themarpe:
More the latter - but given the set of flags, it'd make the examples more obstructed than necessary (making portraying the actual point harder).

I lean towards leaving out the flags for it, and only applying the MSVC flags to resolve the mentioned issue.
Lets move it out to a specific function or as mentioned adding a parameter to go along to only deal with this kind of flags - latter sounds nice, if there is going to be more of this kinds of things to handle...

diablodale:
Would you start the CI again to test new changes?
I made a docker harness and fixed the test and example errors. Those are in the second commit.

I get no errors on my Ubuntu 20.04 docker. Limited in my ability to run tests without docker hardware access.
Windows compiled and tested good.

Warnings are still there like all main apps linking to depthai experience.

diablodale:
Oy. One more time on the CI please.
I learned that gcc didn't support all the declared switches. So now I can also compile with clang13 and made the one fix. 🙄😂

themarpe:
Sorry for the delay - I'd just modify that examples don't get all the flags, but just the ones required for the fix - to not spam with warnings in examples. Maybe `add_default_flags(target [LEAN/...])` with additional optional parameter or a separate function that only adds required flags

Otherwise LGTM

diablodale:
Would you start CI?
I force pushed update that enables the optional flag idea as `add_default_flags(target [LEAN])` and updated test+example cmake's to use `LEAN`

PR #361, opened 2022-02-01
Rgbd-depth alignment optimization

OanaMariaVatavu:
**Adding** 

1. RGB-depth alignment improvements for 
     - U8 case (standard LR-check)
                         -  30+ FPS at full resolution 
     -  U16 case (LR-check + subpixel)
                         -  22 FPS at THE_720_P resolution - limited by subpixel calculation
                         -  20 FPS at THE_800_P resolution - limited by subpixel calculation
2. [fix] disparity/depth shift factor determined by calibration data (before it was center for all cases) for boards having not-in-center RGB sensor(e.g. OAK-D-IoT-40, OAK-D-CM3 , OAK-D-CM4, OAK-D-CM4-PoE)

**Modify**
- stereo resolution THE_400_P -> THE_720_P for the rgb_depth_aligned.py example

**[ToDo]** 
- confidence map - not optimized yet, will follow up in a different PR


szabi-luxonis (approved):
LGTM, update device side commit after FW PR is merged.

PR #368, opened 2022-02-08
OpenVINO Blob structure

themarpe:
Added reader and Blob structure to be able to query network details before using it in a pipeline.
Information include:
 - map of IO /w names, dimensions, datatype, order, ...
 - num shaves
 - num slices
 - stage count
 
Added some light tests for this functionality.

Also added `NeuralNetwork::setBlob` which takes a blob instead of a path

Note: moved `TensorInfo` to `common/` under `depthai-shared`, as its now used in more places

### DIFF
--- a/include/depthai/openvino/OpenVINO.hpp
+++ b/include/depthai/openvino/OpenVINO.hpp
@@ -4,8 +4,11 @@
 #include <exception>
 #include <map>
 #include <string>
+#include <unordered_map>
 #include <utility>
 #include <vector>
+
+#include "depthai-shared/common/TensorInfo.hpp"
 
 namespace dai {
 
@@ -14,6 +17,29 @@ class OpenVINO {
    public:
     /// OpenVINO Version supported version information
     enum Version { VERSION_2020_3, VERSION_2020_4, VERSION_2021_1, VERSION_2021_2, VERSION_2021_3, VERSION_2021_4 };
+
+    /// OpenVINO Blob
+    struct Blob {
+        /**
+         * @brief Construct a new Blob from data in memory
+         *
+         * @param data In memory blob
+         */
+        Blob(std::vector<uint8_t> data);
+        /**
+         * @brief Construct a new Blob by loading from a filesystem path
+         *
+         * @param path Filesystem path to the blob
+         */
+        Blob(const std::string& path);
+        Version version;
+        std::unordered_map<std::string, TensorInfo> networkInputs;
+        std::unordered_map<std::string, TensorInfo> networkOutputs;
+        uint32_t stageCount = 0;
+        uint32_t numShaves = 0;
+        uint32_t numSlices = 0;
+        std::vector<uint8_t> data;
+    };
 
     /// Main OpenVINO version
     constexpr static const Version DEFAULT_VERSION = VERSION_2021_4;
--- a/include/depthai/openvino/OpenVINO.hpp
+++ b/include/depthai/openvino/OpenVINO.hpp
@@ -32,12 +32,20 @@ class OpenVINO {
          * @param path Filesystem path to the blob
          */
         Blob(const std::string& path);
+
+        /// OpenVINO version
         Version version;
+        /// Map of input names to additional information
         std::unordered_map<std::string, TensorInfo> networkInputs;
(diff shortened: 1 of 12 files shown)


src/openvino/BlobReader.cpp
```diff
@@ -47,108 +53,73 @@ void BlobReader::parse(const std::vector<std::uint8_t>& blob) {
...
-    //
-    //     ie::TensorDesc ieDesc = DataDesc(dataType, dimsOrder, vpuDims).toTensorDesc();
-    //     ie::Data outputData(outputName, ieDesc);
-    //
-    //     _networkOutputs[outputData.getName()]    = std::make_shared<ie::Data>(outputData);
-    //     _outputInfo.offset[outputData.getName()] = ioBufferOffset;
-    // }
+    const auto readIO = [this, &blob](uint32_t& ioSectionOffset, uint32_t idx) {
+        auto ioIdx = readFromBlob<uint32_t>(blob, ioSectionOffset);
+        if(ioIdx != idx) {
```
szabi-luxonis, line 58: read also the file length encoded in blob and compare against the loaded size from filesystem

szabi-luxonis (changes requested):

src/openvino/BlobReader.cpp
szabi-luxonis: It can be that the blob is corrupted, was interrupted while writing/downloading etc.

include/depthai/openvino/OpenVINO.hpp
```diff
@@ -15,6 +18,29 @@ class OpenVINO {
...
+         */
+        Blob(std::vector<uint8_t> data);
+        /**
+         * @brief Construct a new Blob by loading from a filesystem path
+         *
+         * @param path Filesystem path to the blob
+         */
+        Blob(const std::string& path);
+        Version version;
+        std::unordered_map<std::string, TensorInfo> networkInputs;
```
szabi-luxonis, line 36: document these fields

szabi-luxonis (changes requested):

include/depthai/pipeline/node/NeuralNetwork.hpp
```diff
@@ -65,6 +65,14 @@ class NeuralNetwork : public NodeCRTP<Node, NeuralNetwork, NeuralNetworkProperti
      */
     void setBlobPath(const std::string& path);
 
+    /**
+     * Load network blob into assets and use once pipeline is started.
+     *
+     * Throws if blob isn't valid
```
szabi-luxonis, line 71: Throws error

src/openvino/BlobReader.cpp
```diff
@@ -47,108 +55,77 @@ void BlobReader::parse(const std::vector<std::uint8_t>& blob) {
...
-    //     }
-    //
-    //     ie::TensorDesc ieDesc = DataDesc(dataType, dimsOrder, vpuDims).toTensorDesc();
-    //     ie::Data outputData(outputName, ieDesc);
-    //
-    //     _networkOutputs[outputData.getName()]    = std::make_shared<ie::Data>(outputData);
-    //     _outputInfo.offset[outputData.getName()] = ioBufferOffset;
-    // }
+    if(blob.size() < blobHeader.file_size) {
+        throw std::logic_error("BlobReader error: Filesize is less than blob specifies. Likely corrupted");
```
szabi-luxonis, line 59: runtime error instead logic error ? Or even better `std::length_error` ?

src/openvino/BlobReader.cpp
```diff
@@ -24,14 +26,20 @@ namespace {
 
 template <typename T>
 T readFromBlob(const std::vector<std::uint8_t>& blob, uint32_t& offset) {
-    assert(offset + sizeof(T) <= blob.size());
+    if(offset + sizeof(T) > blob.size()){
+        throw std::logic_error("BlobReader error: Filesize is less than blob specifies. Likely corrupted");
```
szabi-luxonis, line 30: length_error?

szabi-luxonis (changes requested):

szabi-luxonis (approved):

PR #370, opened 2022-02-09
Add DEPTHAI_DEVICE_MXIDS env variable support

jdavidberger:
This commit adds support for a 'DEPTHAI_DEVICE_MXIDS' environment variable which limits the visibility of devices enumerated. The use case is that when dealing with multiple cameras attached to the same computer, it can be useful to have fine control over which one(s) are invokable in existing scripts and applications which weren't necessarily written with multiple camera support in mind.

### DIFF
--- a/src/xlink/XLinkConnection.cpp
+++ b/src/xlink/XLinkConnection.cpp
@@ -91,7 +91,7 @@ std::vector<DeviceInfo> XLinkConnection::getAllConnectedDevices(XLinkDeviceState
         states = {state};
     }
 
-    auto allowedDeviceIDs = spdlog::details::os::getenv("DEPTHAI_DEVICE_MXIDS");
+    auto allowedDeviceIDs = utility::getEnv("DEPTHAI_DEVICE_MXID_LIST");
 
     // Get all available devices (unbooted & booted)
     for(const auto& state : states) {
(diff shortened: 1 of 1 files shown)


src/xlink/XLinkConnection.cpp
```diff
@@ -70,6 +70,8 @@ std::vector<DeviceInfo> XLinkConnection::getAllConnectedDevices(XLinkDeviceState
         states = {state};
     }
 
+    auto allowedDeviceIDs = spdlog::details::os::getenv("DEPTHAI_DEVICE_MXIDS");
```
themarpe, line 73: I think could be named as `DEPTHAI_DEVICE_MXID_LIST` and later `DEPTHAI_DEVICE_NAME_LIST` (with updated XLink library)

themarpe, line 73: Also replace with `utility::getEnv("DEPTHAI_DEVICE_MXID_LIST")`

themarpe, line 73:
```
auto allowedDeviceIds = utility::getEnv("DEPTHAI_DEVICE_MXID_LIST");
```

themarpe (changes requested):
Modify the PR to target `develop` and run `clangformat`, seems to be some indentation changes

themarpe:
Thanks!

themarpe:
Merged in 9cd0b9748cad842dbcd2b8d859db00b1856722c5

Issue #366, opened 2022-02-06
feature req: `DataOutputQueue` with `maxSize=0` --> received data only sent to callbacks


diablodale:
I request the `DataOutputQueue` support `maxSize=0` so that no data is retained in the internal `std::queue` yet the callbacks continue to be called with the data.

I think the change is trivial and isolated to `LockingQueue.hpp`. I can take on this work as it already has use potential in my app.

### Setup

* all

### Repro

By code review. `LockingQueue::push()` and `tryWaitAndPush()` don't behave appropriately when maxSize=0.
If maxSize=0, they would actually behave as if it was =1.
https://github.com/luxonis/depthai-core/blob/c2ccafd5be7c2b0624addaca62f6c52935c9bdc5/include/depthai/utility/LockingQueue.hpp#L107-L125

A one-line change to check for `if (maxSize == 0) return true` would correct the behavior and support this feature request. I think `true` is the appropriate return since this is not an error condition and the function would have placed the data into the queue. It just so happens that a maxSize=0 queue is a singularity with size 0... or like /dev/null.

And removing the one line check here
https://github.com/luxonis/depthai-core/blob/c2ccafd5be7c2b0624addaca62f6c52935c9bdc5/include/depthai/utility/LockingQueue.hpp#L22

diablodale:
While updating the code, those same two `push` functions have an inconsistency and threading issue.

* blocking: failed attempt to check for `destructed`, but there is no `lock(guard)` in `destruct()` so this check is not thread safe. Another thread could set `destructed=true` using`::destruct()` a nanosecond after the `if` check and negate this attempted logic. This is a race condition.
* non-blocking: never does the check. Making it inconsistent in a `destructed` check and likely a marginal amount less thread-safe.

Then for both cases, it is possible for one thread to be "on the path" to destruct the class instance while another thread is calling `LockingQueue::push()`. This happens in `dai::DataInputQueue` and `dai::DataOutputQueue` classes. Their "reading/writing" thread could be calling e.g. `LQ:push()` or `LQ::tryPop()` at the same time the main app thread is on the path to destruct a `DataQueue`. This is a *very* common scenario.

In the two `dai::DataQueue` classes, I think there is sufficient locking to prevent crashes. 👍

[...]

PR #378, opened 2022-02-15
enable queues of size=0 which only do callbacks

diablodale:
Add support for Queues used by depthai-core to have size=0 and therefore only call callbacks.

- minor optimize LockingQueue
- fixes luxonis/depthai-core#366

I've been using this for about a week; size=0 works well in my app.
Passes test+examples, though they don't actively test size=0.

### DIFF
--- a/include/depthai/utility/LockingQueue.hpp
+++ b/include/depthai/utility/LockingQueue.hpp
@@ -19,7 +19,6 @@ class LockingQueue {
 
     void setMaxSize(unsigned sz) {
         // Lock first
-        if(sz == 0) throw std::invalid_argument("Queue size can't be 0!");
         std::unique_lock<std::mutex> lock(guard);
         maxSize = sz;
     }
@@ -57,7 +56,6 @@ class LockingQueue {
 
             // First checks predicate, then waits
             bool pred = signalPush.wait_for(lock, timeout, [this]() { return !queue.empty() || destructed; });
-            if(destructed) return false;
             if(!pred) return false;
 
             // Continue here if and only if queue has any elements
@@ -76,7 +74,6 @@ class LockingQueue {
             std::unique_lock<std::mutex> lock(guard);
 
             signalPush.wait(lock, [this]() { return !queue.empty() || destructed; });
-            if(destructed) return false;
             if(queue.empty()) return false;
 
             while(!queue.empty()) {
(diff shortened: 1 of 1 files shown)


include/depthai/utility/LockingQueue.hpp
```diff
@@ -129,6 +132,13 @@ class LockingQueue {
     bool tryWaitAndPush(T const& data, std::chrono::duration<Rep, Period> timeout) {
         {
             std::unique_lock<std::mutex> lock(guard);
+            if(!maxSize) {
```
themarpe, line 135: `if(maxSize == 0) {`

include/depthai/utility/LockingQueue.hpp
```diff
@@ -129,6 +132,13 @@ class LockingQueue {
...
+                // necessary if maxSize was changed
+                while(queue.size()) {
```
themarpe, line 137: `while(!queue.empty()) {`

include/depthai/utility/LockingQueue.hpp
```diff
@@ -108,6 +105,13 @@ class LockingQueue {
     bool push(T const& data) {
         {
             std::unique_lock<std::mutex> lock(guard);
+            if(!maxSize) {
```
themarpe, line 108: `if(maxSize == 0) {`

include/depthai/utility/LockingQueue.hpp
```diff
@@ -108,6 +105,13 @@ class LockingQueue {
...
+                // necessary if maxSize was changed
+                while(queue.size()) {
```
themarpe, line 110: `while(!queue.empty()) {`

themarpe (approved):
Just some readability nitpicks
Otherwise LGTM

include/depthai/utility/LockingQueue.hpp
diablodale: hahaha. My love of the boolean behavior with int again.
I'll make the change

include/depthai/utility/LockingQueue.hpp
diablodale: I'll make the change. 👍 And you can tell me in the future to make similar changes.

Caution (not arguing), you will likely see me using `!` within boolean expressions very often as the standard is clear on behavior with int's and float's, and works with a class with operator bool(), etc. I will subconsciously almost always write the logic that way as I'll likely forget this project's style in the moment my fingers are typing. ⌨🤖

diablodale:
I force pushed the requested changes. Would you start CI?

Issue #380, opened 2022-02-15
examples FAIL_REGULAR_EXPRESSION regex for warning triggers incorrectly with DEPTHAI_LEVEL=warn


diablodale:
More than 2/3 of tests+examples fail with current `develop` when logging is set to `WARN` level.
Why? To see warnings from depthai-core, the envvar (or api) has to be set to the warning level or worse. For example
```
set DEPTHAI_LEVEL=warn
```

The example testing cmake/harness recognizes as a test failure using regex `"\\[warning\\];\\[error\\];\\[critical\\]"`
But setting the env var *itself* causes a warning.
```
[2022-02-15 05:01:40.171] [warning] Environment 'DEPTHAI_LEVEL' set to 'warn'
```

But you can't see warnings unless you set it to warn (or worse). 
But setting it to warn causes a warning...which causes the test to be marked failed.
This is a circular issue 😵 caused by commit bd11775f2c747766900ed244f85f0ac295b807c9

### Setup

* All
* depthai-core `develop` 7bbebae2a54bf6879b4c4f3e4c504ed13dde8408

### Repro

1. cmake config, build for debug with all tests and examples
2. set your env var `DEPTHAI_LEVEL=warn`
3. Run ctest for all tests and examples

### Result

Mass failure
```
[ctest] 27% tests passed, 59 tests failed out of 81
[ctest] 
[ctest] Total Test time (real) = 982.53 sec
[ctest] 
[ctest] The following tests FAILED:
[ctest] 	 23 - bootloader_version (Failed)
[ctest] 	 24 - rgb_camera_control (Failed)
[ctest] 	 25 - rgb_preview (Failed)
[ctest] 	 26 - rgb_video (Failed)
[ctest] 	 27 - edge_detector (Failed)
[ctest] 	 28 - feature_detector (Failed)
[ctest] 	 29 - feature_tracker (Failed)
[ctest] 	 30 - device_queue_event (Failed)
[ctest] 	 31 - opencv_support (Failed)
[ctest] 	 32 - queue_add_callback (Failed)
[ctest] 	 33 - image_manip (Failed)
[ctest] 	 34 - image_manip_rotate (Failed)
[ctest] 	 35 - image_manip_tiling (Failed)
[ctest] 	 36 - rgb_rotate_warp (Failed)
[ctest] 	 37 - imu_gyroscope_accelerometer (Failed)
[ctest] 	 38 - imu_rotation_vector (Failed)
[...]

PR #381, opened 2022-02-15
log envvar values at TRACE level only

diablodale:
Fixes luxonis/depthai-core#380

### DIFF
--- a/src/utility/Environment.cpp
+++ b/src/utility/Environment.cpp
@@ -23,7 +23,7 @@ std::string getEnv(const std::string& var) {
 
     // Warn if env variable set
     if(!value.empty()) {
-        spdlog::warn("Environment '{}' set to '{}'", var, value);
+        spdlog::trace("Environment '{}' set to '{}'", var, value);
     }
 
     return value;


themarpe:
Thanks @diablodale 
Can you modify to `debug` instead - `trace` is mostly for high occurrence logging
(Also, if someone might've set the env var and forgot, we'd catch in our default "Please run with `DEPTHAI_LEVEL=debug` and post the log", debugging first step)

themarpe (approved):

diablodale:
@themarpe What do you mean by DEBUG?

If you mean for a DEBUG build, then that doesn't much help the situation. For example, I regularly run the test suite on DEBUG builds. Its the *most* common way I run tests so that I have diagnostics to understand testing failures.

Or do you mean `spdlog::debug`. that log class would work for me. Though we could combine the two so that this output only occurs on DEBUG builds and at the `spdlog::debug` level.

diablodale:
Thanks for clarifying and sharing the intention for your support workflow. 👍

Issue #300, opened 2021-12-10
DeviceBootloader::getFirstAvailableDevice() can return unusable result


diablodale:
Calls to `DeviceBootloader::getFirstAvailableDevice()` can return a "found" boolean yet the deviceinfo returned has a name of `"<error>"`

Timing of the repro steps is important to trigger the error state.

### Setup

* Microsoft Windows [Version 10.0.19044.1348]
* VisStudio v.16.11.7
* `depthai-core` v2.13.3 + local ownership fixes
* OAK-D

### Repro

1. cmake config for static library, x64, debug + build all tests and examples
2. cmake build
3. make a BAT file that runs `tests\image_manip_node_test.exe` in a loop
4. run that BAT file and let it loop at least 5 times
5. press Ctrl-C while the EXE is running during a loop. The exact timing of the ctrl-c seems to matter. You may have to retry this several times until you get the failure state.
6. Notice `image_manip_node_test.exe` stops due to the ctrl-c, and you are back at the command prompt. You might have to type yes to end the BAT file.
7. Run `build\examples\bootloader_version.exe`

### Result

```
Found device with name: <error>
```
and the EXE is still running...seems to be blocked or in a loop somewhere.
I can press ctrl-c to exit the app. And then run `bootloader_version.exe` over and over... it always gives the error and blocks.
Even as I am typing this bug, easily 15 minutes has passed. And still it reports an error if I run the app. This makes me suspect a problem in USB/OS or a device-side problem.
If I look in Windows Device Manager during this error state and the app is not running, I see the oak-d appear as "Movidius MyriadX".

If I unplug the USB cable computer<->oak-d and reconnect it. The app reports "No devices found". And then several seconds later, Windows itself reported: USB device not recognized.
![image](https://user-images.githubusercontent.com/679350/145629040-c43faafc-32f6-48b7-8e21-b6973538210a.png)
I can plug/unplug and every time Windows immediately reports that error.

If I unplug the POWER from the OAK-D, then reconnect the power and usb....then bootloader_ver.exe works correctly again.

### Expected

I get the valid name and bootloader version of the device.
Or...
I get found=false
And...
`DeviceBootloader` constructor doesn't infinitely block

### Notes

When I create the repo and put `bootloader_version` in the debugger, that error output goes to console, and the app doesn't exit.

After the apps's call to `getFirstAvailableDevice()` the vars are
[...]

PR #382, opened 2022-02-15
skip devices named "<error>" in device search

diablodale:
partial fix luxonis/depthai-core#300

This is needed even after the watchdog work on host and device merged in Jan/Feb.
When device is in the error state and host is search for it, the name returned is `<error>` as in #300 and adds this unusable/corrupt device to the collection returned from the API.

With this PR, it will log an error, skip to the next possible device, and not include this errant device in the returned collection of devices. Without it, all callers in all apps have to write the same loop that does the same string compare to `<error>`.

Calling code now gets a correct collection of found devices. So if there was only one OAK hardware attached and it was this errant devices, the collection returned will be empty. A good thing.

Here is the log appearance. You can reproduce the error state and such logging using the steps in #300
```
[2022-02-15 22:13:44.117] [error] skipping X_LINK_UNBOOTED device having name "<error>"
```

Passes the full test+examples suite
```
[ctest] 100% tests passed, 0 tests failed out of 79
[ctest] 
[ctest] Total Test time (real) = 997.82 sec
[ctest] CTest finished with return code 0 
```

### DIFF
--- a/src/xlink/XLinkConnection.cpp
+++ b/src/xlink/XLinkConnection.cpp
@@ -108,6 +108,10 @@ std::vector<DeviceInfo> XLinkConnection::getAllConnectedDevices(XLinkDeviceState
 
         for(unsigned i = 0; i < numdev; i++) {
             DeviceInfo info = {};
+            if(std::strcmp("<error>", deviceDescAll.at(i).name) == 0) {
+                spdlog::error("skipping {} device having name \"{}\"", XLinkDeviceStateToStr(state), deviceDescAll.at(i).name);
+                continue;
+            }
             info.desc = deviceDescAll.at(i);
             info.state = state;
             bool allowedId = allowedDeviceIds.find(info.getMxId()) != std::string::npos || allowedDeviceIds.empty();


themarpe (approved):
I think this is adequate for now - however I'll also bring this into XLink, where we actually can get meaningful information out.
LGTM

I think we could also warn users to check for their udev rules. @Erol444 for some small informative message / direction to get those installed

diablodale:
> I think we could also warn users to check for their udev rules. @Erol444 for some small informative message / direction to get those installed

Humm, from that "support view" I think an error message here about udev could have a worsening affect.
The issue from which this PR and other work was spawned was all Windows. udev has nothing to do with Windows.
An error message about udev to your Windows python/c++ customers would cause customer confusion, frustration, and eventually a support escalation "how do I set my udev on Windows?". 😆

I see an opportunity for Luxonis to set a philosophy regarding textual messaging to customers on errors.
A way of language, grammar, and giving people relevant support info so they can solve their own issues faster.

For example... this PR currently outputs
```
skipping X_LINK_BOOTED device having name "<error>"
```

Perhaps it could instead have a link to support resource.
```
skipping X_LINK_BOOTED device having name "<error>" see https://help.luxonis.com/kb/63752
```

Then your support team can actively update and give correct support info for Linux, Windows, etc.

As I am writing this, I also now think `error` for specific scenario is too severe. I would suggest a `warn` instead since the system continues functioning well and can connect to any other device except that specific device...which is currently in a state that will self-reset in 5 seconds. That's not so bad...so I think a warn is better.

themarpe:
I like this approach, would be beneficial to uptake it one day.

With XLink improvements on the other branch, the ambiguity between not being able to talk with device vs not having permissions will be resolved and the udev / permissions warning will be in place. 

For now I think this suits to solve the issue at hand. Merging

Issue #352, opened 2022-01-26
paths as `std::string` is not safe or cross-platform -- e.g. fails on Windows with all non-ascii paths


diablodale:
All non-ascii paths on OS which are not natively UTF-8 are doomed to fail.
For example, `c:/MeinDeutschPäth` is likely to fail on Windows due to the umlaut.
Non-latin characters, multibyte glyphs, etc. will all fail on operating systems that do not natively default to utf-8.

One example at
https://github.com/luxonis/depthai-core/blob/0caf71c904606fb34116c66cc8e999e0ee17d863/src/pipeline/node/NeuralNetwork.cpp#L26-L27

Likely the cause of issue #294 because Chinese glyphs are definitely not the limited ascii.

### Setup

* all

### Repro

Found while writing my app and then code review. A `std::string` is not a utf-8 string. And it will fail on Windows for any path that is non-ascii. The best fix is to use `<filesystem>` which is available in C++17, or `<experimental::filesystem>` in most C++14 toolchains. The STL library will manage going to/from utf-8 with clearly defined params/funcs. For example, it will manage the differences between native utf-8 on Linux vs. wchar_t on Windows.

If one doesn't use the STL, then one needs to write supporting functions for Windows (perhaps others) to convert between Windows wchar_t UTF-16 and what appears to be the desire of `AssetManager` to have a utf-8 path. There are clear apis in the STL to manually convert between UTF-16 and UTF-8.

I see many locations throughout `depthai-core` which have this bug. For example `AssetManager::set(const std::string& key, const std::string& path, ...`
Also locations using pointers to char like `const char* pathToBootloader` which are insufficient for native UTF-16 usage on Windows. They will have the same path bugs

### Fix

It is not hard to use the STL cvt approaches rather than "filesystem". Pseudocode...
```c++
std::wstring_convert<std::codecvt_utf8_utf16<wchar_t>> utf16convert;
std::wstring widePath = utf16convert.from_bytes(myutf8path);
std::wifstream stream(widePath, std::ios::in | std::ios::binary);
```

### Workarounds

* There is no code workaround. The conversions must be done.
* A user could move the blobs, mesh, calibration data, and all other data defined via a path... to a path that has only ascii characters. This is quite restrictive, but it is a workaround.
* The Windows OS setting to use utf-8 suggested in #294 is not suggested and I discourage that approach due to the high risk compatibility issues it introduces across all apps and all os features.

themarpe:
[...]

PR #384, opened 2022-02-16
new class `dai::Path` for APIs that accept path/filenames

diablodale:
This is initial concept to fix luxonis/depthai-core#352
DO NOT MERGE IT IS NOT READY!

🤔I request feedback on approach, code, everything.

### Approach

A single class `dai::Path`. The first two below are strongly influenced by c++17's `std::filesystem::path`.

* accepts a variety of string-like things
* a single user-defined conversion function that outputs the path in a format compatible with ifstream/ofstream
* no member functions or public member fields
* implemented as a single header
* many test cases included

I used a header because its a simple class, no significant code, and encourages the compilers to inline and/or eliminate codepaths when there is no conversion work to do; this could be the case on posix utf-8 systems.

### Not yet done/tested

* Will this approach allow C++17 apps to pass `std::filesystem::path` to the APIs that accept `dai::Path` when the binary behind the APIs was compiled as C++14. I'm hoping it works (or close to working) since all the code is in the header, and the only class member storage is a simple std::string/wstring.
* Changing the depthai-code APIs to use `dai::Path` instead of the typical `std::string`.
* Only tested on Windows so far. 🧐 Would be fun to see what the matrix thinks ;-)


diablodale:
I think in any implementation approach, we have to deal with the path charset when it is displayed.
```c++
throw std::runtime_error(fmt::format("Cannot load blob, file at path {} doesn't exist.", path));
```

That code will fail to compile if code attempts to give a Unicode (wchar_t) string/path on Windows to `fmt`:
`error C2338: mixing character types is disallowed`

`fmt` requires the specification to be in the same charset. On Windows must prepend `L`
```c++
throw std::runtime_error(fmt::format(L"Cannot load blob, file at path {} doesn't exist.", path));
```

But then the result from `fmt` is a `std::wstring` and is not compatible with `std::runtime_error()` which on Windows only supports an 8-bit ascii string with `char*` or `std::string`

I believe there has to be depthai-core specific handler for the switching between `string` and `wstring` for visual outputs like logging.

1. My preference: write a `fmt` custom handler that will accept `dai::Path` and convert it to ascii. Of course, the result of this for any non-ascii path will be incorrect as ascii can not represent unicode. Alternatively, I think it is possible to detect when a conversion is incomplete and the handler returns default ascii text `<Unicode path can not be displayed>` and it is nicely inserted into the `fmt` result. I found several examples of a custom fmt formatters from which I can be inspired like https://github.com/gabime/spdlog/issues/1783. Code changes for this should be small and isolated.
2. Use the approach similar to Windows SDK. Have a `TEXT()` macro that is used around all strings that will be used for display/log which conditionally prepends `L` in Windows on strings so that it `fmt` will accept wchar_t strings. This is a lot of churn adding the macro everywhere and ensuring all lvalue string parameters to `fmt::format()` are wchar_t. Lots of little changes everywhere, requires ongoing macros/wrapping in these cases.

diablodale:
Easy code for a custom `fmt` formatter. Done and working on Windows, and updated one API `NN::setBlobPath()` to use this PRs approach.

What changes to depthai-core APIs? Change the param type on the function signatures from `std::string` to `dai::Path`. Easy breezy. The function continues to accept std::string + more.

I'm still considering is what part of the underlying code renders the replacement text `<Unicode path can not be displayed>`.

* Current approach: the `dai::Path` class itself returns strings with `<...>`. It doesn't throw exceptions for impossible charset conversion
* Alternative: for `dai::Path` to align with `std::filesystem::path`... I would need to change `dai::Path` to throw on these impossible charset conversions. Therefore, the caller would need to catch and return the `<...>` string. This catch can be easily added to the custom fmt::formatter.

I think the choice is how much does Luxonis want `dai::Path` to align with `std::filesystem::path` so in the future you can change to the STL path. Or for those that compile depthai-core with C++17, can use the actual STL path by me aliasing it with either a `using` or deriving dai::Path from the STL path.

I very slightly prefer the 2nd alternative approach so to prepare everything now for the future.

include/depthai/utility/Path.hpp
```diff
@@ -0,0 +1,96 @@
...
+#if(__cplusplus >= 201703L) || (_MSVC_LANG >= 201703L)
+    #include <filesystem>
+#endif
+#include <string>
+
+namespace dai {
+
+// TODO C++20 char8_t
+// TODO test if caller works when replace "dai::Path" -> "std::filesystem::path"
+// TODO test if can `using Path = std::filesystem::path` on C++17 to completely use STL
```
themarpe, line 11: +1 - maybe we do that ifdef here and create another c++14 vs c++17 compilation matrix entry, to compare.
To reduce the testing load we could for starters do the specific scenario with the test below:
 - dynamic lib / static lib
 - c++14 / c++17
 
 And then consume the lib, compile the test and run

include/depthai/utility/Path.hpp
```diff
@@ -0,0 +1,96 @@
...
+    template <typename FormatContext>
+    auto format(const dai::Path& p, FormatContext& ctx) -> decltype(ctx.out()) {
+        return format_to(ctx.out(), "{}", p.string());
+    }
+    */
+    template <typename FormatContext>
+    auto format(const dai::Path& p, FormatContext& ctx) {
+        return formatter<std::string>::format(p.string(), ctx);
+    }
+};
```
themarpe, line 96: spdlog is private dependency - this must be hidden in `src/utility/Path[...].hpp`

include/depthai/utility/Path.hpp
```diff
@@ -0,0 +1,96 @@
...
+
+   public:
+    Path() = default;
+    ~Path() = default;
+    Path(const Path&) = default;
+    Path(Path&&) = default;
+    Path& operator=(const Path&) = default;
+    Path& operator=(Path&&) = default;
+
+    Path(string_type&& source) noexcept : nativePath(std::move(source)) {}
```
themarpe, line 40: This I assume will nicely do Python -> C++ integration/conversion. We'd still have to either:
 - Create a wrapper in bindings that takes std::string and then calls this
 - Or bind and convience pybind that it also implicitly converts python strings to dai::Path

Eitherway, it'll be easy and small changes required there

include/depthai/utility/Path.hpp
```diff
@@ -0,0 +1,96 @@
+#pragma once
```
themarpe, line 1: What about using something that "mimics" std::filesystem::path, like: https://github.com/gulrak/filesystem
We'd just use the path if easily extractable.

And/or maybe to address the possible ABI issues to not swap the implementation out to std::filesystem::path if consumer is C++17 but if the library was compiled with C++17 as well.
We'd carry this build information along (`build/config.hpp`), and only then swap the API as well - eg `using dai::Path = std::filesystem::path`

include/depthai/utility/Path.hpp
diablodale: I don't understand your feedback.
after a cmake install, that needed spdlog is at:
`.install/Debug/lib/cmake/depthai/dependencies/include/spdlog/fmt/bundled/...`

include/depthai/utility/Path.hpp
diablodale: yes. If your python code outputs a std::string in its binding, this should work with no python changes.
If not, then it should be a code change of a few characters to call the binding api that outputs a std::string.

include/depthai/utility/Path.hpp
diablodale: While you could use gulrak, 95% of it isn't not needed. depthai-core needs only a tiny portion (utf path conversion)...and only on Windows. I don't see value in taking on a 3rd party dependency if you don't need 95% of its features and only on one of your platforms.

My thinking (some already implemented) is...

* library is compiled with C++17:  library will internally use filesystem::path. dai::Path will have a user defined type conversion that knows how to convert from dai::Path -> filesystem::path. The native capabilities of filesystem:path + the UDTC enable accepting all types from all apps.
* library is compiled with C++14: library will use dai::Path as in the current PR code. If a C++17 app (like my own app) wants to use filesystem::path and call dai::loadMeshFiles(dai::Path), then the app needs to convert it to something dai::Path accepts like a std::string with the many filesystem::path::xxxx member functions. But I think I can do this automatically because in the Path.hpp header, I can detect they have std::filesystem support and conditionally enable a UDTC.

include/depthai/utility/Path.hpp
diablodale: To date, depthai-core forces compile depthai-core as C++14
https://github.com/luxonis/depthai-core/blob/e03be4c7b43307b148fd1e4da4306e7166ded2a4/CMakeLists.txt#L244-L247

It is not possible to change to c++17 without someone changing the build or code of depthai-core itself. I don't even do that. 😉 While it is technically possible to ?attempt? a c++17 compile...I (personally) wouldn't prioritize that at this time. There are enough other things to resolve with the SDK. 

In contrast, it *is very* likely there are apps (like mine) that compile themselves as C++17 and link to the c++14 compiled depthai-core. This is the scenario I think is valuable for the testing matrix. I'm already looking forward to the c++20 apps due to c++20's utf-8 handling.

diablodale: Latest commit added optional param to `dai_add_test()` where one can declare the desired c++ standard level for the test. It defaults to c++14 like the toplevel depthai-core target.

For now, it only adds a c++17 test for the filesystem test. If the current toolkit/chain doesn't support C++17, it will probably fail to cmake configure. Is there a way to detect the range of C++ feature levels possible with the current toolchain? I only found the hack `CHECK_CXX_COMPILER_FLAG("-std=c++17" COMPILER_SUPPORTS_CXX17)`

include/depthai/utility/Path.hpp
diablodale: OK. I split the header into two parts. Public header in include tree. Private header in src tree.
I cmake config, build, installed and used that installation successfully with my app.

diablodale:
New force push enables c++17 std::filesystem support.
depthai-core is compiled as c++14 as per the make CMakefile.txt
the single tests/filesystem_test is compiled c++14 and c++17 and links to the c++14 depthai-core.

Works ok in my app and in this single tests/filesystem_test.

👎The test+example suite is too long now due to multiplying with conform + poe.
I think its is going to take more than an hour. (I can test all of OpenCV in less than an hour)

diablodale:
today added dai::Path to dai::Device, DeviceBase, Resources.

One highlight: the `bool usb2Mode` param that appears often is troublesome. I saw several constructors in the code with `char *` to work around C/C++ love to implicitly change most POD into a bool...causing ambiguity/conflicts in overloads.

I resolved this with a template to specifically check the usb2Mode param type = `bool` which then enabled the `dai::Path` param to handle all those needed: char*, wchar_t*, string, wstring, std::path, dai::Path. This reduced the number of constructors and remove the need for some code in cpp files.

I added more test cases. It still passes the full test+example suite. I saw the matrix found some issues on a Linux build, I'll get to it soon.

diablodale:
@themarpe, does your team intend to have duplicate constructors in `Device`? Want to ask before I remove duplications.

`Device` class `using` inherits all `DeviceBase` constructors. The 10 constructors in `Device` are the same code as those in `DeviceBase`. They are duplicated, its the same code in both classes. 🤲

My last commit has a bug in it. While fixing, I noticed this dup and since I need to fix one of these constructors, I have to also fix the dups. 🤔 If it works, can I remove the dups? Or should I leave them.

diablodale:
Hi, please run the CI on this. 3f53255e7f587e6b74235c1de924025295f82874 completes all intended work and time to shift towards bugfix, review feedback.

diablodale:
rebased on develop. Would you please run CI again?

themarpe:
> @themarpe, does your team intend to have duplicate constructors in Device? Want to ask before I remove duplications.

> Device class using inherits all DeviceBase constructors. The 10 constructors in Device are the same code as those in DeviceBase. They are duplicated, its the same code in both classes. palms_up_together

> My last commit has a bug in it. While fixing, I noticed this dup and since I need to fix one of these constructors, I have to also fix the dups. thinking If it works, can I remove the dups? Or should I leave them.

Sorry for the delay @diablodale 
Those are required unfortunatelly. MSVC bug... - cannot bind them for Python otherwise, some ambiguous stuff... Would have to check for the CI run of that, don't have it at hand at the moment

themarpe:
> Hi, please run the CI on this. https://github.com/luxonis/depthai-core/commit/3f53255e7f587e6b74235c1de924025295f82874 completes all intended work and time to shift towards bugfix, review feedback.

Will give it a more thorough check in the upcoming days - otherwise looks good on glance

diablodale:
There is a now/future API concern outstanding.
We agreed that when a `std::string` is given as a "path", we must interpret this as a utf-8 string.
This is different behavior than `std::filesystem::path`.

When a std::string (or char*) is given to `std::fs::path`, it...
> If the source character type is char, the encoding of the source is assumed to be the native narrow encoding (so no conversion takes place on POSIX systems)

This is true also on Windows. Windows will not do any significant conversion. Looking the MSVC STL itself, I can see it does only
```c++
_Convert_narrow_to_wide(__std_fs_code_page(), _Input);
```
...meaning it just converts the current Windows codepage (legacy windows tech) from 8-bits into 16-bits. It does not do a utf-8 conversion.

So...there is not 100% compatibility of `dai::path` and `std::filesystem::path` because the `std::string` constructors behave differently...

```c++
std::filesystem::path::path(std::string);
dai::path::path(std::string);

std::string myutf8 = "🎆🧵👓"

// correct on utf-8 systems like Linux
// garbage on Windows
auto stdPath = std::filesystem::path(myutf8);

// correct on Linux and Windows
auto daiPath = dai::path(myutf8);
```

I would be ok keeping dai::path through c++17 and c++20. It is thin and already uses+converts automatically between std::fs::path.

themarpe (changes requested):
Would you mind opening a PR against `depthai-python` `develop` as well? Just so we can run the CI & compile with bindings in place as well

include/depthai/utility/Path.hpp
```diff
@@ -0,0 +1,199 @@
...
+#if(__cplusplus >= 201703L) || (_MSVC_LANG >= 201703L)
+    #include <filesystem>
+    #define DAI_NODISCARD [[nodiscard]]
```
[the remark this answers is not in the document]
diablodale: * Will do macro change.
* I would also like to squash, but haven't so you can review steps/evolution of this PR. When you are ready, I'll squash and forcepush.
* The bindings I am confident we can get working, but I have never written code in the binding+python space so I'm lacking experience about what exactly to do, different aspects to consider, etc. I suspect one or two places will need an explicit call for python to output its object as a string. Or maybe a user-def type conv (or path constructor) that takes the python object -> dai::path. How "bindings" work...I'm inexperienced 🤔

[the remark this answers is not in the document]
themarpe: > I would also like to squash, but haven't so you can review steps/evolution of this PR. When you are ready, I'll squash and forcepush.

Thanks - I'll recheck later today a tad more, but its looking good already

> The bindings I am confident we can get working, but I have never written code in the binding+python space so I'm lacking experience about what exactly to do, different aspects to consider, etc. I suspect one or two places will need an explicit call for python to output its object as a string. Or maybe a user-def type conv (or path constructor) that takes the python object -> dai::path. How "bindings" work...I'm inexperienced thinking

If you want, you may open the PR and give me push permissions to it, and I'll do any changes so that it works together with this PR. Hopefully there won't be any changes needed back to core, but it might happen. (also in this case, push permissions back to this PR or I'll post back some "suggested changes")
I think a python obj -> dai::Path will be the cleanest, as it will retain the existing mechanisms of binding the functions (without having to cast to string first, for dai::Path to then do the conversion).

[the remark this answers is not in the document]
diablodale: An update. I'm not going to have time to write code for the `depthai-python` repro. Python isn't my target, focus, or knowledge.

I burned all of today trying to get depthai-python cmake config working. Mypy still isn't found so no stubs. Compile quickly fails. I predict large churn like I had to do for `depthai-core`. I don't have a week just to get the cmake working and then tweak a few bindings that someone with binding experience/setup can likely do in an hour.

I will 100% do the code in `depthai-core`. If someone finds a binding that doesn't work and needs depthai-core changes...tell me and I'll write/fix the code of this PR. 💯

I read https://pybind11.readthedocs.io/en/stable/advanced/cast/strings.html#passing-python-strings-to-c and that doc writes the bindings will automatically work if the signature accepts std::string. The dai::Path class has that signature so if the bindings do what their docs write...this should be no code change. 😂 Riiight.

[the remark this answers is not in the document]
themarpe: Sure, no issue - I'll open one and make any (if) necessary bindings changes.

include/depthai/device/DeviceBootloader.hpp
```diff
@@ -168,12 +168,19 @@ class DeviceBootloader {
...
+
     /**
      * Connects to or boots device in bootloader mode depending on devInfo state.
      * @param devInfo DeviceInfo of which to boot or connect to
-     * @param allowFlashingBootloader Set to true to allow flashing the devices bootloader. Defaults to false
+     * @param allowFlashingBootloader (bool) Set to true to allow flashing the devices bootloader
      */
-    explicit DeviceBootloader(const DeviceInfo& devInfo, bool allowFlashingBootloader = false);
+    template <typename T, std::enable_if_t<std::is_same<T, bool>::value, bool> = true>
+    DeviceBootloader(const DeviceInfo& devInfo, T allowFlashingBootloader);
```
themarpe, line 183: Was this change here done purely from documentational standpoint?

themarpe:
Have the branch for `depthai-python` ready - I'll sync both in the next days and merge

include/depthai/utility/Path.hpp
```diff
@@ -0,0 +1,199 @@
...
+     * @brief Observes if path is empty (contains no string/folders/filename)
+     *
+     * @return bool true if the path is empty, false otherwise
+     */
+    DEPTHAI_NODISCARD bool empty() const noexcept {
+        return _nativePath.empty();
+    }
+
+   private:
+    string_type _nativePath;
```
themarpe, line 196:
```suggestion
    string_type nativePath;
```

Nitpick - unless it follows some other source

include/depthai/device/DeviceBootloader.hpp
diablodale: See https://github.com/luxonis/depthai-core/pull/384#issuecomment-1049372357 and the comment after that.

Having all those `Device`, `DeviceBase`, and `DeviceBootloader` constructors and having a `bool` param in the same positions as another POD param hides other constructors. Almost/all POD degrades implicitly to a `bool` (int->bool, float->bool, pointer->bool, etc.). For example, those "Device/base(usb2mode)" bools always have Priority 1 in the compilers type conversion lookup. It was only with a template that I could force only a true `bool` to be accepted which then allowed other parameters types in the same param position to be used (or converted) as desired. If I didn't, then I would have to add ~10 more constructors *each* on Device, Devicebase, etc. to deal with those `bool` params.

I suggest deprecating and deleting all the "usb2mode" constructors on Device and Devicebase. They duplicate sibling constructors. `Device(mypipe, true)` and `Device(mypipe, UsbSpeed::HIGH)` are duplicates. The former is a headache to deal with these "degrade to bool" issues. The latter does everything + more... and... it is self-documenting.

`Devicebootloader` is similar scenario. The `bool allowFlashingBootloader` causes everything in that parameter position to degrade to a bool and then that constructor gets priority 1 and is called.
```c++
// always called
DeviceBootloader(const DeviceInfo& devInfo, bool allowFlashingBootloader);

// hidden and never called
DeviceBootloader(const DeviceInfo& devInfo, const dai::Path& pathToBootloader, bool allowFlashingBootloader = false);

// calls the 1st bool constructor because the compiler degrades mystr pointer to bool and matches it
char* mystr = "asdasd";
auto db = new DeviceBootloader(devi, mystr);

// calls the 1st bool constructor because the compiler prioritizes "degrade to bool" higher
// than the converting constructor of Path::Path(std::string)
std::string mystdstr = "asdasd";
auto db2 = new DeviceBootloader(devi, mystdstr);
```

also the `explicit` only works on 1 parameter converting constructors. After c++11 you can use it to forbid initialization lists on 2+ param constructors but that isn't our intention here.

So...I use the template approach and updated doxygen to be more specific.

include/depthai/utility/Path.hpp
diablodale: I did it for usability and to distinguish it from `std::filesystem::native()`.

`dai::path::_nativePath` is a private member which no dev can access. Some coding tools (like vscode) still show private members in their "intellisense". As I was writing/testing, I found myself often trying to use `nativePath` because it was listed and similar to `std::fs::native`. I considered changing to a different noun but I couldn't come up with another word that worked as well. So I used an underscore which completely separates it in Intellisense and visually suggests this is something internal/temp.

I don't recommend `nativePath`. Do you have another noun you prefer? It can NOT be: native, path, directory, nativePath, nativeDirectory, etc. I think it could be "folder" but...yuck 😝

diablodale:
> Have the branch for `depthai-python` ready - I'll sync both in the next days and merge

Looked at your dai-python commit. That would have taken me a loong time. Must is a search/replace but I can discern there are certain places which need experienced handling. 🙏Thank you for partnering on that.

include/depthai/utility/Path.hpp
diablodale: pathway, trail?

themarpe (approved):
LGTM @diablodale 
Sorry for the delay. Is there anything else you've wanted to do before merge?

Otherwise I'll merge tomorrow /w accompanying python bindings changes:) 

(And thanks a lot for this effort!)

diablodale:
Something before merge?
I'm happy to squash the 23 commits -> 1.

If you want that, just ask me here. Otherwise, I have no changes and ok to merge.

themarpe:
Thats okay, I'll squash merge here :+1:

PR #389, opened 2022-02-19
Add device monotonic timestamp to IMU reports

szabi-luxonis:


themarpe (changes requested):
Shall we then just rename the tsDevice to `timestampDevice` and add `getTimestampHost` for the other cases? While shared wouldn't be explicit in separating host synced and non synced device timestamp, the core would be where applicable

PR #399, opened 2022-02-22
IMU firmware update

szabi-luxonis:
Added option to do IMU firmware update.

To perform firmware update:
`imu->enableFirmwareUpdate(true)`


themarpe:
As per discussion in #431

I think it might make more sense to version any new features that the updated FW brings and have the FW update be an explicit RPC command (also a command to retrieve current version and latest version). In case a feature requires a newer IMU, it could either warn/error or perhaps not allow to run with existing depthai library version.

CC: @ZigiMigi  if/when these get added to core, we should add them to: https://github.com/luxonis/depthai-python/pull/518

szabi-luxonis:
> As per discussion in #431
> 
> I think it might make more sense to version any new features that the updated FW brings and have the FW update be an explicit RPC command (also a command to retrieve current version and latest version). In case a feature requires a newer IMU, it could either warn/error or perhaps not allow to run with existing depthai library version. (IMO main cause being the long downtime it causes)
> 
> CC: @ZigiMigi if/when these get added to core, we should add them to: [luxonis/depthai-python#518](https://github.com/luxonis/depthai-python/pull/518)

RPC wouldn't work with ESP boards.

themarpe:
Do you mean in context of doing OTA for IoT devices? I guess we could leave a property in to cover that option, but shouldn't be the main way or default to true.

themarpe:
Just to pindown the approach here - lets leave it as part of pipeline as an "auto-update" but as default OFF, and add mechanism to update explicitly via RPC calls. Also to add are querying of current version and other parameters + being able to retrieve what is the available verison in the FW.

Then we'll expose these options in "device_manager", as explicit upgrade command, which also displays current information to the user (current ver, available ver, etc...)

PR #410, opened 2022-03-01
XLink device search improvements

themarpe:
Adds additional information to DeviceInfo like mxId and status and some underlying XLink changes:
 - PoE devices can now be discoverable by both IP address as well as MXID
 - Insufficient permissions (caused by missing udev rules) are reported back transparently
 - libusb is no longer a required separate dependency, but part of XLink dependencies
 - Improves device search speed (reduces number of broadcasts required to search for devices to one)
 - Removes the need for public flags to be set when consuming DepthAI library

TODO:
 - Fix warnings for devices with errors. Perhaps skip over invalid ones without issuing a warning, but only issue for missing permissions


include/depthai/xlink/XLinkConnection.hpp
```diff
@@ -25,11 +25,22 @@ namespace dai {
  */
 struct DeviceInfo {
     DeviceInfo() = default;
-    explicit DeviceInfo(const char*);
-    explicit DeviceInfo(std::string);
-    deviceDesc_t desc = {};
```
diablodale, line 30: Hi, this change breaks all previous apps that used `desc` (like mine). It is an incompatible API change.
This code runs well in depthai-core v2.15.3 and previous versions...
```c++
const auto targetDevice = dai::Device::getDeviceByMxId(myId);
if (std::get<dai::DeviceInfo>(targetDevice).desc.protocol != XLinkProtocol_t::X_LINK_TCP_IP)
```

I think somewhere Luxonis writes the 2.x.x is not yet stable API. But want to double-check on this as I (perhaps others) have to accommodate this.

themarpe: True, we could add a compatible holder for this, and populate the information upon creation, but only for that case. Otherwise deprecate and encourage the move over to direct members of DeviceInfo

I'll update this and mark `desc` as deprecated

diablodale:
The last commit of the new xlinkpublic target fails when depthai-core is cmake config on my app.
It is due to `depthaiDependencies.cmake` line 73.

My app's cmake does `find_package(depthai 2.15.0 CONFIG REQUIRED)`
which leads to the `cmake config` fail...

```
[cmake] CMake Error at C:/njs/vcpkg/scripts/buildsystems/vcpkg.cmake:824 (_find_package):
[cmake]   Could not find a package configuration file provided by "XLink" with any of
[cmake]   the following names:
[cmake] 
[cmake]     XLinkConfig.cmake
[cmake]     xlink-config.cmake
[cmake] 
[cmake]   Add the installation prefix of "XLink" to CMAKE_PREFIX_PATH or set
[cmake]   "XLink_DIR" to a directory containing one of the above files.  If "XLink"
[cmake]   provides a separate development package or SDK, be sure it has been
[cmake]   installed.
[cmake] Call Stack (most recent call first):
[cmake]   C:/njs/depthai-core/.install/Debug/lib/cmake/depthai/depthaiDependencies.cmake:73 (find_package)
[cmake]   C:/njs/depthai-core/.install/Debug/lib/cmake/depthai/depthaiConfig.cmake:14 (include)
[cmake]   C:/njs/vcpkg/scripts/buildsystems/vcpkg.cmake:824 (_find_package)
[cmake]   cmake/find-oak.cmake:15 (find_package)
[cmake]   CMakeLists.txt:81 (include)
[cmake] 
[cmake] 
[cmake] -- Configuring incomplete, errors occurred!
[cmake] See also "C:/njs/dp.kinect3/build/CMakeFiles/CMakeOutput.log".
[cmake] See also "C:/njs/dp.kinect3/build/CMakeFiles/CMakeError.log".
```

Since that single header file is now a transitive interface library, we need to adjust the xlink handling in `depthaiDependencies.cmake`. I don't think projects that consume depthai-core need to explicitly `find_package(XLink)`. Any needed static/shared linking to XLink ?should? be transparently handled by the transitive linking downstream from xlink->depthaicore->app

diablodale:
my DLL app (which I used for the test fail) is Windows x64...
1. DLL app, DLL depthai-core, static XLink, DLL libusb
2. xlink: is `device_search_improvements_libusb` at c152a73ebdcae14a0b144207979fa25eefcd6aec + my 4 commits from `fix37-libusb-win-loadpath`
3. depthai-core: is `xlink_device_search_improvements` at 477cb6f84513172286615c747caaee35f0bfa829
4. depthai-core: has DEPTHAI_XLINK_LOCAL=path of (2)
5. depthai-core: cmake clean, config, and build debug
6. depthai-core: cmake install
7. my DLL app: `set(depthai_DIR "$ENV{DEPTHAI_INSTALL}/Debug/lib/cmake/depthai")`
8. my DLL app: cmake config, and get the error that xlink isn't found

🤔Thoughts...

XLink is not needed (or desired) for all downstreams of depthai-core. No downstream should `find_package(XLink)`.
Xlink is a dependency of depthai-core, not downstreams of depthai-core. This is the reason for PRIVATE linking of XLink.
Only the one public header file in the new lib `XLinkPublic` should be used by downstreams.
I am *not* confident if the downstream needs to `find_package(XLinkPublic)` or it will be automatically found/handled by cmake. But experiments below suggest that is likely needed.

In the install directory of depthai-core from step 6 above...
* no results from `find . -iname '*.cmake'|grep -i public`. That public header can not be found with `find_package()`
* if I edit the depthai-core install `lib/cmake/depthai/depthaiDependencies.cmake` line 72 to be `elseif(NOT CONFIG_MODE)` then I get the error `The following imported targets are referenced, but are missing: XLinkPublic`. Since XLink is no longer needed/desired by downstreams, then there should only be a `find_package(XLink)` for depthai-core during its own config mode. Probably. I think. Perhaps. 😉 That makes logic sense to me. Only depthai-core needs XLink and we only need this `if()` to handle the xlink hunter/local change.

I suspect if I create cmake config files for XLinkPublic by above article's instruction, and apply that change to line 72 I might have more success. I'll experiment with some XLink cmake changes + moving the Windows delayload code into a separate file.

themarpe:
Thanks - will work on Wins later today and try to recreate.

> XLink is not needed (or desired) for all downstreams of depthai-core. No downstream should find_package(XLink).

Not XLink target, but the package - when finding the package both XLink & XLinkPublic targets are going to be available.

> I am not confident if the downstream needs to find_package(XLinkPublic) or it will be automatically found/handled by cmake. But experiments below suggest that is likely needed.

Should be necessary - also why nlohmann_json and libnop have to be specified.

So before I have a chance to experimentally try it out - writng down the theory of how CMake should handle this:
 - XLink is installed under `lib/cmake/depthai/dependencies/lib/cmake/XLink`
 - Containing XLinkPublic (& XLink) targets
 ```
# Create imported target XLinkPublic
add_library(XLinkPublic INTERFACE IMPORTED)

set_target_properties(XLinkPublic PROPERTIES
  INTERFACE_COMPILE_FEATURES "cxx_std_11"
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include"
)
```
 - When find_package(depthai), it should find_package(XLink). If that doesn't work (as in initial error), a check should be made where it searches for this XLink. It should find: `lib/cmake/depthai/dependencies/lib/cmake/XLink/XLinkConfig.cmake`
 - Possibility for not finding it are (IMO):
     - Misinterpreted/wrong CMAKE_PREFIX_PATH or _IMPORT_PREFIX (unlikely, others are found)
     - vcpkg's "overriden" `find_package` doing something differently than stock
     - `HINTS` parameter messing up the search - other seem to be found, maybe try removing the HINTS.

**Comment by diablodale** [comment]
if I change the `find_package(xlink HINT)` on depthaiDependencies.cmake line 73 to be up one directory with `"${CMAKE_CURRENT_LIST_DIR}/../XLink"` then my app's config succeeds. Logically, this is correct on this windows cmake install for depthai-core because...
```
C:\njs\depthai-core\.install\Debug\lib\cmake\depthai\depthaiDependencies.cmake
C:\njs\depthai-core\.install\Debug\lib\cmake\XLink\XLinkConfig.cmake
```

There is no subdirectory C:\njs\depthai-core\.install\Debug\lib\cmake\depthai\XLink which would match the current HINT. Adding `..` allows the find_package() to succeed. Perhaps a cmake var that represents the location could also be used.
Removing the HINT fails to find xlink.

**Comment by diablodale** [comment]
Very different with hunter xlink. Cmake changes the dependency/hierarchy

```
local     .install\Debug\include\XLink\XLinkPublicDefines.h
hunter    .install\Debug\lib\cmake\depthai\dependencies\include\XLink\XLinkPublicDefines.h

local     .install\Debug\lib\cmake\XLink\XLinkConfig.cmake
hunter    .install\Debug\lib\cmake\depthai\dependencies\lib\cmake\XLink\XLinkConfig.cmake
```

using `..` continues to work in my downstream app's cmake config and build succeeds.
Ninja is using these related include dirs
```
-IC:\njs\depthai-core\.install\Debug\include
-IC:\njs\depthai-core\.install\Debug\include\depthai-shared\3rdparty
-IC:\njs\depthai-core\.install\Debug\lib\cmake\depthai\dependencies\include
```

themarpe:
@diablodale 
~~Wondering why you have XLink installed under same folder as depthai? Was this tested with DEPTHAI_XLINK_LOCAL? ~~

Got it - this is something we don't have in CI yet... (mostly a development option I guess)

Great, in that case, lets add another hint, the one with one direction higher. The inital one was afaik required for some Android specifics to work, don't want to remove it for the time being.

include/depthai/xlink/XLinkConnection.hpp
themarpe: Due to issues having duplicate data in this structure and syncing it, I've decided to not add compatibility for this. As the use isn't too common and changes are small (removing `desc.` and accessing fields directly) I think this won't cause too much issues.

Issue #415, opened 2022-03-06
OAK-D-Pro-PoE fails after 6 connections and requires power-cycle


diablodale:
OAK-D-Pro-PoE fails after 6 connections and requires power-cycle. I now have repro code. 🙃
When OAK enters this dead state, it does not respond to pings or appear active on the ethernet network (though my LAN diagnostic tools are limited)

A PR is forthcoming that holds the repro code. It is a new "performance" test which I would like to eventually merge so that Luxonis and others can instrument/evaluate the OAK hardware CPU+temperature stats with different builds and different permutations of Pipeline settings.

### Setup

* OAK-D-Pro-PoE preproduction with bootloader v0.0.17
* Ubiquiti POE Injector PoE+ Gigabit 802.3af -- a full Ubi switch is forever not available in Europe 😢
* TP-Link 1gpb ethernet switch
* Microsoft Windows [Version 10.0.19044.1526]
* depthai-core v2.15.0
* VS2019 v16.11.10  (a c++17 compiler is required for the repro code)

### Repro

1. git clone from branch in forthcoming PR. The PR will hold the repro code
2. config, build for: x64, Debug, shared libs, with all examples and tests
3. run the new test `build\tests\perf_hw_cputemp.exe`

### Result

You will see 6 lines, each showing the permutation of device settings `device conf: ....`
You will see 1 or 2 lines of the `samples` pathname where the test is saving sample data.
Sometime between the 6th and 7th run, the OAK-D-Pro-PoE will die and can no longer be used until power-cycled.

This 7th attempt will hang in code until it finally times out ~1 minute.
You will receive Catch2 errors, messages, exceptions, etc.

```
device conf: 1646611432,0,400,0,false,false,false,false,5,30.000000,tcp,
samples: C:\njs\depthai-core\build\tests\hw-cputemp.csv
device conf: 1646611451,0,400,0,false,false,false,true,5,30.000000,tcp,
device conf: 1646611474,0,400,0,false,true,false,false,5,30.000000,tcp,
device conf: 1646611496,0,400,0,false,true,false,true,5,30.000000,tcp,
device conf: 1646611519,0,400,0,true,false,false,false,5,30.000000,tcp,
device conf: 1646611542,0,400,0,true,false,false,true,5,30.000000,tcp,
device conf: 1646611565,0,400,0,true,true,false,false,5,30.000000,tcp,

[...]

PR #416, opened 2022-03-07
hardware device performance test and data collection

diablodale:
DO NOT MERGE
First, this PR is first intended to demonstrate issue #415. 👍
Second, I would like to see an evolved version of this PR to be later merged after discussion on its usage/value. ⏳

* adds a new performance test `tests/hardware_cputemp_perf.cpp`
* written for C++17 -- can't be bothered to make it work with earlier
* updates Flags.cmake to discover the compiler's range of support for C++ standards
* updates cmake for tests so can specify the c++ standard for any test and the cmake harness will do the right thing


diablodale:
Sorry, I might not have been clear enough. The test in this PR is not designed for today's CI to run the test.
Build will hopefully work.
But probably not run. Because on an OAK-D it takes 3 hours for that one test exe. And it writes the collected metrics into the same directory as the exe, so it needs write access to that location.

I recommend you use CI or your own local toolkit to build. And then run the test yourself.

If you have a method in which your CI can run a test for 3 hours (oak-d) or probably a day (for PoE if it ever works), then I'm happy to adapt the code. I can also add whatever is necessary so that the performance metrics are written/collected.

PR #429, opened 2022-03-21
Some things for uploading Fisheye camera calibrations

slitcch:


slitcch:
This also requires some changes in depthai and depthai-python - I factored setDistortionCoefficients and setCameraType into one function - didn't see any other way to validate that the correct number of parameters were supplied. Feel free to reject these changes if you see another way. The main important bit that's definitely necessary is `getDistortionModel` - unless I'm crazy there wasn't yet a way to see if the camera was a Fisheye camera.

src/device/CalibrationHandler.cpp
```diff
@@ -524,18 +531,60 @@ void CalibrationHandler::setCameraIntrinsics(CameraBoardSocket cameraId, std::ve
...
+    // Adding CALIB_THIN_PRISM_MODEL adds four more - s1, s2, s3, s4, totaling to 12. This isn't a common choice.
+    //
+    // Adding CALIB_TILTED_MODEL adds two more - tau1 and tau2, totaling to 14. This is a common choice - it sounds right to ENABLE ALL THE THINGS. While it
+    // does work, I've seen it result in a lot more failed calibrations, that's why I recommend kb8.
+    //
+    // For RATIONAL, THIN_PRISM and TILTED, zeroing out all the coefficients is the same as ignoring them - there's no effect.
+    //
+    // Soooo....
+
+    size_t num = distortionCoefficients.size();
```
saching13, line 553: Since the EEPROM data in the firmware is static POD. I think it is best to assign the rest to zero. And when reading them check the rest of them for zero. if it is zero get the partial of it based on the model. If it is not done here and want to have it only 4 parameters. will will just do the same on the firmware side. So easier to just crop it to return the first 4 .

And as of all the OpenCV model. We don't have plan to support all the distortion model for perspective yet. Currently, we are supporting only the Rational Polynomial distortion model and does not support the rest. I think in the comments we can add the default supported is Rational Model. Don't want other users who uses it on device to confuse and try to use them.

src/device/CalibrationHandler.cpp
```diff
@@ -524,18 +531,60 @@ void CalibrationHandler::setCameraIntrinsics(CameraBoardSocket cameraId, std::ve
...
+
+            if(!((num == 5) || (num == 8) || (num == 12) || (num == 14))) {
+                throw std::runtime_error("for Perspective distortion, distortionCoefficients size should be 5, 8, 12, or 14.");
+            }
+        } break;
+        // I know what Equirectangular is - that's one way of mapping a 360 degree image onto a 2D texture, but I don't know how the parameters work there at all.
+        //
+        // I have no idea what RadialDivision is.
+        //
+        // Either way, I can't find any references to them outside the enum they're defined in, so I think they're safe to ignore for now.
```
saching13, line 572: that's fine. They are added just in case we need them in future.

src/device/CalibrationHandler.cpp
```diff
@@ -524,18 +531,60 @@ void CalibrationHandler::setCameraIntrinsics(CameraBoardSocket cameraId, std::ve
     return;
 }
 
-void CalibrationHandler::setDistortionCoefficients(CameraBoardSocket cameraId, std::vector<float> distortionCoefficients) {
-    if(distortionCoefficients.size() != 14) {
-        throw std::runtime_error("distortionCoefficients size should always be 14");  // should it be ??
+void CalibrationHandler::setDistortion(CameraBoardSocket cameraId, std::vector<float> distortionCoefficients, CameraModel distortionModel) {
```
saching13, line 534: Don't change and deprecate the functions. it will affect existing implementations. 

If the incoming distortionCoefficients is not 14. Append zeros to make it 14. And use `setCameraType` to set the distortion model.

slitcch: sounds good!

PR #449, opened 2022-04-09
Openvino: Fix error reporting when blob is not set

szabi-luxonis:
Reports: `RuntimeError: DetectionNetwork(1) - No blob is loaded`
instead: `RuntimeError: Pipeline - 'DetectionNetwork' node with id: 1, isn't compatible with forced OpenVINO version`
when openvino version is set explicitly and blob is not set (`setBlobPath`)


themarpe:
Tests are failing - seems like reporting the network version from pipeline isn't working as expected..

szabi-luxonis:
> Tests are failing - seems like reporting the network version from pipeline isn't working as expected..

Fixed. It was an interesting one.

Issue #458, opened 2022-04-23
regression in test case: missing "_conforming" suffix in target_compile_options()


diablodale:
Commit https://github.com/luxonis/depthai-core/commit/dcc77eff124d60c4219b96f12c7e5c8e628da3b6 introduced a cmake bug and surfaces a compiler/CI issue.

https://github.com/luxonis/depthai-core/blob/abfbd6ae78f6b9fbee34bf88ecaf4e659bd0976a/tests/CMakeLists.txt#L60

1. CI should have failed to compile all conforming preprocessor tests that need c++17 or c++20 and surfaced error report(s)...but it didn't
2. code above is missing the "_conforming" suffix. It should be `${test_name}_conforming`

### Setup

* Microsoft Windows [Version 10.0.19044.1645]
* VS2019 v16.11.13
* depthai-core v2.15.3

### Repro 1 of CI issue

1. Config and build for x64, release, shared lib

### Result 1 of CI issue

The existing `dai_add_test(filesystem_test_17 src/filesystem_test.cpp CXX_STANDARD 17)` should fail to compile.
But it doesn't. The compile succeeds and there are no visible warnings or errors.
The "conforming" test in the generated pair should fail to compile because without C++17 there is no `<filesystem>`.
But it doesn't fail in the CI or on my local computer. I do not know why. 😕

The failure of cmake/compiler/CI to compile bad code and not surface a problem...is a problem itself.

This is ninja.build for that test...you can see there is no c++ standard flag 🙈
```
build tests\CMakeFiles\filesystem_test_17_conforming.dir\src\filesystem_test.cpp.obj: CXX_COMPILER__filesystem_test_17_conforming_Release ..\tests\src\filesystem_test.cpp || cmake_object_order_depends_target_filesystem_test_17_conforming
  DEFINES = -DBLOB_PATH=\"C:/.hunter/_Base/PrivateData/3329bb8f3a9c881ef9756d232055f9d6f38aa07b/3329bb8/raw/mobilenet-ssd_openvino_2021.2_8shave.blob\" -DDEPTHAI_TARGET_CORE -DJSON_USE_IMPLICIT_CONVERSIONS=1 -DXLINK_USE_MX_ID_NAME=ON -D__PC__ -Dfilesystem_test_17_conforming_EXPORTS
[...]

PR #459, opened 2022-04-23
reduce num conforming tests; add missing _conforming test suffix

diablodale:
Fixes luxonis/depthai-core#458

* adds missing "_conforming" suffix for test cmakelists.txt
* passes all tests `[ctest] 100% tests passed, 0 tests failed out of 26`

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -41,7 +41,7 @@ function(dai_add_test test_name test_src)
 
     # Set compiler c++ standard
     target_compile_options(${test_name} PRIVATE "${CMAKE_CXX${DAT_CXX_STANDARD}_STANDARD_COMPILE_OPTION}")
-    
+
     # Add to list of tests
     add_test(${test_name} ${test_name})
     # Add ubsan halt on error
@@ -57,8 +57,8 @@ function(dai_add_test test_name test_src)
         target_link_libraries(${test_name}_conforming PRIVATE depthai-core Catch2::Catch2)
 
         # Set compiler c++ standard
-        target_compile_options(${test_name} PRIVATE "${CMAKE_CXX${DAT_CXX_STANDARD}_STANDARD_COMPILE_OPTION}")
-        
+        target_compile_options(${test_name}_conforming PRIVATE "${CMAKE_CXX${DAT_CXX_STANDARD}_STANDARD_COMPILE_OPTION}")
+
         if(MSVC_VERSION LESS 1925)
             target_compile_options(${test_name}_conforming PRIVATE "/experimental:preprocessor")
         else()
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -11,7 +11,7 @@ find_package(Catch2 CONFIG REQUIRED)
 include(CMakeParseArguments)
 function(dai_add_test test_name test_src)
     # parse arguments
-    cmake_parse_arguments(DAT "" "CXX_STANDARD" "" ${ARGN})
+    cmake_parse_arguments(DAT "CONFORMING" "CXX_STANDARD" "" ${ARGN})
 
     # check compiler for C++ standard
     if(NOT DAT_CXX_STANDARD)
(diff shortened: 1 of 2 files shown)


themarpe:
Thanks @diablodale 
As discussed a while a go, because of shear number of tests - was thinking to reduce the `_conforming` to only one or a couple of tests. Maybe /w 14,17 & 20 standards set, like filesystem test, or maybe the device test.

diablodale:
I get it. I was thinking 🤔 about my Device(usb2mode) bug and how test cases didn't catch it because not enough apis were exercised...no get queues...no get/parse frames...etc. I would recommend a conforming test case that parses frames. A simple Script node...
```
script->setScript(R"(
    import time
    teststr = b'go'
    buf = Buffer(len(teststr))
    buf.setData(teststr)
    while True:
        node.io['out'].send(buf)
        time.sleep(20)
)");
```

with host code to verify it actually works.
```
const auto buffer = identDevice.getOutputQueue("out", 2, false)->get<dai::Buffer>(std::chrono::seconds(10), timeout);
const auto msg = buffer->getData();
if (msg == dai_buffer_type{'g', 'o'}) {
```

I can add this to the device test with most of it copy/paste from my app code.
Use either the existing SysInfo node I added https://github.com/luxonis/depthai-core/commit/4938333b6fc8ef62ecff6dc951130f62b8ada552
or switch to a Script node like above.

themarpe:
Sounds good on parsing the frames - I'd go for something else than Buffer - SysInfo sounds good (Buffer doesn't have any metadata to serialize)

diablodale:
pushed commit to strengthen the device test and use sysinfo.

themarpe:
Thanks - LGTM
In this case now, should we also reduce the conforming only to this test?

diablodale:
I see repeating errors across CI jobs. I think this is not the PR's code
```
-- Downloading Depthai device side binaries from server...
-- Downloading depthai and patch
-- commit: 648d4c1ae6dd8c6dc1540721a8960a85de3b761d
-- Status error: 22;"HTTP response code said error"
-- Status error: 22;"HTTP response code said error"
-- Status error: 22;"HTTP response code said error"
-- Status error: 22;"HTTP response code said error"
CMake Error at cmake/DepthaiDownloader.cmake:183 (message):
-- Status error: 22;"HTTP response code said error"
  Aborting.
Call Stack (most recent call first):
  CMakeLists.txt:315 (DepthaiDownload)
-- Couldn't check if depthai-shared codebase matches between device and host
-- Configuring incomplete, errors occurred!
```

> Reduce conforming to only this test?

I think that's reasonable...

* I assume the CI workflow+tests of https://github.com/luxonis/libnop/commit/4f4943ff16d6bf2b13cdf1c89ef2c4a756bac1f4 are working and always run on pushed commits. And that any libnop included via hunter will have passed CI tests. If that is not true, then I recommend that CI be fixed.
* The gnu/clang preprocessor is cross-platform conforming preprocessor; much of the "conforming" codepath is being comprehensively tested with that.
* MSVC only enables the conforming preprocessor with explicit flags like `/Zc:pre...`. (I was surprised it was not the default with c++20.) I suspect there will be a minority of MSVC dev's that enable that flag -- which further reduces the people that might be affected if this PR + libnop CI +  gnu/clang testing misses a bug.
* The bugs I've seen on this topic cause compile failures. It is not hidden/corrupted data on live devices. If/when some bug is not caught by this PR + libnop CI +  gnu/clang testing... then that developer can raise the issue to your team for triage.

themarpe:
Reran the jobs - we had some issues with server storing artifacts before, but is resolved now.

> I assume the CI workflow+tests of https://github.com/luxonis/libnop/commit/4f4943ff16d6bf2b13cdf1c89ef2c4a756bac1f4 are working and always run on pushed commits. And that any libnop included via hunter will have passed CI tests. If that is not true, then I recommend that CI be fixed.

Yes, these are run on each commit/PR to `master/develop` branches - (there is one case right now that I have to fix regarding a clean looking "all pass" CI, but isn't relevant to other tests performed)

Then lets only do it for device test, I agree with the observations.

diablodale:
pushed commit to add param on test function to control which tests are run with the MSVC conforming processor
And then updated the tests to use that to limit/reduce the conforming tests to only be run on the device usb speed

PR #476, opened 2022-05-10
fix of tryWaitAndPop

oto313:
The current implementation of tryWaitAndPop has bug. If the queue was destroyed and something waiting on condition_variable signalPush it called queue.front() because predicate result was true (!queue.empty() || destructed)


oto313:
ok thanks @themarpe, I did not notice that you already have fix. My program crashes with current version of depthai when i disconnect luxonis. But good to know that it will be eventually in main branch.

themarpe:
@oto313 
This is already merged - the changes were to remove this checks.

Can you share more details on the crash you are experiencing? Stacktraces, etc... 
Also run the app with `DEPTHAI_LEVEL=debug` env var set, and post all/any applicable logs here so we can debug further.

Issue #468, opened 2022-04-29
Simplify CMakeLists.txt


windelbouwman:
During the creation of a yocto recipe, I had to make several changes to this package `CMakeLists.txt` file.

In my opinion the top level `CMakeLists.txt` file can be simplified.

First thing I encountered was the modification of the `CMAKE_TOOLCHAIN_FILE` variable (https://github.com/luxonis/depthai-core/blob/main/CMakeLists.txt#L16). I believe this is wrong, position independent code should be turned on differently. I commented out this section, since yocto uses its own toolchain.

Furthermore, I disabled hunter, since yocto has its own package manager for dependencies. Building without hunter is possible, but required several tweaks to dependencies. This seems not tested. For example, this line breaks and should be surrounded with if-hunter-enabled checks: https://github.com/luxonis/depthai-core/blob/main/CMakeLists.txt#L595

Third, there exists a `DEPTHAI_INSTALL` variable, which is not required, we can create the install targets allways, `make install` is a seperate step, so no harm in creating those targets. In other words, this `if` can be removed: https://github.com/luxonis/depthai-core/blob/main/CMakeLists.txt#L575

I'm unsure how to proceed, should I create pull requests for those changes?

themarpe:
Hi @windelbouwman 
Sorry for late reply

The modification of toolchain is required for Hunter to build its dependencies accordingly (eg enabling PIC, etc...)

I've added "combining" of toolchains, which will be merged soon with #410

Regarding disabling Hunter - I agree, those should have the check added.

The install variable's reasoning is same as https://github.com/nlohmann/json/commit/f049836d686f5e7817682e700c3f4822fa03c9e7 - I'd leave it there.

Please do open a PR. We can then continue discussion of any potential changes there. 

Thanks!

boberfly:
@windelbouwman I'm putting my hand up for those tweaks, I'd really benefit here with a non-hunter build setup. I am trying to build in Buildstream where there is no internet connection at the build step for predictability/reproducible reasons.

Cheers

windelbouwman:
Another idea to simplify things, is to merge two repositories into this repository:

- FP16 (https://github.com/luxonis/FP16)
- Xlink (https://github.com/luxonis/XLink)

[...]

PR #492, opened 2022-05-31
Copy FP16 library into vendor folder.

windelbouwman:
This is an idea to simplify maintenance and non-hunter builds.

Feedback welcome.

Reasoning is that including such small header only projects simplifies the life of packagers and other users not using the hunter package manager.


themarpe:
Continued discussion here: https://github.com/luxonis/depthai-core/issues/468#issuecomment-1144855802

moratom:
Closing this as `v3` dropped hunter alltogether and `vcpkg` suports overlay ports in case users want to bring their own dependanices per dependancy.

PR #496, opened 2022-06-03
Set lens position from calibration data

szabi-luxonis:
Initally we used value `130` for lens position during calibration, which got hard-coded in examples. but that is now changed, camera dependent.

Reading the value from calibration data prior to setting lens position ensures correct RGB alignment for all boards.

### DIFF
--- a/examples/Script/script_forward_frames.cpp
+++ b/examples/Script/script_forward_frames.cpp
@@ -11,7 +11,6 @@ int main() {
 
     // Define a source - color camera
     auto cam = pipeline.create<dai::node::ColorCamera>();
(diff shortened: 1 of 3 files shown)


themarpe (changes requested):
Shall we rather make this auto behavior and apply it from FW if not explicitly set otherwise?
CC: @Erol444

szabi-luxonis:
> Shall we rather make this auto behavior and apply it from FW if not explicitly set otherwise? CC: @Erol444

Up to debate, merged for now, since it's an upgrade till we decide if we should or not.

PR #497, opened 2022-06-04
Release 2.16.0

themarpe:
## Features
- **OpenVINO 2022.1 support**
- **XLink device search improvements**
  - cross subnets for PoE devices
  - drastically reduced discovery time
  - Separated name / IP and MXID, to be able to query both USB path / IP and MXID
- **Android support**
- libusb dependency is now managed by Hunter
- IMU FW update for BNO sensor
- **Added DetectionParser node as a standalone component of DetectionNetwork**
- StereoDepth - subpixel fractional bits API
- VideoEncoder - lifted 16 frame pool limitation

## Bug fixes
- ColorCamera setIspScale resource allocation bugfix

## Misc
- XLinkConnection::bootBootloader improved error handling


examples/bootloader/bootloader_version.cpp
```diff
@@ -6,7 +6,7 @@ int main(int argc, char** argv) {
     std::tie(res, info) = dai::DeviceBootloader::getFirstAvailableDevice();
 
     if(res) {
-        std::cout << "Found device with name: " << info.desc.name << std::endl;
```
szabi-luxonis, line 9: FYI If we want to keep a deprecated alias for this in python, it's best to do now. Otherwise LGTM

szabi-luxonis (approved):

examples/bootloader/bootloader_version.cpp
themarpe: Will add one in Python, thanks!

PR #500, opened 2022-06-09
macOS / Linux shared library and CI improvements

themarpe:
Modifies RPATH for depthai-core.so/.dylib to point to [.so/.dylib]/cmake/depthai/dependencies/lib. This enables installed depthai core library to find libusb.so even after the build artifacts are deleted (previously located in Hunter cache).

This modification only happens for the installed library. Integration with library residing in build directory will still use the build artifacts (eg libusb from Hunter, etc...)

Context: https://github.com/luxonis/depthai-core/runs/6803956283?check_suite_focus=true


szabi-luxonis (approved):
LGTM.

PR #514, opened 2022-07-06
Fix build from sources error

themarpe:
- Properly logs when building from Source Code (no submodules) downloaded package from GitHub
- Fixes when proper core (with submodules) package is extracted and used under an existing parent git repo 

Both are improvements for initial Getting Started UX


themarpe:
A possible TODO is adding a CI check for these cases:
 - Using the proper package, extract & build (should succeed)
 - Using the proper package under an existing parent git repo, extract & build (should succeed)
 - Using "Source Code" package, extract & build (should fail, with specific error message)
 - Cloning without submodules initialized, extract & build (should fail, with specific error message)

szabi-luxonis (approved):

PR #523, opened 2022-07-14
Resolve compilation issues with spdlog and newer fmtlib (external)

oysstu:
Implicit casting of enum class to integers have been disabled in fmt v8.1.0 and newer, which causes depthai-core compilation to fail when using external fmtlib with spdlog. This PR resolves that through explicit casts and also corrects an include which bypasses the SPDLOG_FMT_EXTERNAL macro.

Related fmtlib issue: https://github.com/fmtlib/fmt/issues/1841

Signed-off-by: Øystein Sture <os@skarvtech.com>


themarpe (approved):
Thanks!

PR #543, opened 2022-08-03
Refactor asset support for Keembay

moratom:
Refactoring done as discussed with @themarpe  over Slack.

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -353,3 +353,4 @@ dai_add_example(host_pipeline_opencv host_side/host_pipeline_opencv.cpp ON)
 
 # TMP
 dai_add_example(tmp_preview tmp/tmp_preview.cpp OFF)
+dai_add_example(webcam_keembay tmp/webcam_keembay.cpp OFF)
--- a/examples/tmp/webcam_keembay.cpp
+++ b/examples/tmp/webcam_keembay.cpp
@@ -1,0 +1,128 @@
+
+#include <iostream>
+
+#include "utility.hpp"
+
+// Includes common necessary includes for development using depthai library
+#include "depthai/depthai.hpp"
+
+#define MODEL_IN_WIDTH 512
+#define MODEL_IN_HEIGHT 512
+
+#define MIN_SCORE 0.3
+
+int main(int argc, char** argv) {
+    using namespace std;
+
+    // If path to blob specified, use that
+    int camId = 0;
+    if(argc > 1) {
+        camId = std::stoi(argv[1]);
+    }
+
+    using namespace std;
+
+    // Create pipeline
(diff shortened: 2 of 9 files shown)


themarpe (changes requested):
On first glance the following would error out:
 - Create NN1, set blob (blob A)
 - Create NN2, set blob (blob B)
 - loadResource from NN2 -> returns NN1's blob A.

That said, the current mechanism might not be best suited for such refactor.
Check the following means:
 - each Node's asset manager gets equipped with its own "root path" (eg on node creation, etc...)
 - `assetManager.set("__blob")` then saves the path as eg: NN2: `/node/2/__blob`

moratom:
> On first glance the following would error out:
> 
> * Create NN1, set blob (blob A)
> * Create NN2, set blob (blob B)
> * loadResource from NN2 -> returns NN1's blob A.

Yes, it would.

> That said, the current mechanism might not be best suited for such refactor. Check the following means:
> 
> * each Node's asset manager gets equipped with its own "root path" (eg on node creation, etc...)
> * `assetManager.set("__blob")` then saves the path as eg: NN2: `/node/2/__blob`

That can be done, yes.

I just want to clarify what the desired behavior of `loadResource` is?
It would make most sense to me, that `loadResource` only checks for assets in the assetManager of the node where it is called from. So if you expect something to be in the pipeline, you can explicitly call it like `parent.lock()->loadResource("resource")`. Otherwise, if you silently "fall back" on checking the pipeline assetManager, you might get some hard to debug bugs later. 
Thoughts?

themarpe:
@moratom 
The desired behavior of loadResource is that it find an asset in the "unified" space. Eg Pipeline asset manager has an asset A, and node 1 has asset B.

Those should be accessible from anywhere by: `/pipeline/A` or `/node/1/B`, but also accessible from "current working directory" of Pipeline as such: `A` and from node 1 as: `B` (where the cwd in former is `/pipeline` and in latter `/node/1`)

> Otherwise, if you silently "fall back" on checking the pipeline assetManager, you might get some hard to debug bugs later.

Thats why the managers are prefixed - all are under unique paths. There shouldn't be no clashes. 

Another option is having just as single `Pipeline` asset manager, but I wouldn't change it now (was reworked in `3d735e0647bda393602c3f24418fd982fad5968f`, but not from the exact idea mentioned before)

moratom:
The new implementation is in [faf09b3](https://github.com/luxonis/depthai-core/pull/543/commits/faf09b3c9906dc8b84730dcd1129fe62d41cdf9f) commit.

The `loadResource` function now works as you described the desired behavior in the comment above, while the assetManager only extends the previous interface so the current examples should still work fine.
Let me know what you think @themarpe.

PR #566, opened 2022-08-25
System logger for S3 support

moratom:
A new message proposal for series 3 system logger message.


themarpe (approved):
LGTM for the time being.

Before RVC3 support is mainlined, we'll have to address these differences...

PR #574, opened 2022-08-31
Frame sync example

moratom:
Added an example that showcases how to sync frames with timestamps on the device itself.
This [depthai-experiment](https://github.com/luxonis/depthai-experiments/blob/timestamp_on-device/gen2-syncing/device-timestamp-sync.py) was took as a refrence.

CC: @Erol444

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -251,6 +251,7 @@ dai_add_example(rgb_encoding_mono_mobilenet mixed/rgb_encoding_mono_mobilenet.cp
 dai_add_example(rgb_encoding_mono_mobilenet_depth mixed/rgb_encoding_mono_mobilenet_depth.cpp ON)
 dai_add_example(rgb_encoding_mobilenet mixed/rgb_encoding_mobilenet.cpp ON)
 dai_add_example(multiple_devices mixed/multiple_devices.cpp OFF)
+dai_add_example(frame_sync mixed/frame_sync.cpp OFF)
 
 target_compile_definitions(mono_depth_mobilenetssd PRIVATE BLOB_PATH="${mobilenet_blob}")
 target_compile_definitions(rgb_encoding_mono_mobilenet PRIVATE BLOB_PATH="${mobilenet_blob}")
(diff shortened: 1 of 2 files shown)


examples/CMakeLists.txt
```diff
@@ -251,6 +251,7 @@ dai_add_example(rgb_encoding_mono_mobilenet mixed/rgb_encoding_mono_mobilenet.cp
 dai_add_example(rgb_encoding_mono_mobilenet_depth mixed/rgb_encoding_mono_mobilenet_depth.cpp ON)
 dai_add_example(rgb_encoding_mobilenet mixed/rgb_encoding_mobilenet.cpp ON)
 dai_add_example(multiple_devices mixed/multiple_devices.cpp OFF)
+dai_add_example(frame_sync mixed/frame_sync.cpp OFF)
```
themarpe, line 254: Set to ON (whether to test or not) 
OFF option is usually for non-volatile examples like flashing calibration / bootloader / etc... or for examples with specific requirements

themarpe (changes requested):

examples/mixed/frame_sync.cpp
```diff
@@ -0,0 +1,120 @@
...
+                            if check_sync(frames, f.getTimestamp()): # Check if we have any synced frames
+                                # Frames synced!
+                                node.warn(f"Synced frame!")
+                                for name, list in frames.items():
+                                    syncedF = list.pop(0) # We have removed older (excess) frames, so at positions 0 in dict we have synced frames
+                                    node.warn(f"{name}, ts: {str(syncedF.getTimestamp())}, seq {str(syncedF.getSequenceNum())}")
+                                    node.io[name+'_out'].send(syncedF) # Send synced frames to the host
+
+
+                    time.sleep(0.001)  # Avoid lazy looping
```
themarpe, line 93: Dedent to be 4spaces after script->...

eg:
```
script->setScript(R"(
    FPS=15
    import time
    import math
    ...
```

Unless clangformat formats it otherwise

themarpe (changes requested):

themarpe (approved):

PR #577, opened 2022-09-05
Updated the versiona nd change log

saching13:


themarpe (approved):
Changelog includes previous versions as well - might want to skip those and or add as separate entries

PR #587, opened 2022-09-14
Respect log level, even if higher than WARN

moratom:


tests/src/logging_test.cpp
```diff
@@ -0,0 +1,133 @@
...
+        case LogLevel::ERR:
+            return spdlog::level::err;
+        case LogLevel::CRITICAL:
+            return spdlog::level::critical;
+        case LogLevel::OFF:
+            return spdlog::level::off;
+    }
+    // Default
+    return defaultValue;
+}
```
themarpe, line 30: spdlog isn't exposed by DepthAI per se - also not necessary.
Should only used the given dai::LogLevel mechanism

themarpe (changes requested):

tests/src/logging_test.cpp
moratom: I didn't know I could cast enum class to integer as well, my bad, I will change it.

themarpe: `std::underlying_type<Enum>` can come handy to not have to guess the actual type (or assume it)
https://en.cppreference.com/w/cpp/types/underlying_type

themarpe (approved):

PR #595, opened 2022-09-22
Added Warp Node

themarpe:
Warp node exposes Warp block mostly as is.
Current limitations are 16B aligned input and output frame sizes, as well as planar formats (no RGB/BGR interleaved and NV12)

TBD - add additional facilities to skip 16B alignment requirement

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -203,6 +203,7 @@ add_library(${TARGET_CORE_NAME}
     src/pipeline/node/StereoDepth.cpp
     src/pipeline/node/NeuralNetwork.cpp
     src/pipeline/node/ImageManip.cpp
+    src/pipeline/node/Warp.cpp
     src/pipeline/node/VideoEncoder.cpp
     src/pipeline/node/DetectionNetwork.cpp
     src/pipeline/node/Script.cpp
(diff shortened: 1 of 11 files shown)


chengguizi:
@themarpe , looks very interesting! A few questions on this new node, compared to the existing ImageManip:

1. Warp node has the option of interpolation for warping operation. ImageManip does not have such an API, so what is the default behaviour in this case?
2. Any reason Warp node does not have color conversion API? Will warp node have similar performance warping color and monochrome images of the same size? Does ImageManip color conversion executes somewhere else?
3. If all I use is warpMesh and resize, would Warp node be marginally faster, as it is a more `as it is` node?

themarpe:
@chengguizi 
1. Bicubic in ImageManip
2. Warp is a "simpler" node, in sense that it only exposes the capabilities of Warp engine mostly as is. That is warping a plane of specified bit depth. That can be u8-16 or FP16, and it applies this over X planes. Mono is faster, as it only processes a single plane. ImageManip color conversion executes on a shave, while Warp doesn't have an accompanying shave to go along.
3.  Yes and no, boils down to usage, but due to some extra processing in ImageManip's case, I'd say that most of that the, that will hold true.

A todo is also to add a proper error checking, right now things don't terminate neatly if a passed frame isn't properly aligned:)

PR #601, opened 2022-10-03
Improve gate logging

moratom:
- Print out the response code and message of requests to gate when they fail
- Promoted logging levels to warn and error when POST-ing to gate fails

### DIFF
--- a/src/device/DeviceBase.cpp
+++ b/src/device/DeviceBase.cpp
@@ -572,8 +572,13 @@ void DeviceBase::init2(Config cfg, const dai::Path& pathToMvcmd, tl::optional<co
         gate = std::make_unique<DeviceGate>(deviceInfo);
 
         // Create and start session
-        gate->createSession();
-        gate->startSession();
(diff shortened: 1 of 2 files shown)


src/device/DeviceBase.cpp
```diff
@@ -572,8 +572,13 @@ void DeviceBase::init2(Config cfg, const dai::Path& pathToMvcmd, tl::optional<co
         gate = std::make_unique<DeviceGate>(deviceInfo);
 
         // Create and start session
-        gate->createSession();
-        gate->startSession();
+        if(!gate->createSession()) {
+            spdlog::error("Could not start the session on gate!");
+        }
+
```
themarpe, line 578: Add a TODO that these will have to be tied together in terms of successful return

themarpe (approved):

src/device/DeviceBase.cpp
moratom: Added. I left it split for now since otherwise my workaround doesn't work :)

PR #602, opened 2022-10-03
Fix typo and add calibration_factory_reset example

moratom:


examples/CMakeLists.txt
```diff
@@ -213,6 +213,7 @@ dai_add_example(calibration_flash_version5 calibration/calibration_flash_v5.cpp
 dai_add_example(calibration_load calibration/calibration_load.cpp ON)
 dai_add_example(calibration_reader calibration/calibration_reader.cpp ON)
 dai_add_example(calibration_dump calibration/calibration_dump.cpp ON)
+dai_add_example(calibration_factory_reset calibration/calibration_factory_reset.cpp ON)
```
themarpe, line 216:
```suggestion
dai_add_example(calibration_factory_reset calibration/calibration_factory_reset.cpp OFF)
```

We don't want to factory reset calibration as part of a testing suite (only stateless programs preferably run as part of the testing suite)

themarpe (changes requested):
Thanks!

themarpe (approved):

PR #603, opened 2022-10-03
Add a setter and a getter directly for EEPROM data

moratom:
Needed for RVC3 eeprom port.

### DIFF
--- a/include/depthai/pipeline/ThreadedNode.hpp
+++ b/include/depthai/pipeline/ThreadedNode.hpp
@@ -1,6 +1,7 @@
 #pragma once
 
 #include "depthai/pipeline/Node.hpp"
+#include "depthai/pipeline/Pipeline.hpp"
 #include "depthai/utility/AtomicBool.hpp"
 #include "depthai/utility/JoiningThread.hpp"
 #include <spdlog/async_logger.h>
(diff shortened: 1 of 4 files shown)


include/depthai/pipeline/ThreadedNode.hpp
```diff
@@ -1,6 +1,7 @@
 #pragma once
 
 #include "depthai/pipeline/Node.hpp"
+#include "depthai/pipeline/Pipeline.hpp"
```
moratom, line 4: I made this change, because otherwise all nodes need to include the `Pipeline.hpp` to use `getParentPipeline()` since it is just forward declared in `Node.hpp` to avoid circular includes.

src/pipeline/Pipeline.cpp
```diff
@@ -657,6 +661,22 @@ CalibrationHandler PipelineImpl::getCalibrationData() const {
...
+    return globalProperties.calibData.has_value();
+}
+
+EepromData PipelineImpl::getEepromData() const {
+    if(globalProperties.calibData) {
+        return globalProperties.calibData.value();
+    } else {
+        return EepromData{};
+    }
+}
```
themarpe, line 678:
```suggestion
tl::optional<EepromData> PipelineImpl::getEepromData() const {
    return globalProperties.calibData;
}
```

For this one, as we don't have to retain API

themarpe (changes requested):

themarpe (changes requested):
Just to be complete - the suggested change will allow to 'unset' EepromData with eg:
```
pipeline.setEepromData(tl::nullopt_t)
```
or empty optional (AFAIK)
```
pipeline.setEepromData({})
```

PR #609, opened 2022-10-14
Expand spatial image detections with original config

szabi-luxonis:
Related: https://github.com/luxonis/depthai-shared/pull/109


themarpe (approved):
No examples to modify in core?

Issue #355, opened 2022-01-29
`CalibrationHandler` missing `const` on member functions - cascades to apps


diablodale:
Throughout `CalibrationHandler` the member functions are not declared `const` which prevents apps from using const calibration handlers. This cascades up to app classes containing a CalibrationHandler to be non-const, and further up the chain. Or cascades sideways to force app member functions to be non-const so they can call these handler functions. It metastasizes.

The hope of related https://github.com/luxonis/depthai-core/issues/10 didn't happen. 😢

### Setup

* all

### Repro

Found while writing my app. And then `depthai-core` code review.
Here are four examples of errant functions in CalibrationHandler... there are substantially more.

```c++
dai::CameraBoardSocket CalibrationHandler::getStereoRightCameraId();
float CalibrationHandler::getBaselineDistance(...);
std::vector<std::vector<float>> CalibrationHandler::getCameraIntrinsics(...);
float CalibrationHandler::getFov(...);
...
```

### Fix

Often just adding `const` to the end of the member function signatures. There maybe a small subset which need minor changes. This is less than a day's work.

### Workaround

None pretty. It might include making things `mutable` and removing `const` from member functions from apps, which then makes the app's code harder to maintain. :-/

chengguizi:
This would be very helpful! Personally encountered this issue as well.

themarpe:
Addressed in latest `develop`:)

PR #614, opened 2022-10-19
Release v2.18.0

themarpe:
## Features
 - Improved multiple camera support (`multi_cam_support`)
 - User Bootloader support for Network bootloader type
 - Yolov6R2 support
 - ImageManip improvements
 - Warp Node
 - Timesync configurability 

## Bug fixes 
- Added monitor to DeviceBootloader
- #355 

## Misc
 - Added capability to query Bootloader version from Device (if it was booted by it)


developer-mayuan:
Hi @themarpe : Thanks for releasing this new version of depthai-core with improved multicam support. I'm wondering if you have added support to use  vo9782 sensors on LEFT/RIGHT socket as well?

alex-luxonis:
@developer-mayuan Yes, this should be supported. You can try from `depthai-python`:
`python3 utilities/cam_test.py -cams rgb,c left,c right,c` -- all 3 cams as ColorCamera (`c`) at 1280x800 resolution.

Enabling only `left` and `right` (B and C on OAK-FFC-4P) should work as well, but I just noticed now an issue with the resolution, need to specify 720p (instead of 800p default), I'll fix that ASAP
`python3 utilities/cam_test.py -cams left,c right,c -cres 720`

developer-mayuan:
@alex-luxonis Yes, you are right, just verified what you mentioned and it did work. My pipeline was left + right with 800p and I got firmware error, that's why I asked my previous question. Pipeline works now with 720p resolution. Glad to know you are aware of the resolution issue and going to fix it soon. Please just let me when the fix is ready, thanks!

PR #617, opened 2022-10-24
get/set camera FrameEvent filter

whoactuallycares:


themarpe (approved):
Lgtm, just remove the enforce check off and document the new functions briefly

PR #619, opened 2022-10-25
S3 spatial nodes

moratom:
Expose image transformations for spatial nodes


themarpe (approved):
LGTM if we go with ImgTransformation being its own message

themarpe:
@moratom 
Lets move the code for this case into ImgTransformation to not overcomplicate too much (seems to be limited scope).
Also, rename the common from Raw to just ImgTransformation.

PR #645, opened 2022-11-21
Fixed check for user-provided toolchain file when given relative path

### LINKED ISSUE #644: [BUG] Unable to specify custom toolchain file using relative path
depthai-core uses its own custom toolchain file which is supposed to utilize the user-specified toolchain file if one is given. To do this, it first checks if said toolchain file exists.
```
    if(NOT EXISTS "@CMAKE_TOOLCHAIN_FILE@")
      message(
          FATAL_ERROR "File not found: '@CMAKE_TOOLCHAIN_FILE@'"
      )
    endif()
```
According to the CMake docs, "[EXISTS] is well-defined only for explicit full paths (a leading ~/ is not expanded as a home directory and is considered a relative path)". In my limited experience, the result is that I am unable to pass a relative file at all. `-DCMAKE_TOOLCHAIN_FILE=toolchain.cmake`, `./toolchain.cmake` (relative to project root), and `../../toolchain.cmake` (relative to generated toolchain) all cause CMake to trip on the above error message. If the full path is provided, then the generated toolchain is able to successfully find the user-specified file.

It's a minor inconvenience with a known workaround, but I would still expect to be able to pass a relative path for the toolchain.

Tested on CMake 3.16 and 3.24

camblomquist:
`if(EXISTS path-to-file-or-directory)` was used to check for the existence of a user-provided toolchain file. According to the CMake docs, "[EXISTS] is well-defined only for explicit full paths (a leading \~/ is not expanded as a home directory and is considered a relative path)". This leads to two problems. First, we cannot pass a relative path as a toolchain as it may not be detected correctly. Second, we cannot use `if(IS_ABSOLUTE path)` to enforce an absolute path or correct a relative one because "On non-Windows hosts, any path that begins with a tilde (\~) evaluates to true." (It appears the left hand of Kitware doesn't know what the right hand is doing...)

To fix this, use `get_filename_component` to canonicalize a relative path while keeping absolute paths unchanged. If the file doesn't exist, CMake will exit with a more obnoxious error than what came before it. Unfortunately, I don't think there's much to be done about that.

`@CMAKE_CURRENT_SOURCE_DIR@` is passed in as the base for relative paths, forcing the path relative to the depthai-core root when the toolchain file is generated. This is less than ideal if the library is configured as part of a larger project using `add_subdirectory`, but I suppose not every workflow can be accommodated perfectly. 
This may also appear strange as the default base is `CMAKE_CURRENT_SOURCE_DIR`. It turns out that the custom toolchain file is invoked multiple times by CMake and that the value of `CMAKE_CURRENT_SOURCE_DIR` will change in at least one invocation. This explains why it appeared that relative paths would break completely rather than just finding some path that would work as expected.

Closes #644

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,6 +6,9 @@ if(NOT WIN32)
     set(HUNTER_CONFIGURATION_TYPES "Release" CACHE STRING "Hunter dependencies list of build configurations")
 endif()
 
+# Set type to canonicalize relative paths for user-provided toolchain
+set(CMAKE_TOOLCHAIN_FILE "" CACHE FILEPATH "CMake toolchain path")
+
 # Create a custom toolchain to pass certain options to dependencies
 set(gen_toolchain "${CMAKE_CURRENT_BINARY_DIR}/generated/toolchain.cmake")
 
--- a/cmake/toolchain/custom.cmake.in
+++ b/cmake/toolchain/custom.cmake.in
@@ -26,12 +26,11 @@ if(NOT "@CMAKE_TOOLCHAIN_FILE@" STREQUAL "")
   if("@_INTERNAL_DEPTHAI_ORIGINAL_CMAKE_TOOLCHAIN_FILE@" STREQUAL "")
     # Variable not set, initial project configuration.
     # Path to original toolchain saved in CMAKE_TOOLCHAIN_FILE.
-    if(NOT EXISTS "@CMAKE_TOOLCHAIN_FILE@")
-      message(
-          FATAL_ERROR "File not found: '@CMAKE_TOOLCHAIN_FILE@'"
-      )
-    endif()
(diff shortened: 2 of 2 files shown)


themarpe (commented):
WRT to relative, I think the following could be used actually:
```
set(CMAKE_TOOLCHAIN_PATH "" CACHE FILEPATH "CMake toolchain path")
```
Which does seem to correctly convert to absolute path beforehand. (And might also resolve `~`)
Eg pre line `CMakeLists.txt:10`

Do you mind giving it a try?

I think at the moment changing toolchain after it was once generated isn't possible. Would need a closer look..

themarpe:
Great!
Feel free to reuse this PR - I'll squash merge afterwards.

> That said, it appears that changing this would require that the logic for detecting the user toolchain be moved into the main CMakeLists. This is also outside the scope of the Issue.

If you do happen to stumble upon a fix, feel free to add it in / open a PR. Thanks!

PR #646, opened 2022-11-22
Move hunter to install in the build directory

PeterQFR:
Adding a build dependency to a relatively random location in the user's home directory is, well, a bit random and causes a couple of issues.
A clean build required removal of that directory
It clutters the users home directory
in cases where the home directory is always new (such as building in a docker container) requires rebuild of the hunter package at each build. (assuming that the build directory is bind mounted to the build container)

This change moves the default location for the hunter archive and make to the build directory.

This has been tested on the main branch, but merging into develop branch.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -27,6 +27,7 @@ else()
     message(STATUS "Using toolchain file: ${CMAKE_TOOLCHAIN_FILE}")
 endif()
 
+set(HUNTER_ROOT ${CMAKE_CURRENT_BINARY_DIR})
 include("cmake/HunterGate.cmake")
 HunterGate(
     URL "https://github.com/cpp-pm/hunter/archive/v0.23.322.tar.gz"
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -27,10 +27,7 @@ else()
     message(STATUS "Using toolchain file: ${CMAKE_TOOLCHAIN_FILE}")
 endif()
 
-if(NOT WIN32)
-    set(HUNTER_ROOT ${CMAKE_CURRENT_BINARY_DIR}/hunter)
-endif()
-
+set(HUNTER_ROOT ${CMAKE_CURRENT_BINARY_DIR}/hunter)
 include("cmake/HunterGate.cmake")
 HunterGate(
     URL "https://github.com/cpp-pm/hunter/archive/v0.23.322.tar.gz"
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -27,6 +27,11 @@ else()
     message(STATUS "Using toolchain file: ${CMAKE_TOOLCHAIN_FILE}")
 endif()
 
+# Move binary dir if windows, to shorten the path
+if(WIN32)
+    set(HUNTER_BINARY_DIR "${HUNTER_GATE_ROOT}/_bin" CACHE STRING "Hunter binary directory")
+endif()
+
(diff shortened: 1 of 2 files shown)


themarpe (commented):
Thanks for the PR.

I agree - thought its not completely off that such option was chosen by Hunter devs.

One thing that might be breaking, is whether Windows build goes through. AFAIK the binary dir is set separately there and that being out of root and retained elsewhere is all good, then LGTM

Issue with Windows in this case is that max length of paths is too short causing issues.

CMakeLists.txt
```diff
@@ -27,6 +27,7 @@ else()
     message(STATUS "Using toolchain file: ${CMAKE_TOOLCHAIN_FILE}")
 endif()
 
+set(HUNTER_ROOT ${CMAKE_CURRENT_BINARY_DIR})
```
themarpe, line 30: Currently the hunter root contains just a `_Base` folder, which is fine as it likely won't clash with any build artifacts, but might be tad more descriptive setting that to `[build]/hunter` (or .hunter, though that hides it again, likely better to be more visible)

themarpe (changes requested):

CMakeLists.txt
PeterQFR: Yeah, good idea.

themarpe:
https://github.com/luxonis/depthai-core/actions/runs/3520321187/jobs/5918364058

Seems to do trip up Windows unfortunately...

PeterQFR:
I dont have a windows workflow so can't help too much. What are the issues?

themarpe:
> One thing that might be breaking, is whether Windows build goes through. AFAIK the binary dir is set separately there and that being out of root and retained elsewhere is all good, then LGTM
Issue with Windows in this case is that max length of paths is too short causing issues.

I think it hits this again. Binary dir would have to be moved. Checking...

themarpe:
```
# Move binary dir if windows, to shorten the path
if(WIN32)
    set(HUNTER_BINARY_DIR "${HUNTER_GATE_ROOT}/_bin" CACHE STRING "Hunter binary directory")
endif()
```

It happens afterwards. Thus moving it to binary dir as well.

For Windows, lets save the HUNTER_GATE_ROOT, then set it to build dir, but set the HUNTER_BINARY_DIR to be the previously saved `ROOT/_bin`

That'd move binary dir out of build, but likely that if root is cleared that bin isn't taken into account anymore. Though I'm not 100% sure.

PeterQFR:
Maybe add a check for linux for setting the HUNTER_ROOT eg 
if (not WIN32)
set(HUNTER_ROOT .....?

themarpe:
I'd not deviate between these two flows. As we then have to document clearing hunter cache separately for Windows.

Lets just move the binary dir, but retain the Hunter root where it is. (I guess just moving the following before setting new HUNTER_ROOT will do)
```
# Move binary dir if windows, to shorten the path
if(WIN32)
    set(HUNTER_BINARY_DIR "${HUNTER_GATE_ROOT}/_bin" CACHE STRING "Hunter binary directory")
endif()
```

Also can remove clearing hunter cache from README in this case, so the PR is complete

And I think while we are at it, I'd detect if `BUILD_SHARED_LIBS=ON` and create a separate ROOT for it (as on Linux due to not detecting -fpic options, builds fail).

themarpe:
@PeterQFR CI is still failing. I think it be best to pindown this on a Windows machine / VM to be sure.
I unfortunately won't have time to dig into it for a the near future:/

PR #648, opened 2022-11-23
Changes made for RVC3 integration

moratom:


src/pipeline/Node.cpp
```diff
@@ -152,7 +162,7 @@ void Node::Output::unlink(Input& in) {
 }
 
 void Node::Output::send(const std::shared_ptr<ADatatype>& msg) {
-    for(auto& conn : parent.connections) {
```
moratom, line 155: This line here was the reason for the bug we discussed on Slack @themarpe, each output
was trying to send messages to inputs of all outputs of a particular node, not just its own.

themarpe: Thanks for the find!

themarpe (approved):

PR #653, opened 2022-12-02
Split detection nodes

moratom:
Split detection nodes + add an interface for passing custom settings to NN node.
Maybe it's best, that we first merge `message_pipeline_improvements` to rvc3, for the PR to be more "reviewable". 

The split nodes are somewhat ugly, as the interface is copied from previous monolithic nodes - this is done to keep
backwards compatibility, we can depreciate some functions later and move the examples to set the more "specific"
properties on the nodes directly.


include/depthai/pipeline/node/DetectionParser.hpp
```diff
@@ -17,24 +17,23 @@ namespace node {
...
     /**
      * Outputs image frame with detected edges
      */
-    Output out{*this, "out", Output::Type::MSender, {{DatatypeEnum::ImgDetections, false}}};
+    Output out{true, *this, "out", Output::Type::MSender, {{DatatypeEnum::ImgDetections, false}}};
+
+    /**
+     * Input for image that produced the detection - image size can be taken from here
+     */
+    Input imageIn{true, *this, "imageIn", Input::Type::SReceiver, {{DatatypeEnum::ImgFrame, false}}};
```
themarpe, line 36: `inputImage` and define the defaults. I assume those would likely be same as `input` (blocking, 5 queue size)

include/depthai/pipeline/node/DetectionParser.hpp
```diff
@@ -55,6 +54,14 @@ class DetectionParser : public NodeCRTP<DeviceNode, DetectionParser, DetectionPa
      */
     void setBlob(const OpenVINO::Blob& blob);
 
+    /**
+     * Set input image size
+     *
+     * This should only be used instead of setBlob, not besides it
+     * @param size Size of the input image (needed for yolo)
+    */
+   void setInputImgSize(std::vector<unsigned> size);
```
themarpe, line 63: setInputSize / setInputImageSize & (int w, int h) & (std::tuple<int,int> size)

include/depthai/pipeline/node/NeuralNetwork.hpp
```diff
@@ -120,6 +118,12 @@ class NeuralNetwork : public NodeCRTP<DeviceNode, NeuralNetwork, NeuralNetworkPr
      */
     void setBackend(std::string backend);
 
+    /**
+     * Set custom settings
+     * @param settings Settings map
+     */
+    void setCustomSettings(std::map<std::string, std::string> settings);
```
themarpe, line 125:
```suggestion
    /**
     * Set custom settings
     * @param settings Settings map
     */
    void setBackendProperties(std::map<std::string, std::string> settings);
```

themarpe (changes requested):
Remove `NodeGroup.md`, otherwise okay for now, we'll tweak xtensor submodules later

PR #678, opened 2023-01-12
Rvc3 develop bugfixes

moratom:


src/pipeline/datatype/NNData.cpp
```diff
@@ -48,8 +48,13 @@ static std::size_t getTensorDataSize(const TensorInfo& tensor) {
...
     // get data from u8Data and fp16Data and place properly into the underlying raw buffer
-    rawNn.tensors = {};
-    auto mem = std::make_shared<VectorMemory>();
+    std::shared_ptr<VectorMemory> mem;
+    if(std::dynamic_pointer_cast<VectorMemory>(data) == nullptr) {
+        auto prev = std::vector<uint8_t>(data->getData().begin(), data->getData().end());
+        mem = std::make_shared<VectorMemory>(std::move(prev));
+    } else {
+        mem = std::dynamic_pointer_cast<VectorMemory>(data);
+    }
```
moratom, line 57: Needed to be able to use the APIs (setTensor and setLayer interchangeably, otherwise serialize() just deletes throws away all memory that was set with setTensor())

themarpe (approved):

PR #690, opened 2023-01-24
Rvc3 detection parser refactor

moratom:
Add an option to set a NNData tensor in place.
This API greatly improves the performance on RVC3.


include/depthai/pipeline/datatype/NNData.hpp
```diff
@@ -295,14 +300,33 @@ class NNData : public Buffer {
...
+            case TensorInfo::DataType::FP16:
+                for(uint32_t i = 0; i < tensor.size(); i++) {
+                    tensor.data()[i] = fp16_to_fp32(reinterpret_cast<uint16_t*>(data->getData().data())[it->offset / sizeof(uint16_t) +  i]);
+                }
+                break;
+            case TensorInfo::DataType::FP32:
+                for(uint32_t i = 0; i < tensor.size(); i++) {
+                    tensor.data()[i] = reinterpret_cast<float_t*>(data->getData().data())[it->offset / sizeof(float_t) + i];
+                }
+                break;
```
moratom, line 329: I generalized the `getTensor()` since the new API can set the tensors of any supported type into memory.

src/pipeline/datatype/NNData.cpp
```diff
@@ -149,6 +149,27 @@ NNData& NNData::setLayer(const std::string& name, const std::vector<int>& data)
     return *this;
 }
 
+NNData& NNData::setLayerInPlace(TensorInfo& tensor, uint8_t*& dataIn, size_t& size){
```
themarpe, line 152: Might consider naming this as "Tensor" (perhaps add / emplace / ...) and returning a span (either as ret value our output parameter). Does xtensor allow span line view over existing memory? Could be reused

themarpe (approved):
LGTM otherwise
Thanks!

src/pipeline/datatype/NNData.cpp
themarpe: Perhaps mark for pre RVC3 merge

moratom: Good point, I'll refactor this so it return a span.
Xtensor does allow a span like view over existing memory, but I think we should stick to `TensorInfo` here, since we need to pass other information about the tensor than what `xtensor` can hold anyways (quantization, data order, etc).

themarpe: I meant xtensor as in span, which we can already construct it to be of certain size and type, etc... - TensorInfo would remain the "in place" operation param

moratom: [xtarray_adaptor](https://xtensor.readthedocs.io/en/latest/api/xarray_adaptor.html) might be a good fit here, but the problem with using it is that it uses templates for deducing the underlying type, while we (`TensorInfo`) are handling types in runtime.

 I don't think there's an easy way (at least not with this passing of `TensorInfo` as an input parameter, which contains type information) of "merging" both. We can either template our inPlace function as well (but would have to split the type information out of `TensorInfo`) and add or return a simple `uint8_t` span.
 
 I'll note down for the xtensor_adaptor to be checked again pre RVC3 merge but will just return a regular span for now.

PR #704, opened 2023-01-28
Removed possible spdlog calls from destructors

themarpe:
Can help eliviate some segfaults when `Device` is a global object.

TODO: Proper fix on spdlog side


themarpe:
Resolved in 870f1947427ec13030d830a08c9919e84232dc98

PR #724, opened 2023-02-10
Crash dump support for debugging

szabi-luxonis:
Outputs stack trace of threads from the device in case of a crash in json format, parsable by `addr2line`.
TODO: tool

### DIFF
--- a/include/depthai/device/DeviceBase.hpp
+++ b/include/depthai/device/DeviceBase.hpp
@@ -457,6 +457,11 @@ class DeviceBase {
     dai::CrashDump getCrashDump();
 
     /**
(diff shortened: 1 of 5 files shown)


include/depthai/device/DeviceBase.hpp
```diff
@@ -450,6 +451,11 @@ class DeviceBase {
      */
     std::vector<std::tuple<std::string, int, int>> getIrDrivers();
 
+    /**
+     * Retrieves crash dump for debugging.
+     */
+    dai::CrashDump getCrashDump();
```
themarpe, line 457: add `bool hasCrashDump();`

themarpe (changes requested):

Issue #744, opened 2023-03-03
[BUG] IMU firmware updater re-writes duplicate versions of firmware, no API to check versions, log spelling issue


diablodale:
The IMU firmware updater does not check for duplicate versions of firmware. It blindly writes the same version over itself.
The updater, EEPROM apis, IMU node, etc. do not have any capability to query for the versions or know new firmware is available.

### Setup

* depthai-core v2.20.2
* all platforms
* all build tools
* OAK-D

### Repro 1: enableFirmwareUpdate() blindly writes new firmware

`imu->enableFirmwareUpdate(true)` writes new firmware even if the IMU already has the same firmware

1. Set log level to INFO.
2. Run an app that uses IMU

```
[14442C10C1A1C6D200] [3.12] [0.871] [IMU(0)] [info] IMU product ID:
[14442C10C1A1C6D200] [3.12] [0.871] [IMU(0)] [info] Part 10004148 : Version 3.9.7 Build 224
[14442C10C1A1C6D200] [3.12] [0.871] [IMU(0)] [info] Part 10003606 : Version 1.7.3 Build 337
[14442C10C1A1C6D200] [3.12] [0.871] [IMU(0)] [info] Part 10004135 : Version 5.5.1 Build 159
[14442C10C1A1C6D200] [3.12] [0.871] [IMU(0)] [info] Part 10004149 : Version 5.1.11 Build 182
```

3. Write c++ app that uses `imu->enableFirmwareUpdate(true)`
4. Run that firmware app

```
[14442C10C1A1C6D200] [3.12] [0.684] [IMU(0)] [warning] Performing IMU firmware update, do NOT disconnect the device.
[14442C10C1A1C6D200] [3.12] [0.793] [IMU(0)] [warning] IMU firmware update status: 0%
[14442C10C1A1C6D200] [3.12] [4.021] [IMU(0)] [warning] IMU firmware update status: 10%
[14442C10C1A1C6D200] [3.12] [7.244] [IMU(0)] [warning] IMU firmware update status: 20%
[14442C10C1A1C6D200] [3.12] [10.470] [IMU(0)] [warning] IMU firmware update status: 30%
[14442C10C1A1C6D200] [3.12] [13.708] [IMU(0)] [warning] IMU firmware update status: 40%
[14442C10C1A1C6D200] [3.12] [16.932] [IMU(0)] [warning] IMU firmware update status: 50%
[...]

Issue #683, opened 2023-01-17
[Feature-Request] Add ability to query IMU model (BNO/BMI) from IMU Node


chengguizi:
### Start with the `why`:
There are functional difference between the two IMUs, so many times a different handling logic is required. At the same time, due to parts shortage, the same product may carry different IMUs depending on the batch. However, there seems no good way to query the IMU installed on PCB in the software API. We should have one!

### Move to the `what`:
Add a API output the currently detected IMU by the VPU firmware. Exposed through IMU Node API

### Move to the `how`:
Probably A firmware feature with depthai-core API addition. Hopefully not much work needed!

Related discussion: https://discord.com/channels/790680891252932659/924797392115601439/1061121384849354864

themarpe:
+1 on the request, its in the pipeline to be added

chengguizi:
Any update on this yet? :)

szabi-luxonis:
> Any update on this yet? :)

This is a low priority task at the moment, once I get to it, this issue will be updated.

szabi-luxonis:
Added in https://github.com/luxonis/depthai-core/commit/60e7ee56361efa7e50cd42081f0249cac898202e

Issue #520, opened 2022-07-12
inquiry into mutex locking changes in `Device` and `Queue`


diablodale:
I'm unsure of some mutex locking changes in commit https://github.com/luxonis/depthai-core/commit/55066c5078f47eea588433945c0fec2840cbe12a

### `dai::DeviceBase::closed` changed from `atomic<bool>` to pair of `std::mutex` + `bool`

The two places I find these vars are used `DeviceBase::isClosed()` and `DeviceBase::close()` correctly implement the mutex+bool pair.
My concern is that there is no known case for this code change and no functional bug.
This change adds code complexity, more cpu, and additional memory use when not needed.
And finally, it makes a substantial behavior change. This change creates a blocking behavior where the original code did not.
Now multiple threads which call these functions will block rather than quickly returning.
Since this is a blocking behavior change, this is a risky change for no known need.

When I refactored this code months ago, I had my head around the locking needs of the multiple threads running. At that time I felt the `atomic<bool>` was the better choice for simplicity and its non-blocking behavior. It didn't matter which thread closed a device; it only mattered that *one* thread ran the close code. I acknowledge that I can be wrong and write bugs. But that isn't the case here as there is no known bug and no imagined case.

```c++
int DeviceBase::getXLinkChunkSize() {
    checkClosed();

    return pimpl->rpcClient->call("getXLinkChunkSize").as<int>();
}
```

`checkClosed()` there is useless.
It calls `isClosed()` which locks the mutex and returns the bool value of `closed`.
If it is `true` then `checkClosed()` throws which then prevents the call to `pimpl->rpcClient->call()`.
But the checking of the value and the throw are disconnected making the test useless.

Time | Thread 1 | Thread 2
-- | -- | --
1 | DeviceBase::getXLinkChunkSize() |  
2 | checkClosed() returns `false` |  
3 | …os pauses thread… | DeviceBase::close()
4 | …os pauses thread… | std::unique_lock<std::mutex> lock(closedMtx);
5 | …os pauses thread… | if(!closed) {
6 | …os pauses thread… | closeImpl();
7 | …os pauses thread… | closed = true;
8 | pimpl->rpcClient->call("getXLinkChunkSize").as<int>(); |  
9 | call(…) |  
10 | 💥 or something fails/bad/errors/etc |  

[...]

PR #752, opened 2023-03-06
Add ability to query IMU model (BNO/BMI) from IMU Node + firmware version, firmware update status

szabi-luxonis:
Related: 
https://github.com/luxonis/depthai-core/issues/744
https://github.com/luxonis/depthai-core/issues/683

### DIFF
--- a/examples/IMU/imu_firmware_update.cpp
+++ b/examples/IMU/imu_firmware_update.cpp
@@ -15,9 +15,9 @@ int main() {
 
     auto imuType = device.getConnectedIMU();
     auto imuFirmwareVersion = device.getIMUFirmwareVersion();
-    auto latestImuFirmwareVersion = device.getEmbeddedIMUFirmwareVersion();
+    auto embeddedIMUFirmwareVersion = device.getEmbeddedIMUFirmwareVersion();
     std::cout << "IMU version: " << imuType << " firmware version: " << imuFirmwareVersion
-              << " latest available firmware version: " << latestImuFirmwareVersion << std::endl;
+              << " latest available firmware version: " << embeddedIMUFirmwareVersion << std::endl;
 
     std::cout << "Warning! Flashing IMU firmware can potentially soft brick your device and should be done with caution." << std::endl;
     std::cout << "Do not unplug your device while the IMU firmware is flashing." << std::endl;
--- a/examples/IMU/imu_firmware_update.cpp
+++ b/examples/IMU/imu_firmware_update.cpp
@@ -16,8 +16,8 @@ int main() {
     auto imuType = device.getConnectedIMU();
     auto imuFirmwareVersion = device.getIMUFirmwareVersion();
     auto embeddedIMUFirmwareVersion = device.getEmbeddedIMUFirmwareVersion();
-    std::cout << "IMU version: " << imuType << " firmware version: " << imuFirmwareVersion
-              << " latest available firmware version: " << embeddedIMUFirmwareVersion << std::endl;
+    std::cout << "IMU type: " << imuType << " firmware version: " << imuFirmwareVersion
+              << " embedded firmware version: " << embeddedIMUFirmwareVersion << std::endl;
 
     std::cout << "Warning! Flashing IMU firmware can potentially soft brick your device and should be done with caution." << std::endl;
     std::cout << "Do not unplug your device while the IMU firmware is flashing." << std::endl;
--- a/include/depthai/device/DeviceBase.hpp
+++ b/include/depthai/device/DeviceBase.hpp
@@ -29,6 +29,7 @@
 #include "depthai-shared/common/ChipTemperature.hpp"
 #include "depthai-shared/common/CpuUsage.hpp"
 #include "depthai-shared/common/MemoryInfo.hpp"
+#include "depthai-shared/datatype/RawIMUData.hpp"
 #include "depthai-shared/device/BoardConfig.hpp"
 #include "depthai-shared/device/CrashDump.hpp"
 #include "depthai-shared/log/LogLevel.hpp"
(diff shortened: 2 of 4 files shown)


themarpe (approved):
:tada:

include/depthai/device/DeviceBase.hpp
```diff
@@ -513,6 +514,47 @@ class DeviceBase {
      */
     std::unordered_map<CameraBoardSocket, std::string> getCameraSensorNames();
 
+    /**
+     * Get connected IMU version
+     *
+     * @returns IMU version
+     */
+    std::string getConnectedIMU();
```
diablodale, line 522: I suggest do not use the word "version" in this API or it's documentation. For me, this is a chip model or component. Not a version.

include/depthai/device/DeviceBase.hpp
```diff
@@ -513,6 +514,47 @@ class DeviceBase {
...
+    /**
+     * Starts IMU firmware update asynchronously only if IMU node is not running.
+     * If current firmware version is the same as latest available firmware version that it's no-op. Can be overridden by forceUpdate parameter.
+     * State of firmware update can be monitored using getIMUFirmwareUpdateStatus API.
+     *
+     * @param forceUpdate Force firmware update or not. Will perform FW update regardless of current version and latest available firmware version.
+     *
+     * @returns Returns whether firmware update can be started. Returns false if IMU node is started.
+     */
+    bool startIMUFirmwareUpdate(bool forceUpdate = false);
```
diablodale, line 547: My opinion-> allow downgrades easily.  There have already been dot releases of depthai-core to reverse changes of various things. I think it wiser to plan also for bad IMU firmware.

With that in mind, the concept and API for "latest" is somewhat unclear. For example...

Good IMU firmware is v12
Depthai-core includes v13
Developer burns v13 to IMU
Depthai-core update goes back to v12
How does dev get v12?

Linguistically, the "latest" is v13. Maybe consider changing the API to express the idea of "recommended" or "available" or "canonical"

include/depthai/device/DeviceBase.hpp
```diff
@@ -513,6 +514,47 @@ class DeviceBase {
...
+
+    /**
+     * Get IMU firmware update status
+     *
+     * @returns Whether IMU firmware update is pending and last firmware update progress as percentage.
+     * return value false and 100 means that the update was successful
+     * return value false and other than 100 means that the update failed
+     */
+    std::tuple<bool, float> getIMUFirmwareUpdateStatus();
```
diablodale, line 556: Personally, I prefer a callback. Same approach used for queue callbacks.

I suggest also to reverse the boolean logic. As it is itoday it is negative logic. The developer wants to know if it is complete. When complete it should be true. But today's logic when complete it is false.

include/depthai/device/DeviceBase.hpp
```diff
@@ -513,6 +514,47 @@ class DeviceBase {
...
+     * @returns IMU version
+     */
+    std::string getConnectedIMU();
+
+    /**
+     * Get connected IMU firmware version
+     *
+     * @returns IMU firmware version
+     */
+    dai::Version getIMUFirmwareVersion();
```
diablodale, line 529: When getting the IMU version already burned and the IMU version depthai-core has available for burning... What is this version? When I look at the log, I can see 3 or 4 parts of my IMU, and each part has it's own version number.

Maybe a version, like a server, it not needed. Instead, we only need a hash. I can imagine firmware bundles are by hash. This removes the new, old, up, down linguistics. And instead there is only the currently burned hash and the hash available/recommended in depthai-core.

include/depthai/device/DeviceBase.hpp
themarpe: On this note, could also be named "getEmbeddedIMUFirmwareVersion` as in DeviceBootloader

include/depthai/device/DeviceBase.hpp
szabi-luxonis: IMU firmware is part of the FW binary. On update we update the IMU FW binary, together with the FW version. We have full control of IMU FW version, it's not part of the binary we receive from the vendor, but part of a header. See [here](https://github.com/hcrest/bno080-nucleo-demo/blob/master/Hillcrest/firmware.c#L56)
Therefore `dai::Version` is enough to distinguish between different versions.

include/depthai/device/DeviceBase.hpp
szabi-luxonis: Agreed, reverted logic.
+1 on callback, it would be nice to have. A thread could be spun up, listening to FW update request. (That's how it's done in FW, so we don't block RPC call)
However it's preferred to have async API, in production we test the cameras etc. while doing IMU FW update in parallel.
We might add later callback support, only host-side changes are needed for that, everything is in place.

include/depthai/device/DeviceBase.hpp
szabi-luxonis: Renamed to getEmbeddedIMUFirmwareVersion

include/depthai/device/DeviceBase.hpp
szabi-luxonis: Updated description

examples/IMU/imu_firmware_update.cpp
```diff
@@ -11,6 +11,14 @@ int main() {
     using namespace std;
     using namespace std::chrono;
 
+    dai::Device device;
+
+    auto imuType = device.getConnectedIMU();
+    auto imuFirmwareVersion = device.getIMUFirmwareVersion();
+    auto embeddedIMUFirmwareVersion = device.getEmbeddedIMUFirmwareVersion();
+    std::cout << "IMU version: " << imuType << " firmware version: " << imuFirmwareVersion
```
diablodale, line 19: recommend changing string "IMU version: " to "IMU type: "

src/device/DeviceBase.cpp
```diff
@@ -821,6 +821,45 @@ std::unordered_map<CameraBoardSocket, std::string> DeviceBase::getCameraSensorNa
     return pimpl->rpcClient->call("getCameraSensorNames").as<std::unordered_map<CameraBoardSocket, std::string>>();
 }
 
+std::string DeviceBase::getConnectedIMU() {
+    checkClosed();
```
diablodale, line 825: All uses of `checkedClosed()` in this PR are not thread-safe and therefore makes it useless. This is bug https://github.com/luxonis/depthai-core/issues/520 that I have on my list to fix.

diablodale: maybe leave them as is. And then I'll fix both `checkClosed()` and all the places it is used in one PR. If there is a thread somewhere else changing state, there will be some random chance that the test+throw won't work correctly and the code after will run while that other thread did what it did

themarpe (approved):

src/device/DeviceBase.cpp
szabi-luxonis: Will leave it as is, at least compliant with the rest of RPCs, so the behaviour is the same

PR #764, opened 2023-03-14
correct thread-safety in several classes

### LINKED ISSUE #520: inquiry into mutex locking changes in `Device` and `Queue`

diablodale:
Work in progress 👀 and I will rebase later

Fixes #520

### DIFF
--- a/include/depthai/device/DataQueue.hpp
+++ b/include/depthai/device/DataQueue.hpp
@@ -43,6 +43,10 @@ class DataOutputQueue {
 
     /**
      * Check whether queue is closed
+     *
+     * @warning This function is thread-unsafe and may return outdated incorrect values. It is
+     * only meant for use in simple single-threaded code. Well written code should handle
+     * exceptions when calling any DepthAI apis to handle hardware events and multithreaded use.
      */
     bool isClosed() const;
 
@@ -343,7 +347,7 @@ class DataInputQueue {
     std::atomic<bool> running{true};
     std::string exceptionMessage;
     const std::string name;
-    std::size_t maxDataSize = device::XLINK_USB_BUFFER_MAX_SIZE;
+    std::atomic<std::size_t> maxDataSize{device::XLINK_USB_BUFFER_MAX_SIZE};
 
    public:
     DataInputQueue(const std::shared_ptr<XLinkConnection> conn,
@@ -355,6 +359,10 @@ class DataInputQueue {
 
     /**
      * Check whether queue is closed
+     *
+     * @warning This function is thread-unsafe and may return outdated incorrect values. It is
+     * only meant for use in simple single-threaded code. Well written code should handle
+     * exceptions when calling any DepthAI apis to handle hardware events and multithreaded use.
      */
     bool isClosed() const;
 
--- a/include/depthai/device/DeviceBase.hpp
+++ b/include/depthai/device/DeviceBase.hpp
@@ -751,6 +751,10 @@ class DeviceBase {
 
     /**
      * Is the device already closed (or disconnected)
+     *
+     * @warning This function is thread-unsafe and may return outdated incorrect values. It is
+     * only meant for use in simple single-threaded code. Well written code should handle
+     * exceptions when calling any DepthAI apis to handle hardware events and multithreaded use.
      */
     bool isClosed() const;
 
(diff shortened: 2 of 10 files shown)


diablodale:
Ready for review.  All the work on this PR is in the first commit "fix isClosed...".

The other two allow for successful test runs. I opened separate PRs for these. So depending on the order in which you review or merge...I can rebase again or the merge with automatically drop duplicate commits.

themarpe (approved):
Thanks, looks good - I'll wait till the other two get in and then bring this one in after those changes are applied

diablodale:
Please wait before merge. I've rebased to remove the two commits that are already separate PRs. And I see the STYLE ci didn't pass. I'd like to run that step locally and see if I can fix them (or if it is something beyond my changes).

I'll force push soon.

diablodale:
force-pushed. I only removed the two commits in other PRs. No other changes.
Ready for your review and hopefully merge.

The style CI fail is happening on code that I didn't write.
I ran clang-format on just the lines in this PR and I didn't get any reports of formatting issues.

themarpe:
Style issue comes from shared code - updating now with also other branches in & style fix
LGTM - if ready, will merge

diablodale:
I have no more code or known changes to add to this PR.

PR #770, opened 2023-03-16
Transformation rewrite

moratom:
Refactored image transformations.
The general idea is that (advanced) users will be able to transform rectangles left and right to the desired image on host as well.
For example get the detections on the ISP frame from a frame that went through the "cam.video->imageManip->nnNode" process and where detections were done on the nnNode.

Would you mind giving it a look architecturally, before I iron out the details @themarpe?


src/pipeline/datatype/ImgFrame.cpp
```diff
@@ -146,4 +167,306 @@ ImgFrame& ImgFrame::setType(RawImgFrame::Type type) {
...
+
+    // Calculate the VFOV in radians
+    float diagonalFovRadians = 2 * std::atan(tanDiagonalFovHalf);
+
+    // Convert VFOV to degrees
+    float diagonalFovDegrees = diagonalFovRadians * (180.0f / static_cast<float>(M_PI));
+    return diagonalFovDegrees;
+}
+
+float ImgFrame::getSourceVFov() {
```
moratom, line 437: This here depends on the frame being rectilinear, but in case it isn't mappings don't make much sense anyways (unless maybe the intrinsic matrices are (if possible) used to map a point between two warped images).

We would/will have to set the HFov correctly though, depending on which resolution is selected and if the config for the resolution is set to crop. (FOV, might be narrower than specified on some resolutions)

moratom:
Relevant commit for depthai-shared: https://github.com/luxonis/depthai-shared/pull/133

Issue #247, opened 2021-11-09
multiple conflicting/hiding variables and class members -> compiler warnings/fails


diablodale:
In many places of the codebase (including the headers of the precompiled Windows depthai distribution), the code uses duplicate variable names which conflict and/or hide other variables (e.g. class members). Leading to compiler warnings and in some cases compiler failures.

Fix is to not conflict/hide variables with same names. Can be found easily with a compile. And fixed easily by changing conflicting names (e.g. following Google style guide).

### Setup

* all compilers and OS
* throughout codebase, yet can also be seen in the headers of precompiled Windows distrib:  `depthai-core-v2.11.1-win64`

### Repro

1. Enable warnings and errors in compiler. e.g. in MSVC it is with `/W4`
2. Compile an app that uses the `depthai-core-v2.11.1-win64` and simple code that cascades to headers like `common/Rect.hpp` or `pipeline/Node.hpp`. Most any code, even just a dai device, will do this.

### Result

```
...
[build] Microsoft (R) C/C++ Optimizing Compiler Version 19.29.30136 for x64
[build] Copyright (C) Microsoft Corporation.  All rights reserved.
[build] 
[build] C:\repos-nobackup\depthai-core-v2.11.1-win64\include\depthai-shared/common/Rect.hpp(100): error C2220: the following warning is treated as an error
[build] C:\repos-nobackup\depthai-core-v2.11.1-win64\include\depthai-shared/common/Rect.hpp(100): warning C4458: declaration of 'width' hides class member
[build] C:\repos-nobackup\depthai-core-v2.11.1-win64\include\depthai-shared/common/Rect.hpp(129): note: see declaration of 'dai::Rect::width'
[build] C:\repos-nobackup\depthai-core-v2.11.1-win64\include\depthai-shared/common/Rect.hpp(100): warning C4458: declaration of 'height' hides class member
[build] C:\repos-nobackup\depthai-core-v2.11.1-win64\include\depthai-shared/common/Rect.hpp(130): note: see declaration of 'dai::Rect::height'
[build] C:\repos-nobackup\depthai-core-v2.11.1-win64\include\depthai-shared/common/Rect.hpp(116): warning C4458: declaration of 'width' hides class member
[build] C:\repos-nobackup\depthai-core-v2.11.1-win64\include\depthai-shared/common/Rect.hpp(129): note: see declaration of 'dai::Rect::width'
[...]

Issue #766, opened 2023-03-16
`CalibrationHandler.cpp` bugs found in code review


diablodale:
While compiling depthai-core with MSVC and `/W4` several issues are being exposed.
I recommend core devs and CI compile with /W4 so the compiler can help.

### Setup

* VS2019 v16.11.24
* Microsoft Windows 11 [Version 10.0.22621.1413]
* depthai-core `develop` at ddca0b29f672a88a86c2aec9876fef5e33535114

### Repro

1. config and build depthai-core for: windows x64, debug, shared libs, opencv support, build test and examples, test test and examples

### Result

Warnings and errors throughout, here are some specifically for `CalibrationHandler.cpp`. I think they are valid bugs.
```
[build] [20/114  17% :: 4.276] Building CXX object CMakeFiles\depthai-core.dir\src\device\CalibrationHandler.cpp.obj
[build] ..\src\device\CalibrationHandler.cpp(589): warning C4244: '=': conversion from 'int' to 'uint16_t', possible loss of data
[build] ..\src\device\CalibrationHandler.cpp(590): warning C4244: '=': conversion from 'int' to 'uint16_t', possible loss of data
[build] ..\src\device\CalibrationHandler.cpp(594): warning C4244: '=': conversion from 'int' to 'uint16_t', possible loss of data
[build] ..\src\device\CalibrationHandler.cpp(595): warning C4244: '=': conversion from 'int' to 'uint16_t', possible loss of data
[build] ..\src\device\CalibrationHandler.cpp(618): warning C4127: conditional expression is constant
[build] C:\njs\depthai-core\src\utility\matrixOps.hpp(21) : warning C4702: unreachable code
[build] C:\njs\depthai-core\src\device\CalibrationHandler.cpp(386) : warning C4702: unreachable code
[build] ninja: build stopped: subcommand failed.
```

https://github.com/luxonis/depthai-core/blob/e5dae8fd2f9e543bd99f9d1a73959398ef5bd0b4/src/device/CalibrationHandler.cpp#L611-L625

Above it seems the code wants to trim or pad so the coefficients are always 14. The code does not do that. Line 618 will never pad because `num` is a constant. I recommend two changes. 1) make num constexpr. 2) fix the logic on line 618 or entirely remove it.

https://github.com/luxonis/depthai-core/blob/e5dae8fd2f9e543bd99f9d1a73959398ef5bd0b4/src/device/CalibrationHandler.cpp#L376-L387

[...]

PR #772, opened 2023-03-17
fix var hides class member, this-> in constructors

diablodale:
- fix few compile warn 'ex' unreferenced local variable
- rename setBusId() param to not hide class member
- partial fix luxonis/depthai-core#247

### DIFF
--- a/examples/FeatureTracker/feature_tracker.cpp
+++ b/examples/FeatureTracker/feature_tracker.cpp
@@ -72,9 +72,7 @@ class FeatureTrackerDrawer {
         }
     }
 
-    FeatureTrackerDrawer(std::string trackbarName, std::string windowName) {
-        this->trackbarName = trackbarName;
-        this->windowName = windowName;
+    FeatureTrackerDrawer(std::string trackbarName, std::string windowName) : trackbarName(trackbarName), windowName(windowName) {
         cv::namedWindow(windowName.c_str());
         cv::createTrackbar(trackbarName.c_str(), windowName.c_str(), &trackedFeaturesPathLength, maxTrackedFeaturesPathLength, nullptr);
     }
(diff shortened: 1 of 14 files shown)


themarpe (changes requested):
Thanks - 2 things
 - Use different naming than prepended with `_` 
 - & would make sense to add the flags for this to CI, as we don't work with such high warnings, and codebase will "derail" again. 

Though I see these as style more so than "issues"

diablodale:
I will make other names that with `_`.

I don't understand what you want me to do with "& would make sense to add the flags for this to CI". 🤷‍♀️

There are bugs in depthai-core like all code. Some of those bugs the compiler has identified but your team doesn't get the reports because more warnings (like /W4 with MSVC) are not enabled. When /W4 is enabled, there is a lot lot lot. Some are true bugs like I reported at #766. Others are style. But there are so many...it is difficult to separate bugs from style. Therefore, reducing style issues allows for bugs to be more easily seen. That is my intention.

themarpe:
> I will make other names that with _.

Thanks

> Therefore, reducing style issues allows for bugs to be more easily seen. That is my intention

Got it - lets keep it this way for now, and see how for we get to a full /W4 clean build

diablodale:
forcepushed

* removed the `_` names
* slight refactor of XLinkConnection to fix clang-tidy warn of "use initializers" and "initializer should not duplicate what is already the default initializer in the class header".

> Got it - lets keep it this way for now, and see how for we get to a full /W4 clean build

Fixing issues in depthai headers tends to be my personal priority because those issues surface and break my build when I compile my own apps with `/W4`. This PR is progress...a step along the path towards a clean /W4 build.

I don't recommend /W4 for release/retail builds. In contrast, I do recommend that all new code should pass with no /W4 warns. New code is checked before adding, developer skill/style/technique improves, and the codebase continues on a path of less and less warns. Enforcing this on push is something your github+CI setup would need to enforce on push and/or pr merge. That's a topic/process that is completely within your team's purview. I can only recommend.

themarpe (approved):

Issue #807, opened 2023-04-12
RPC serialization and its mutex locks and/or shared_ptr use are unclear and likely crash


diablodale:
The RPC mechanism has unclear use of locks to protect it and shared_ptr's to hold its resources. The current code in v2.21.2 is not consistent enough for me to infer the intent. Therefore, I ask for clarity in intention by showing the code that concerns me.

I can definitely miss something in my code review. Please do point out any mistakes I made.

### Setup

* all compilers, os, platforms
* depthai-core and several versions earlier too

#### `pimpl->rpcStream`

I see no code that uses `pimpl->rpcStream`. Why does it exist?

https://github.com/luxonis/depthai-core/blob/125feb8c2e16ee4bf71b7873a7b990f1c5f17b18/src/device/DeviceBase.cpp#L625-L629

* line 626 creates a `shared_ptr<XLinkStream>` 
* line 626 then assigns that shared_ptr to `pimpl->rpcStream` which no code will meaningfully use.
* line 627 creates a new shared_ptr `rpcStream` and copy assigns `pimpl->rpcStream` to it. So now the XLinkStream has two (2) shared_ptrs pointing to it. Why do we need two?
* line 629 lambda COPY captures `rpcStream`. So now there are three (3) shared_ptrs pointing to the XLinkStream.
* `auto rpcStream`, the one outside the lambda, goes out of scope at the end of `init2()`. It had no purpose to exist.
* `pimpl->rpcStream` is eventually set to `nullptr` in `Device::close()`. Nothing ever used it. It had no purpose to exist.

In addition to no purpose, having the extra 2 refs on the shared_ptr adds minor unneeded code and cpu use.
I suggest...

* remove `pimpl->rpcStream`. I see no need for it.
* merge line 626 and 627 to a single `auto rpcStream = ...`
* change line 629 to be a move capture like `rpcStream = std::move(rpcStream)`
* Could put the make_shared in the capture to eliminate the local if OK with a long/wrap capture. It will emplace construct rather than move.

#### `pimpl->rpcMutex`

`pimpl->rpcMutex` seems to be used to protect RPC activity. It is used only one place...
https://github.com/luxonis/depthai-core/blob/125feb8c2e16ee4bf71b7873a7b990f1c5f17b18/src/device/DeviceBase.cpp#L630-L632

Yet, the resources on which the RPC mechanism depends, like `rpcStream` and `rpcClient` are not protected by this mutex. Of most concern is the code in `Device::close()`...

[...]

Issue #805, opened 2023-04-12
[REGRESSION] Device APIs crash application when called after `Device::close()`


diablodale:
Commit https://github.com/luxonis/depthai-core/commit/4b3c687dc3058bdc3d55166276ee811d4076f4c3 is a regression. It causes most Device APIs to crash applications with fatal errors like SIGSEGV when Device apis are called after `Device::close()`.

I suggest that commit be reverted.

### Setup

* all platforms, os, and compilers
* Depthai-core v2.21.1 and all later versions

### Repro

I found during code review. Then confirmed with the following test case.

```c++
TEST_CASE("Use Device APIs after close") {
    constexpr auto TEST_TIMEOUT = 20s;

    // Wait for 3s to acquire a device.
    auto t1 = steady_clock::now();
    vector<dai::DeviceInfo> availableDevices;
    do {
        availableDevices = dai::Device::getAllAvailableDevices();
        this_thread::sleep_for(500ms);
    } while(availableDevices.empty() && steady_clock::now() - t1 <= 3s);
    if(availableDevices.empty()) throw std::runtime_error("No devices found");

    // Create pipeline
    dai::Pipeline pipeline;

    // Define color source and output
    auto camColor = pipeline.create<dai::node::ColorCamera>();
    auto xoutColor = pipeline.create<dai::node::XLinkOut>();
    xoutColor->setStreamName("color");
    camColor->video.link(xoutColor->input);

    // Color properties
    camColor->setBoardSocket(dai::CameraBoardSocket::RGB);
    camColor->setResolution(dai::ColorCameraProperties::SensorResolution::THE_1080_P);
    camColor->setFps(30.0f);

    // Create device
    auto device = make_shared<dai::Device>(pipeline, availableDevices[0], dai::UsbSpeed::SUPER);
    cout << "MXID: " << device->getMxId() << "\nConnected cameras: ";
    for(const auto& cam : device->getConnectedCameras()) {
        cout << cam << " ";
    }
    cout << endl;

    // Get color queue and wait for one frame
    bool ignore{};
    if(device->getOutputQueue("color")->get<dai::ImgFrame>(TEST_TIMEOUT, ignore)) {
        cout << "Device " << device->getMxId() << " message arrived\n";
    }

    // Close device
    device->close();

    // Validate Device API behaviors
    CHECK_THROWS(std::ignore = device->getMxId());
    CHECK_THROWS(device->setXLinkChunkSize(1024));
[...]

PR #809, opened 2023-04-13
add Device tests for close and threading crashes

diablodale:
~Not ready to merge as it is Windows-only as I refine new tests.~

* Adds two unit tests to expose crashing bugs related to luxonis/depthai-core#805 and luxonis/depthai-core#807.
* Updates the cpp file to use Catch2
* Minor changes to existing 3rd tests to remove unneeded code and handle timeouts/missing hardware more cleanly

Test `Multiple devices created and destroyed in parallel` will usualy crash when 3 USB devices are attached. If it doesn't run the test again until it crashes.

Test `Device APIs after Device::close()` will always crash, SEG fault, etc.

### DIFF
--- a/src/xlink/XLinkConnection.cpp
+++ b/src/xlink/XLinkConnection.cpp
@@ -304,7 +304,7 @@ void XLinkConnection::close() {
     if(closed.exchange(true)) return;
 
     constexpr auto RESET_TIMEOUT = 2s;
-    constexpr auto BOOTUP_SEARCH = 5s;
+    constexpr auto BOOTUP_SEARCH = 5s;  // majority of usb reboots on Windows need more than 5 seconds; this leads to a cascade failure
 
     if(deviceLinkId != -1 && rebootOnDestruction) {
         auto previousLinkId = deviceLinkId;
@@ -428,7 +428,8 @@ void XLinkConnection::initDevice(const DeviceInfo& deviceToInit, XLinkDeviceStat
 
         const bool bootStatus = bootWithPath ? bootAvailableDevice(foundDeviceDesc, pathToMvcmd) : bootAvailableDevice(foundDeviceDesc, mvcmd);
         if(!bootStatus) {
-            throw std::runtime_error("Failed to boot device!");
+            // this throw will cause an app to crash if not caught by a try/catch around the app's `Device` constructor
+            throw std::runtime_error("Failed to begin the boot of a device!");
         }
     }
 
@@ -444,7 +445,7 @@ void XLinkConnection::initDevice(const DeviceInfo& deviceToInit, XLinkDeviceStat
         // Use "name" as hint only, but might still change
         bootedDescInfo.nameHintOnly = true;
 
-        logger::debug("Searching for booted device: {}, name used as hint only", bootedDeviceInfo.toString());
+        logger::debug("Searching for booted device: {}, name/path used as hint only", bootedDeviceInfo.toString());
 
         // Find booted device
         deviceDesc_t foundDeviceDesc = {};
--- a/tests/src/multiple_devices_test.cpp
+++ b/tests/src/multiple_devices_test.cpp
@@ -103,7 +103,7 @@ TEST_CASE("Multiple devices with 50 messages each") {
     REQUIRE(finished);
 }
 
-TEST_CASE("Multiple devices created and destroyed in parallel") {
+TEST_CASE("Multiple devices created and destroyed in parallel", "[.][multi_test_devices]") {
     constexpr auto TEST_TIMEOUT = 30s;
     constexpr auto COLOR_FPS = 30;
     constexpr auto MONO_FPS = 30;


themarpe (commented):
Thanks!

Just wondering on rpcClient - when is that now destructed - such that streams and other resources are destructed also (and that we bring still mitigate the initial issue that kept the device in an odd state).

Otherwise :+1: on this, targeting to bring in for next release

themarpe:
> I can not (yet) find any sideaffect that rpcClient would cause to keep "the device in an odd state". @szabi-luxonis did the work on https://github.com/luxonis/depthai-core/commit/4b3c687dc3058bdc3d55166276ee811d4076f4c3. I would like to understand their evidence or thinking on that. My suspicion is that it was other work (like the device firmware) that made a difference -or- there is a bug in the XLinkConnection::close() codepath causing an issue. I don't think rpcClient itself is the problem.

I think at the end of the day it boiled down to whether that rpcStream was destructed, in turn the underlying XLinkStream was closed.

Though I cannot confirm this to be the case. Would have to be retested prehaps with original post change vs just closing the XLinkStream (instead of "nulling out the rpcClient). In that way the rpcClient could remain valid and issues with interfacing with post close would not be required.

> For example, in XLinkConnection::close() is a code comment you wrote // TODO(themarpe) - revisit for TCPIP protocol. That seems relevant to PoE devices. thinking What is that about?

This waits till USB device is back online or times out - it helps with doing device down/up events as otherwise issues could occur (afaik while device was still (on usb level) destructing, it'd reconnect). The TODO is there as this does not happen for PoE devices (as the "bootDevice" is never true in that case)

> Which the atomic approach supports as it is non-blocking. This may not be the specific cause of this specific PoE problem. Yet, I think it is problematic to use a mutex and maybe it contributes to these types of race/timing fails. 

I agree that /w atomic this would make these simpler - I am up for changing that back, but IMO the crux here still lies somewhere around the XLinkStream being closed or not.

diablodale:
I'm not arguing 🙂 just trying to understand the thinking/logic as it isn't aligning with the code I see.
Why does it matter if the underlying XLinkStream of rpcClient was closed?

Anytime `Device::close()` or `Device::~Device` is called then the first meanful line of code in `DeviceBase::closeImpl()` is `XLinkConnection::close()`. It blocks doing work like `XLinkResetRemoteTimeout()`. The monitor thread with failing pings also calls `XLinkConnection::close()`. Both of their calls to `XLinkConnection::close()` call `XLinkResetRemoteTimeout()`.

`XLinkResetRemoteTimeout()` causes hardware "down" aka `getXLinkState(link) != XLINK_UP`. All `XLink*` apis calls check to see if the stream or link given as a param is XLINK_UP and return XLINK errors when false. And these XLINK errors are caught in depthai-core code as return values to those `XLink*` apis. This keeps the host in a well-defined state.

The call to `XLinkResetRemoteTimeout()` also sends a `XLINK_RESET_REQ` message to the remote OAK hardware for it to down itself and may force it with `DispatcherDeviceFdDown()`. Finally, it `XLink_sem_wait(&link->dispatcherClosedSem)` waits for the dispatcher to be closed. None of that codepath needs any `XLinkStream` to be closed.

That's my reasoning of the code...which I definitely can be wrong 😅. I can't align my reasoning with why you want the underlying XLinkStream of rpcClient to be closed...and why you have a belief it puts "the device in an odd state".

I don't have visibility into the OAK device firmware. The firmware must support any disconnect type at any time regardless of the state of XLinkStream's. Meaning the firmware should support: tcpip/ether failures, usb cable failures, sudden `DispatcherDeviceFdDown()`, sudden `XLINK_RESET_REQ`, etc. All of them should be handled cleanly and independently by the firmware itself. There is nothing for the depthai-core host code to do nicely (e.g. always close rpcClient's XLinkStream). Why? Because there are an unbounded number of disconnect scenarios and the OAK firmware needs to handle them by itself in isolation of any host.

diablodale:
I've isolated an unexpected grouping of XLink failures with rapid restarts. When this PR's test "Multiple devices created and destroyed in parallel" is run with 2 OAK-D-Lite and 1 OAK-D attached, I get failures. Might have to run the test more than once, but it should eventually fail. On my laptop, it fails almost every run.

Fails in `DeviceBase::startPipelineImpl` when it makes rpc calls. Error is `dai::XLinkWriteError` and it occurs on `pimpl->rpcClient->call("printAssets")` or `pimpl->rpcClient->call("startPipeline")`. Those two call locations are additionally curious to me because:

1. when both of those calls occur, there have been previous rpcClient calls that didn't throw. That suggests to me that the underlying XLink dispatcher for this connection has transitioned from a good state to a bad state that can't write -or- there is something in common with the two calls and unique from the earlier calls.
2. Both calls have no rpc parameters. Both calls have no `.as<>` for a return value. That makes these two unique in the whole function as earlier calls **do** have params or return values. This might be a false lead, but so far...no other rpcs are failing.

diablodale:
### Probably found bug

95% sure I found a somewhat relevant bug here
https://github.com/luxonis/depthai-core/blob/125feb8c2e16ee4bf71b7873a7b990f1c5f17b18/src/xlink/XLinkConnection.cpp#L342

`!found` should not be there. It errantly causes early exit of the while loop when anything in any state is found.

### OAK devices are not closing cleanly

For the below test run...
1. I changed the if test in the above loop to only exit on `X_LINK_BOOTLOADER`
2. added more logging throughout `XLinkConnection::close()`
3. set `XLINK_LEVEL=fatal`

I am assuming that `X_LINK_BOOTLOADER` is the "at rest" state of a USB device like OAK-D and OAK-D-Lite after it is plugged into a computer's USB port and letting the OAK device power on and load its internally-flashed bootloader. It appears in the OS as `Movidius MyriadX` and `USB\VID_03E7&PID_2485&REV_0001`

```
Found device count: 3
MXID: 1844301031490A1300
MXID: 18443010C17F0C1300
[14442C10C1A1C6D200] [3.1] [1.729] [system] [critical] Error executing RPC call: [json.exception.parse_error.110] parse error at byte 2: syntax error while parsing MessagePack value: expected end of input; last byte: 0x00
[2023-04-26 12:31:26.317] [warning] RPC mxid(14442C10C1A1C6D200) error: Couldn't read data from stream: '__rpc_main' (X_LINK_ERROR)
[2023-04-26 12:31:26.320] [warning] XLinkResetRemoteTimeout(1) mxid(14442C10C1A1C6D200) returned: X_LINK_ERROR
[2023-04-26 12:31:36.181] [warning] Reboot after XLinkResetRemoteTimeout(1) mxid(18443010C17F0C1300) failed to find device in 10s
[2023-04-26 12:31:36.208] [warning] Reboot after XLinkResetRemoteTimeout(0) mxid(1844301031490A1300) failed to find device in 10s
[2023-04-26 12:31:36.356] [warning] Reboot after XLinkResetRemoteTimeout(1) mxid(14442C10C1A1C6D200) failed to find device in 10s

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
multiple_devices_test.exe is a Catch2 v3.2.1 host application.
Run with -? for options

-------------------------------------------------------------------------------
Multiple devices created and destroyed in parallel
-------------------------------------------------------------------------------
..\tests\src\multiple_devices_test.cpp(106)
...............................................................................

..\tests\src\multiple_devices_test.cpp(173): FAILED:
explicitly with message:
  Fail construct Device() with DeviceInfo(name=3.1, mxid=14442C10C1A1C6D200,
  X_LINK_UNBOOTED, X_LINK_USB_VSC, X_LINK_MYRIAD_X, X_LINK_SUCCESS) Device
  already closed or disconnected: io error

MXID: 18443010C17F0C1300
MXID: 1844301031490A1300
MXID: 14442C10C1A1C6D200
[2023-04-26 12:31:48.160] [warning] Reboot after XLinkResetRemoteTimeout(2) mxid(1844301031490A1300) failed to find device in 10s
[2023-04-26 12:31:48.352] [warning] Reboot after XLinkResetRemoteTimeout(3) mxid(14442C10C1A1C6D200) failed to find device in 10s
[2023-04-26 12:31:48.434] [warning] Reboot after XLinkResetRemoteTimeout(1) mxid(18443010C17F0C1300) failed to find device in 10s
MXID: 1844301031490A1300
MXID: 18443010C17F0C1300
MXID: 14442C10C1A1C6D200
[2023-04-26 12:32:00.295] [warning] Reboot after XLinkResetRemoteTimeout(3) mxid(1844301031490A1300) failed to find device in 10s
[2023-04-26 12:32:00.314] [warning] Reboot after XLinkResetRemoteTimeout(4) mxid(18443010C17F0C1300) failed to find device in 10s
[2023-04-26 12:32:00.788] [warning] Reboot after XLinkResetRemoteTimeout(4) mxid(14442C10C1A1C6D200) failed to find device in 10s
===============================================================================
test cases:  1 |  0 passed | 1 failed
assertions: 11 | 10 passed | 1 failed
```

In that single run, I see many unexpected things
* 9 times the devices were rebooted yet couldn't be found in 10 seconds. Why are these devices taking so long?
* 1 time the OAK hardware failed to execute an RPC call. I can't find any host code that has the text `Error executing RPC call` so I think the OAK devices received a message on the RPC named stream, tried to unpack it and it was invalid. This should not happen; it suggests to me a bug somewhere between nanorpc on the host and nanorpc on the device.

I have also run the test with `XLINK_LEVEL=debug` and see often libusb errors about permission problems. Which on my Windows setup is unexpected as I remember permissions being related to UDEV
```
...
D: [global] [         0] [ThreadN] getLibusbDeviceMxId:326      libusb_open: Access denied (insufficient permissions) 
D: [global] [         0] [ThreadN] getUSBDevices:171    getLibusbDeviceMxId returned: Access denied (insufficient permissions)

...
```

themarpe:
Thanks a bunch for all the exploration @diablodale 

> That's my reasoning of the code...which I definitely can be wrong sweat_smile. I can't align my reasoning with why you want the underlying XLinkStream of rpcClient to be closed...and why you have a belief it puts "the device in an odd state".

This seems to be what resolved the issue with the initial commit - though it might not be the actual culprit. More testing required.

> XLink dispatcher for this connection has transitioned from a good state to a bad state that can't write

Most likely the case - the RPCs themselfs likely aren't the culprit.

> The rpc to printAssets...can we remove that? Or make it debug only?

Yes - we we can remove this outright.

> 95% sure I found a somewhat relevant bug here
> `!found` should not be there. It errantly causes early exit of the while loop when anything in any state is found.

Good catch - Yes, that would in fact exit out too soon. Shall we add it here: https://github.com/luxonis/depthai-core/pull/809/files#diff-17c7352c03ae1d3bb07f3e7a3c1b6f98312f549a440e186f450e3a8f563e3121L338-L340 with `else found = false`?

> I am assuming that X_LINK_BOOTLOADER is the "at rest" state of a USB device like OAK-D and OAK-D-Lite after it is plugged into a computer's USB port and letting the OAK device power on and load its internally-flashed bootloader. It appears in the OS as Movidius MyriadX and USB\VID_03E7&PID_2485&REV_0001

That would be X_LINK_UNBOOTED state. X_LINK_BOOTLOADER would be for cases where device is presented as Luxonis Bootloader (the former is ROM BL/1st stage, the latter is NOR flashed custom BL (2nd stage))

> I have also run the test with XLINK_LEVEL=debug and see often libusb errors about permission problems. Which on my Windows setup is unexpected as I remember permissions being related to UDEV

I think this relates to device being in a state that comms aren't possible. Though not 100% sure what is the underlying cause for this error message from libusb on Windows

diablodale:
Unfortunately, I don't think I can make progress on the connection issue(s) because I don't have visibiity into the OAK device and its firmware. There is not enough public knowledge on its state changes, internal logic, etc.

* All my debugging to date indicates to me hardware problems in state transitions. Either libusb code failing to transition states timely, or the OAK device failing to transition states promptly. This is backed by my repeated test failure evidence and logs showing that even after 15 seconds OAK devices can not be seen by libusb-xlink code (meaning libusb-xlink, WIndows, or OAK have failed to work as expected). If it doesn't work on USB, it is probably even worse with PoE. I recomemnd focus first on USB and get it working. My expanded test case in this PR is an easy repro to expose the bug.
* To isolate, I attempted to use XLink from depthai-core v2.15.5 (the last xlink with winusb). I failed because the XLink interface had large changes with winusb->libusb and broke both the xlink API and its behavior. Getting current depthai-core to work with the winusb-XLink was more than a half-day of changes (I stopped then).
* Do you have any docs and flow charts that document the state changes for OAK devices? I can find none in code or in the online depthai docs. At a mimum it would have circles for each `XLinkDeviceState_t` and arrows showing all possible transitions and how they occur. For example, all circles except the first would have an arrow for `kill -9 on app` that goes back to the beginning. I suspect some of those arrows are not coded correctly in today's depthai+xlink+firmware. Such a state diagram would be used for all XLink coding and test case creation.

If you want, I can cleanup, or close an open a new PR, to include the one bug I found and all the cleanup work. It doesn't fix the connection problems -- instead forward movement to have cleaner build, enables more analyzers/warnings, etc.

themarpe:
@diablodale 
The following diagram is how the boot process flows. The ROM BL decides (based on HW + NOR contents) on whether or not it boots BL (if its present, some devices don't have NOR, and can only ever boot directly from UNBOOTED->BOOTED).

Each of transitions is a USB down->up event as reenumeration is done.

One extra case is if you call DeviceBootloader, then it boots into X_LINK_BOOTLOADER state if it was UNBOOTED before. Otherwise it just connects to running Bootloader

(and neat thing, this mermaid flowchart:) )

```mermaid
flowchart TD

DEVICE_TYPE{USB device}

X_LINK_UNBOOTED[[X_LINK_UNBOOTED<br/>'Movidius MyriadX' 0x03E7, 0x2485]]
X_LINK_BOOTLOADER[[X_LINK_BOOTLOADER<br/>'Luxonis Bootloader' 0x03E7, 0xf63c]]
X_LINK_BOOTED[[X_LINK_BOOTED<br/>'Luxonis Device' 0x03E7, 0xf63b]]
X_LINK_FLASH_BOOTED[[X_LINK_FLASH_BOOTED<br/>0x03E7, 0xf63d]]

HOST_XLINK_BOOT["XLinkBootMemory(firmwareBin)"]
HOST_RPC_READ["rpc(readPipelineCodeFromXLink)"]
HOST_RPC_PIPELINE["rpc(build) + rpc(startPipeline)"]

HOST_BOOTLOADER_BOOT["DeviceBootloader::bootMemory(firmwareBin)"]

STANDALONE_READ["device reads its pipeline from flash"]
STANDALONE_PIPELINE["device builds+starts its pipeline"]


DEVICE_TYPE -->|no BL or BOOT btn is pressed| X_LINK_UNBOOTED
DEVICE_TYPE -->|has BL| X_LINK_BOOTLOADER

X_LINK_UNBOOTED --> HOST_XLINK_BOOT

X_LINK_BOOTLOADER -->|standalone flashed| X_LINK_FLASH_BOOTED
X_LINK_FLASH_BOOTED -.-> STANDALONE_READ
STANDALONE_READ -.-> STANDALONE_PIPELINE
X_LINK_BOOTLOADER -->|normal| HOST_BOOTLOADER_BOOT
HOST_BOOTLOADER_BOOT --> X_LINK_BOOTED

HOST_XLINK_BOOT --> X_LINK_BOOTED

X_LINK_BOOTED --> HOST_RPC_READ
HOST_RPC_READ --> HOST_RPC_PIPELINE

subgraph "host app code: Device(pipeline) [after boot part]"
    X_LINK_BOOTED
    HOST_RPC_READ
    HOST_RPC_PIPELINE
end

subgraph "Automatic standalone flash behavior"
    X_LINK_FLASH_BOOTED
    STANDALONE_READ
    STANDALONE_PIPELINE
end

```

themarpe:
> If you want, I can cleanup, or close an open a new PR, to include the one bug I found and all the cleanup work. It doesn't fix the connection problems -- instead forward movement to have cleaner build, enables more analyzers/warnings, etc.

You may include it here

src/xlink/XLinkConnection.cpp
```diff
@@ -303,24 +292,21 @@ XLinkConnection::XLinkConnection(const DeviceInfo& deviceDesc, XLinkDeviceState_
...
-    constexpr auto RESET_TIMEOUT = seconds(2);
-    constexpr auto BOOTUP_SEARCH = seconds(5);
+    constexpr auto RESET_TIMEOUT = 2s;
+    constexpr auto BOOTUP_SEARCH = 5s;
 
     if(deviceLinkId != -1 && rebootOnDestruction) {
         auto previousLinkId = deviceLinkId;
 
-        auto ret = XLinkResetRemoteTimeout(deviceLinkId, duration_cast<milliseconds>(RESET_TIMEOUT).count());
+        const auto ret = XLinkResetRemoteTimeout(deviceLinkId, static_cast<int>(std::chrono::duration_cast<std::chrono::milliseconds>(RESET_TIMEOUT).count()));
```
themarpe, line 307:
```suggestion
        const auto ret = XLinkResetRemoteTimeout(deviceLinkId, static_cast<int>(duration_cast<milliseconds>(RESET_TIMEOUT).count()));
```

diablodale: That won't compile.

* `duration_cast` is only used once in the whole file, and has no `using` (nor woud it benefit from `using` since that would *add* unneeded lines of code).
* `std::chrono::milliseconds` is used in several places in the file, all other places use the full `std::chrono::milliseconds`, and no `using` which would enable the short name. This is in alignment with the coding approach in this file.

Alternative: I could change all locations with `std::chrono::milliseconds` -> `milliseconds` and then put `using std::chrono::milliseconds` at the top. I hesitate doing things like that on my own to prettify code unless I get compilers warnings or the existing code is unreadable/can't understand...or maybe I've had too much caffine. ;-)

diablodale:
I don't see any "code" faults. Woohoo 😃

The CI thinks there are failures because:
1. the extra debug has the word "warning" in it which is then detected by the ctest regex and fails the test
2. the multiple_devices_test case 2 REQUIRES 2+ hardware devices to be attached. And it fails otherwise. The only machine that ran in CI was a windows machine which appears to have only one device attached.

I need to know what are next steps / answers to questions.
1. What do you want to do about only 1 OAK on the test machine? Should that case be marked/filtered using flags? Or what do you want?
2. Is your team good with this PR's state so that I can remove the extra debug and push?

themarpe:
> What do you want to do about only 1 OAK on the test machine? Should that case be marked/filtered using flags? Or what do you want?
IMO best if we tag it (like PoE/USB) and skip that tag by default for now (as we don't have HIL In this repo that handles this scenario ATM - we'll add it afterwards)

> Is your team good with this PR's state so that I can remove the extra debug and push?
LGMT, thanks!

diablodale:
Force-pushed. I removed the last commit that had the extra debug in it -- replaced it with a small commit that clarified some comments and debug log messages.

### Major concern

The majority 50+% of OAK usb devices do not reboot fast enough with the current depthai-core codebase on Windows.
PoE is unknown. Other OS are unknown.

https://github.com/luxonis/depthai-core/blob/5f75ae74ade6ae12bd56c71c0149c0482f23c18f/src/xlink/XLinkConnection.cpp#L316

This PR's new test case with multiple devices --- more than half the USB device reboots do not complete within the BOOTUP_SEARCH 5 seconds. This easily cascades to failed connections when depthai-core attempts to boot and upload firmware to the device. When depthai-core attempts this after-reboot activity, some part of the USB device accepts a connection (myraid firmware?) but that connection seems to be brain-dead or not functioning correctly. This series of states causes various levels of depthai-core errors...some may cause exceptions to be thrown. The app will crash if these exceptions are not caught with a try/catch around the Device() constructor.

I extended BOOTUP_SEARCH to 10 seconds. Some reboots still did not complete within 10 seconds. I estimate 25-33% of them failed. Eventually, all 3 of my USB devices reboot, yet the time needed to reboot can be many seconds beyond 10.

I recommend BOOTUP_SEARCH be extended or new reboot logic for depthai-core to accommodate the time needed to reboot. While new firmware/bootloader may help, it doesn't help all those devices which are not burned with the new firmware.

It is not useful for depthai-core to start uploading firmware to a device when it hasn't completely rebooted. The existing codebase does this. The code might have been written to accommodate the scenario of a device being unplugged after/during a reboot....and not wanting that user to wait more than 5 seconds. However, this choice leads to the majority of USB reboots failing. That doesn't seem a wise choice given the failure scenarios I describe above.

Separately, is the concern of a 10+ second reboot for a usb device. This is a topic to be explored by the Luxonis team who has access to the firmware and multiple host OS+hardware.

Issue #469, opened 2022-04-30
startPipeline deadlocks when two xlinkouts have the same stream name


jdavidberger:
After debugging for a while, it turns out that startPipeline will deadlock when you have two xlink nodes which share a streamname. While this configuration wouldn't make sense to run; it would be beneficial if this was validated and throws an exception instead of deadlocking.

themarpe:
I agree, we'll add a check for this case

PR #811, opened 2023-04-14
prevent duplicate xlink stream names

diablodale:
- fixes luxonis/depthai-core#469
- add test case

passes all tests


themarpe (approved):
Thanks!

PR #825, opened 2023-05-07
UVC support

alex-luxonis:
Currently, a single UVC (Universal Video Class, standard camera/webcam) node can be created. 

When the Device() is created , a DeviceConfig/BoardConfig is required to be sent, that will enable the UVC interface in the USB descriptor. Several UVC camera configuration must be set in advance. Available UVC configuration:
- width and height (1920 x 1080 default) of the frames expected to be received, as well as the frame type/format: 
  - `NV12` (e.g. ColorCamera `video`) - default
  - `I420`/`YUV420p` (e.g. ColorCamera `isp`)
  - `RAW8`/`GRAY8` (MonoCamera)
  - `RAW16`
  - `BITSTREAM` - H.264 from VideoEncoder (configured accordingly) - was just briefly tested on Linux
- camera name string (max 64 characters), defaults to `"Luxonis UVC Camera"`

Also this adds the possibility to configure USB device `productName` and `manufacturer` strings (max 64 characters each).

A python example is available. TODO: add a C++ example.

Known limitation on Windows OS: the stream could not be opened when running at USB2 speeds (`HIGH`). But works fine at USB3 speeds.

Related PR: https://github.com/luxonis/depthai-shared/pull/140


chengguizi:
@Erol444  Can I quickly check that based on the description here, currently only a single UVC node is allowed? Meaning that it is not possible to stream multiple camera streams over different UVC channels, correct? Thanks!

alex-luxonis:
Yes, that's a current limitation, due to difficulties on dynamically modifying USB descriptors in FW.

It is possible though to link multiple nodes to a single UVC node, but the host will have to de-multiplex the frames (and may need to encode some info in each frame to differentiate between cameras, for example with Script nodes).

For example with https://github.com/luxonis/depthai-python/blob/main/examples/MonoCamera/mono_preview.py
```py
uvc = pipeline.create(dai.node.UVC)
monoLeft.out.link(uvc.input)
monoRight.out.link(uvc.input)
```

Or even creating multiple UVC nodes, but will have the same behavior as above, with an additional log printed:
```
uvcLeft = pipeline.create(dai.node.UVC)
uvcRight = pipeline.create(dai.node.UVC)
monoLeft.out.link(uvcLeft.input)
monoRight.out.link(uvcRight.input)
```
`[UVC(4)] [warning] More than one UVC instance is not properly supported - all packets will be sent to the same USB UVC interface`

Common code needed for both cases:
```py
config = dai.Device.Config()
config.board.uvc = dai.BoardConfig.UVC(*monoLeft.getResolutionSize())
config.board.uvc.frameType = dai.ImgFrame.Type.RAW8
pipeline.setBoardConfig(config.board)
```

PR #865, opened 2023-07-04
Changes needed for a clean exit on the FW side

moratom:
Also added in thread naming on Linux for easier profiling.

### DIFF
--- a/src/pipeline/ThreadedNode.cpp
+++ b/src/pipeline/ThreadedNode.cpp
@@ -1,5 +1,11 @@
 #include "depthai/pipeline/ThreadedNode.hpp"
 
+#include <spdlog/spdlog.h>
+
+#ifdef __linux__
+    #include <pthread.h>
+#endif
+
 namespace dai {
 
 void ThreadedNode::start() {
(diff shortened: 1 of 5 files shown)


src/pipeline/ThreadedNode.cpp
```diff
@@ -8,12 +14,19 @@ void ThreadedNode::start() {
...
+            running = false;
+        } catch(const std::runtime_error& ex){
             logger->error("Node threw exception, stopping the node. Exception message: {}", ex.what());
             running = false;
         }
     });
+#ifdef __linux__
+    auto handle = thread.native_handle();
+    pthread_setname_np(handle, fmt::format("{}-{}", getName(), id).c_str());
+#endif
```
themarpe, line 29: Lets move this to platform specific utils
And naming format is `{}({})` -> name, id

src/pipeline/ThreadedNode.cpp
```diff
@@ -1,5 +1,11 @@
 #include "depthai/pipeline/ThreadedNode.hpp"
 
+#include <spdlog/spdlog.h>
```
themarpe, line 3: Needed?

src/pipeline/ThreadedNode.cpp
```diff
@@ -1,5 +1,11 @@
...
+
+#ifdef __linux__
+    #include <pthread.h>
+#endif
```
themarpe, line 7: Move to platform specific utils as well

themarpe (changes requested):

src/pipeline/ThreadedNode.cpp
moratom: No, removed.

PR #878, opened 2023-07-18
Support for RVC3 core dump

moratom:


src/device/DeviceGate.cpp
```diff
@@ -51,7 +57,9 @@ DeviceGate::DeviceGate(const DeviceInfo& deviceInfo) : deviceInfo(deviceInfo) {
 
     // Discover and connect
     pimpl->cli = std::make_unique<httplib::Client>(deviceInfo.name, DEFAULT_PORT);
+    pimpl->cli->set_read_timeout(60);  // 60 seconds timeout to allow for compressing the core dumps without async
     // pimpl->cli->set_connection_timeout(2);
+    stateMonitoringThread = std::thread(&DeviceGate::threadedStateMonitoring, this);
```
themarpe, line 62: We know when device conn falls down abruptly - perhaps that time would be better to ping the gate for this information? (destructor of DeviceBase)

src/utility/Platform.cpp
```diff
@@ -60,5 +64,17 @@ void setThreadName(JoiningThread& thread, const std::string& name) {
...
+#if defined(_WIN32) || defined(__USE_W32_SOCKETS)
+    char tmpPathBuffer[MAX_PATH];
+    GetTempPathA(MAX_PATH, tmpPathBuffer);
+    tmpPath = tmpPathBuffer;
+#else
+    tmpPath = "/tmp/";
+#endif
+    return tmpPath;
+}
+
```
themarpe, line 78: https://man7.org/linux/man-pages/man3/mkdtemp.3.html
To have a unique place to store this - perhaps some codebases have mkdtemp cross platfrom impl already (CPython has one, though its in Python)

themarpe (changes requested):

themarpe:
Main idea being that as currently things fail with XLinkException (generically), they'd instead fail with more specific exceptions as we continue adding side channels. This kind of "Crash" exception would also carry paths to dumps, etc... within

src/device/DeviceGate.cpp
moratom: I think this can be moved to the destructor of DeviceBase yeah. 

I would still do it in a "blocking" manner though - to wait if the FW crashes on the exit or not / if it was killed by the watchdog on the gate side.

Short term it will be a bit of the pain, but it will help/force us clean up the exits properly, unless this isn't worth pinning down and likely isn't the cause for many issues?
I don't have a good grasp of how robustly Linux cleans up after the abrupt FW exits.

moratom: @themarpe the way I imagined it is pushed now, let me know if this aligns with what you had in mind.

moratom: This is how the logs look like on a crash with latest gate & this branch. 
```
matevz@matevz-luxmmPC:~/Luxonis/RVC3/depthai-device-kb/external/depthai-python/examples/ColorCamera$ python rgb_video.py 
[2023-07-18 19:07:45.049] [warning] FW crashed - trying to get out the logs and the core dump
[2023-07-18 19:07:48.058] [warning] Log file found - trying to save it
[2023-07-18 19:07:48.058] [warning] Log file saved to /tmp/depthai_vbk0Px/depthai_gate_log_0.0.1+fdb7d1769b30fc549d59393352a2bef56fbd7f0b_2023_07_18_17_07.log - please report to developers
[2023-07-18 19:07:48.058] [warning] Getting the core dump out - this can take up to a minute, because it first needs to be compressed.
[2023-07-18 19:07:52.599] [warning] Core dump found - trying to save it
[2023-07-18 19:07:52.600] [warning] Core dump saved to /tmp/depthai_vbk0Px/depthai_core_dump_0.0.1+fdb7d1769b30fc549d59393352a2bef56fbd7f0b_2023_07_18_17_07.core.gz - please report to developers
```

src/device/DeviceBase.cpp
```diff
@@ -419,6 +419,10 @@ void DeviceBase::closeImpl() {
     // At the end stop the monitor thread
     if(monitorThread.joinable()) monitorThread.join();
 
+    // If the device was operated throgh gate, wait for the session to end
+    if(gate){
+        gate->waitForSessionEnd();
+    }
```
themarpe, line 425: Is this the right place to put it, and not after all the XLink related coms were closed (including below RPC, etc...)

themarpe (changes requested):
I'd merge the "crashdump" into one single file. Its main idea is not easy digestion, but easy sharing/reporting.

Likewise will backport to RVC2 and continue onwards to next revisions as well

PR #900, opened 2023-09-07
Timesync optimization

asahtik:
Changed timesync to only take into account the time between send and remote receive + remote send and receive as the RTT.
Moved the receive timestamp to XLink - get data to write back through the stream packet.

### DIFF
--- a/include/depthai/xlink/XLinkStream.hpp
+++ b/include/depthai/xlink/XLinkStream.hpp
@@ -54,8 +54,8 @@ class XLinkStream {
     void write(const void* data, std::size_t size);
     void write(const std::uint8_t* data, std::size_t size);
     void write(const std::vector<std::uint8_t>& data);
-    std::vector<std::uint8_t> read();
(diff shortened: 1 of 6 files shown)


include/depthai/xlink/XLinkStream.hpp
```diff
@@ -54,8 +54,8 @@ class XLinkStream {
     void write(const void* data, std::size_t size);
     void write(const std::uint8_t* data, std::size_t size);
     void write(const std::vector<std::uint8_t>& data);
-    std::vector<std::uint8_t> read();
-    void read(std::vector<std::uint8_t>& data);
+    std::vector<std::uint8_t> read(struct timespec* out_time=NULL);
+    void read(std::vector<std::uint8_t>& data, struct timespec* out_time=NULL);
```
themarpe, line 58: Overload with `struct timespec&` instead, and rename to `timestampReceived` (camelCase)

themarpe (changes requested):
Also bump the FW once ready

Issue #882, opened 2023-07-24
[BUG] {setWarpTransformFourPoints() unexpeced behaviour}


jakaskerl:
**Describe the bug**
`setWarpTransformFourPoints()`works as expected in some scenarios, but acts strange in some other scenarios. The problem is that it seems to work great with some point configuration, but moving one point of the polygon by a single pixel causes the warp to zoom-in too much. 
I tried to figure out what causes the problem, but to no avail. I assumed it was something related to the data-types and how the warp was programmed.

**Minimal Reproducible Example - not entirely minimal to ease debugging** 
```
# !/usr/bin/env python3

import depthai as dai
import cv2
import numpy as np

# Create pipeline
pipeline = dai.Pipeline()

"""Main nodes"""
# Camera
colorCameraNode = pipeline.create(dai.node.ColorCamera)
# Properties
colorCameraNode.setPreviewSize(600, 400)
colorCameraNode.setResolution(dai.ColorCameraProperties.SensorResolution.THE_1080_P)
colorCameraNode.setInterleaved(False)
colorCameraNode.setColorOrder(dai.ColorCameraProperties.ColorOrder.BGR)
# Img Transf
imgTransfNode = pipeline.create(dai.node.ImageManip)
# Properties
imgTransfNode.setMaxOutputFrameSize(2000 * 1500 * 3)

"""LinkIn/Out nodes"""
# Camera
colorCameraOutNode = pipeline.create(dai.node.XLinkOut)
colorCameraOutNode.setStreamName("colorCameraOutQueue")
# Config In
imgTransfConfigInNode = pipeline.create(dai.node.XLinkIn)
imgTransfConfigInNode.setStreamName("imgTransfConfigQueue")
# Tranf Img out
imgTransfOutNode = pipeline.create(dai.node.XLinkOut)
imgTransfOutNode.setStreamName("imgTransfOutQueue")

"""Connections"""
# Camera Node to Camera Out
colorCameraNode.preview.link(colorCameraOutNode.input)
# Camera Node to Img Transf Node
colorCameraNode.preview.link(imgTransfNode.inputImage)
# Config to Img Transf Node
imgTransfConfigInNode.out.link(imgTransfNode.inputConfig)
# Img Transf Node to Img Out
imgTransfNode.out.link(imgTransfOutNode.input)

# Connect to device
device = dai.Device(pipeline)

# Create queues
colorCameraOutQueue = device.getOutputQueue(
    name="colorCameraOutQueue",
[...]

PR #906, opened 2023-09-22
Fixed setWarpTransformFourPoints() unexpected behaviour #882

asahtik:
Using setWarpTransformFourPoints() left `keepAspectRatio` on `true` which caused unexpected behaviour in some cases.


themarpe (approved):
Thanks!

PR #917, opened 2023-11-02
Message groups

asahtik:
Implemented Sync and MessageDemux nodes with the MessageGroup datatype.


include/depthai/pipeline/node/VideoEncoder.hpp
```diff
@@ -29,6 +29,11 @@ class VideoEncoder : public NodeCRTP<Node, VideoEncoder, VideoEncoderProperties>
      */
     Output bitstream{*this, "bitstream", Output::Type::MSender, {{DatatypeEnum::ImgFrame, false}}};
 
+    /**
+     * Outputs EncodedFrame message that carries encoded (MJPEG, H264 or H265) frame data.
+     */
+    Output out{*this, "out", Output::Type::MSender, {{DatatypeEnum::EncodedFrame, false}}};
```
themarpe, line 35: Mutually exclusive between bitstream, correct? Do note it in comment if this is the case

include/depthai/pipeline/datatype/MessageGroup.hpp
```diff
@@ -0,0 +1,63 @@
...
+#include <unordered_map>
+#include <vector>
+
+#include "../../utility/TypeToEnum.hpp"
+#include "depthai-shared/datatype/RawMessageGroup.hpp"
+#include "depthai/pipeline/datatype/Buffer.hpp"
+namespace dai {
+
+/**
+ * IMUData message. Carries normalized detection results
```
themarpe, line 14: Update the comment

src/device/DataQueue.cpp
```diff
@@ -35,8 +38,23 @@ DataOutputQueue::DataOutputQueue(const std::shared_ptr<XLinkConnection> conn, co
             while(running) {
                 // Blocking -- parse packet and gather timing information
                 auto packet = stream.readMove();
+                DatatypeEnum type;
                 const auto t1Parse = std::chrono::steady_clock::now();
-                const auto data = StreamMessageParser::parseMessageToADatatype(&packet);
+                const auto data = StreamMessageParser::parseMessageToADatatype(&packet, type);
+                if(type == DatatypeEnum::MessageGroup) {
+                    auto rawMsgGrp = std::dynamic_pointer_cast<RawMessageGroup>(data->serialize());
```
themarpe, line 45: This seems unnecessary - static_cast (due to previous type check above) data itself and then access the underlying group variable directly (no need to "serialize" again)

src/device/DataQueue.cpp
```diff
@@ -35,8 +38,23 @@ DataOutputQueue::DataOutputQueue(const std::shared_ptr<XLinkConnection> conn, co
...
+                    unsigned int size = rawMsgGrp->group.size();
+                    std::vector<std::shared_ptr<ADatatype>> packets;
+                    packets.reserve(size);
+                    for(unsigned int i = 0; i < rawMsgGrp->group.size(); ++i) {
+                        auto dpacket = stream.readMove();
+                        packets.push_back(StreamMessageParser::parseMessageToADatatype(&dpacket));
+                    }
+                    for(auto& msg : rawMsgGrp->group) {
+                        logger::info("Message group index {}", msg.second.index);
+                        msg.second.buffer = packets[msg.second.index]->serialize();
```
themarpe, line 55: This seems like we are storing Raw* messages in MessageGroup? Group should preferably hold something along the lines of: `map<string, shared_ptr<ADatatype>>`

src/pipeline/datatype/MessageGroup.cpp
```diff
@@ -0,0 +1,97 @@
...
+            return std::make_shared<FeatureTrackerConfig>(std::dynamic_pointer_cast<RawFeatureTrackerConfig>(ptr));
+        case DatatypeEnum::ToFConfig:
+            return std::make_shared<ToFConfig>(std::dynamic_pointer_cast<RawToFConfig>(ptr));
+        case DatatypeEnum::TrackedFeatures:
+            return std::make_shared<TrackedFeatures>(std::dynamic_pointer_cast<RawTrackedFeatures>(ptr));
+        case DatatypeEnum::MessageGroup:
+            return {};
+    }
+    return {};
+}
```
themarpe, line 62: As per above comment - if we can make this not necessary, that'd be best 
(eg MessageGroup has no Raw* equivalent, but deals purely in core space - this will be inline with merging of core & shared as well)

include/depthai/pipeline/node/MessageDemux.hpp
```diff
@@ -0,0 +1,28 @@
...
+
+    /**
+     * Input message of type MessageGroup
+     */
+    Input input{*this, "input", Input::Type::SReceiver, {{DatatypeEnum::MessageGroup, false}}};
+
+    /**
+     * A map of outputs, where keys are same as in the input MessageGroup
+     */
+    OutputMap outputs;
```
themarpe, line 24: If MessageGroup can be "sync successful or not", perhaps might make sense having this also have a check here, to only demux on successful syncs, etc...

src/utility/H26xParsers.hpp
```diff
@@ -0,0 +1,14 @@
+#pragma once
+
+#include <cstdint>
+#include <vector>
+
+namespace dai {
+
```
themarpe, line 7:
```suggestion
namespace utility {
```

src/utility/H26xParsers.hpp
```diff
@@ -0,0 +1,14 @@
...
+enum class Profile { H264, H265 };
+enum class SliceType { P, B, I, SP, SI, Unknown };
+
+std::vector<SliceType> getTypesH264(const std::vector<std::uint8_t>& bs, bool breakOnFirst = false);
+std::vector<SliceType> getTypesH265(const std::vector<std::uint8_t>& bs, bool breakOnFirst = false);
+
```
themarpe, line 13:
```suggestion
}
```

src/utility/TypeToEnum.cpp
```diff
@@ -0,0 +1,86 @@
+#include "depthai/utility/TypeToEnum.hpp"
```
themarpe, line 1: Static field or using in message would be better. Note, conditional to structure of this as I've pointed out above, if its even necessary

themarpe (changes requested):

include/depthai/pipeline/node/Sync.hpp
```diff
@@ -0,0 +1,61 @@
...
+     * Set the number of attempts to get the specified max interval between messages in the group
+     * @param syncAttempts Number of attempts to get the specified max interval between messages in the group
+     */
+    void setSyncAttempts(unsigned int syncAttempts);
+
+    /**
+     * Specifies how many frames will be available in the pool
+     * @param numFrames How many frames will pool have
+     */
+    void setNumFramesPool(unsigned int numFramesPool);
```
moratom, line 42: @asahtik does the sync node need its own pools?

asahtik: It's for getting a MessageGroup object for storing the metadata (map, whether the sync was successful)

include/depthai/pipeline/datatype/MessageGroup.hpp
```diff
@@ -0,0 +1,65 @@
...
+    virtual ~MessageGroup() = default;
+
+    /// Group
+    std::shared_ptr<ADatatype> operator[](const std::string& name);
+    template <typename T>
+    std::shared_ptr<T> get(const std::string& name) {
+        return std::dynamic_pointer_cast<T>(group[name]);
+    }
+    void add(const std::string& name, const std::shared_ptr<ADatatype>& value);
+
```
moratom, line 34: Add a method `getMessageNames` or something along these lines, since the map itself is private.

moratom:
Please add an example to `depthai-core` & `depthai-python` (Create a `Sync` sub-directory in `examples`)

include/depthai/pipeline/datatype/MessageGroup.hpp
themarpe: IMO, rather expose more of the map like functionality altogether. But nice to have, non blocking, can be added later

include/depthai/pipeline/node/Sync.hpp
themarpe: I think we should default to dynamic here (and later on we can replace with pools as we fully unabstract them, as part of the pipeline)

themarpe (changes requested):
Add a nice example of syncing eg: left & right imagers, etc... then sending on host & displaying both synced.

examples/Sync/depth_video_synced.cpp
```diff
@@ -0,0 +1,62 @@
...
+    dai::Device device(pipeline);
+
+    auto queue = device.getOutputQueue("xout", 10, true);
+
+    while(true) {
+        auto msgGrp = queue->get<dai::MessageGroup>();
+        std::cout << "Got message group" << std::endl;
+        for(auto& frm : *msgGrp) {
+            auto imgFrm = std::dynamic_pointer_cast<dai::ImgFrame>(frm.second);
+            cv::imshow(frm.first, imgFrm->getFrame());
```
moratom, line 53: Same nitpick for depth as in `depthai-python`

src/pipeline/datatype/MessageGroup.cpp
```diff
@@ -0,0 +1,77 @@
...
+std::shared_ptr<RawBuffer> MessageGroup::serialize() const {
+    return raw;
+}
+
+MessageGroup::MessageGroup() : Buffer(std::make_shared<RawMessageGroup>()), rawGrp(*dynamic_cast<RawMessageGroup*>(raw.get())) {}
+MessageGroup::MessageGroup(std::shared_ptr<RawMessageGroup> ptr) : Buffer(std::move(ptr)), rawGrp(*dynamic_cast<RawMessageGroup*>(raw.get())) {}
+
+std::shared_ptr<ADatatype> MessageGroup::operator[](const std::string& name) {
+    return group[name];
+}
```
moratom, line 20: Could we return a reference here to match how the unordered queue handles the operator?

Otherwise I'd use the `.at` method and (re) throw if no message is found, I think what will happen right now is we will return default ADatatype message in case the key does not exist.

moratom (changes requested):
Apart from the requested changes, looks great!

include/depthai/pipeline/node/Sync.hpp
```diff
@@ -0,0 +1,61 @@
...
+
+    /**
+     * A map of inputs
+     */
+    InputMap inputs;
+
+    /**
+     * Output message of type MessageGroup
+     */
+    Output out{*this, "out", Output::Type::MSender, {{DatatypeEnum::MessageGroup, false}}};
```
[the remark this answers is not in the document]
szabi-luxonis: I guess should be renamed to `outputs` to match `inputs` format?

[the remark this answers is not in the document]
themarpe: `outputs` refers to a map usually, so I'd leave as `out` as it sends as single (albeit combined) message

themarpe (approved):

moratom (approved):

PR #920, opened 2023-11-08
Get connectivity rpc

asahtik:


include/depthai/device/DeviceBase.hpp
```diff
@@ -570,6 +571,13 @@ class DeviceBase {
      */
     std::vector<CameraBoardSocket> getConnectedCameras();
 
+    /**
+     * Get connectivity for device
+     *
+     * @returns Vector of connection type
+     */
+    std::vector<Connectivity> getConnectivity();
```
moratom, line 579: I think something like `getAvailableInterfaces()` would be better here. 
And `[Connection?]Interface` for the type.

moratom (approved):
Apart from the naming, LGTM!

include/depthai/device/DeviceBase.hpp
themarpe: Note Available usually denotes a subset of all existing. In this case, the function should return "all" device's interfaces that it has. (the "current / what is connectable by" is explorable through device discovery)

Lets just document it accordingly and LGTM

moratom: @themarpe just to clarify, do you mean leave it at `getConnectivity()` or changing it to `getAllInterfaces()` ?

themarpe: Perhaps `getConnectionInterfaces` and later on / if need be we can expose available interfaces

PR #924, opened 2023-11-10
IR Intensity API

zrezke:
## Added ability to control IR brightness based on a normalized intensity instead of specifying the current.

Previous implementation of setting ir brightness was based on specifying the current in mA. After 765mA, the duty cycle was linearly interpolated from 30% to 6%, resulting in 765mA @ 30% duty cycle being the peak brightness, altho you could go up to 1200mA (for dot projector).
The new API takes a float on range [0, 1] as the ir (dot or flood) brightness and computes the dot proj. and flood light current, aswell as the duty cycle.

### Duty cycle and current calculations:
- Duty cycle = 30% when exposure time is longer than 30% of frame time.
  - In this case the dot current is computed as dotIntensity * 765 [mA]
- Duty cycle is 100% of exposure time, when exposure time is shorter than 30% of frame time.
  - In this case the dot current is dotIntensity * 765 [mA] + linear upto 1200mA in respect to the difference in (30%frame time - exposure time), to accommodate for the shorter duty cycle.
- Flood current is always computed as floodIntensity * 1500 [mA]

```
30 FPS, 30% duty cycle --> max 10ms from 33.3ms frame-time

frame-time |---------33.3ms---------|---------33.3ms---------|---------33.3ms---------|
exposure          <------20ms------>                <--10ms->                  <-8ms->
          _        _________                         ________                   ______
STROBE     |______|   10ms  |_______________________|  10ms  |_________________| 8ms  |
                                                                                 ^ Compensate with higher current
```

### DIFF
--- a/cmake/Depthai/DepthaiDeviceSideConfig.cmake
+++ b/cmake/Depthai/DepthaiDeviceSideConfig.cmake
@@ -2,7 +2,7 @@
 set(DEPTHAI_DEVICE_SIDE_MATURITY "snapshot")
 
 # "full commit hash of device side binary"
-set(DEPTHAI_DEVICE_SIDE_COMMIT "1cf15832ab1f408d8e4dab72e901ce050bf8850b")
+set(DEPTHAI_DEVICE_SIDE_COMMIT "3bba9ec3238bfb254174402253c0ecbe7dd9a71a")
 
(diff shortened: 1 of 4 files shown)


moratom (approved):
LGTM.

PR #933, opened 2023-12-05
Use uint64_t for profilingData read/write counts

asahtik:


themarpe (approved):
LGTM if xlink datatype changes propagated through (ProfilingData is long long, so looks good)

PR #944, opened 2023-12-16
Fix build with gcc 13.2 toolchain

SJLC:
To fix a couple compile errors along these lines,

/usr/include/fmt/core.h: In instantiation of 'constexpr fmt::v10::detail::value<Context> fmt::v10::detail::make_arg(T&) -- snip some typical C++ compiler noise --
src/device/DataQueue.cpp:296:45:   required from here
/usr/include/fmt/core.h:1580:7: error: static assertion failed: Cannot format an argument.

avoid trying to format atomic_t directly, and instead pass in the simple type being protected by that atomic_t (which in this case, is size_t).


themarpe (approved):
Interesting case - thanks for the PR!

PR #947, opened 2024-01-10
WIP: Record and Replay

asahtik:
TODO:
 - Make playable video files (including non-encoded ImgFrame) + separate metadata (timestamp, ...) files
 - Pin down the format of saved recordings and metadata
 - Replay: decode, format frames
 - Replay: add missing metadata (height, width, frame type, IMU sequence, ...)

---
 
 Record functionality is enabled by setting the DEPTHAI_RECORD environment variable. The variable value can be either a path to an output directory (where temporary stream files and the final tarball will be written), or a path to a config json file with the following format (suggestions welcome):
```
{
	"outputDir": "/home/user/depthai/recordings",
	"videoEncoding": {
		"enabled": true,
		"bitrate": 8000000,
		"profile": "mjpeg", // h264_baseline, avc_baseline, h264_high, avc_high, h264_main, avc_main, h265_main, hevc_main, mjpeg
		"lossless": true, // Only compatible with mjpeg
		"quality": 80 // Only compatible with mjpeg
	},
	"byteEncoding": {
		"enabled": true,
		"compressionLevel": 6 // 1-9
	}
}
```


src/device/Device.cpp
```diff
@@ -252,11 +295,167 @@ std::string Device::getQueueEvent(std::chrono::microseconds timeout) {
...
+                        recordStreams[nodeName].compress =
+                            recordConfig.byteEncoding.enabled && recordConfig.byteEncoding.compressionLevel > 0 && !nodeParams.isVideo;
+                        recordStreams[nodeName].compressionLevel = recordConfig.byteEncoding.enabled ? recordConfig.byteEncoding.compressionLevel : 0;
+                        auto xout = pipelineCopy.create<dai::node::XLinkOut>();
+                        xout->setStreamName(nodeName);
+                        if(nodeParams.isVideo) {
+                            if(recordConfig.videoEncoding.enabled) {
+                                auto imageManip = pipelineCopy.create<dai::node::ImageManip>();
+                                imageManip->initialConfig.setFrameType(ImgFrame::Type::NV12);
+                                imageManip->setMaxOutputFrameSize(3110400);
```
moratom, line 408: magic number?

include/depthai/utility/RecordConfig.hpp
```diff
@@ -0,0 +1,94 @@
...
+            p.videoEncoding.profile = Profile::H264_BASELINE;
+        } else if(profile == "h264_high" || profile == "avc_high") {
+            p.videoEncoding.profile = Profile::H264_HIGH;
+        } else if(profile == "h264_main" || profile == "avc_main" || profile == "h264" || profile == "avc") {
+            p.videoEncoding.profile = Profile::H264_MAIN;
+        } else if(profile == "h265_main" || profile == "hevc_main" || profile == "h265" || profile == "hevc") {
+            p.videoEncoding.profile = Profile::H265_MAIN;
+        } else if(profile == "mjpeg") {
+            p.videoEncoding.profile = Profile::MJPEG;
+        }
```
moratom, line 85: Why is serialization needed here? Are we sending this to device?

src/device/Device.cpp
```diff
@@ -252,11 +295,167 @@ std::string Device::getQueueEvent(std::chrono::microseconds timeout) {
...
+            recordStream->file.write((char*)buf, size);
+
+            recordStream->fileMeta.write((char*)&byteOffset, sizeof(byteOffset));
+            int64_t seqNum = msgBuf->getSequenceNum();
+            recordStream->fileMeta.write((char*)&seqNum, sizeof(msgBuf->getSequenceNum()));
+            int64_t seconds = std::chrono::duration_cast<std::chrono::seconds>(msgBuf->getTimestampDevice().time_since_epoch()).count();
+            int64_t nanoseconds = std::chrono::duration_cast<std::chrono::nanoseconds>(msgBuf->getTimestampDevice().time_since_epoch()).count() % 1000000000;
+            recordStream->fileMeta.write((char*)&seconds, sizeof(seconds));
+            recordStream->fileMeta.write((char*)&nanoseconds, sizeof(nanoseconds));
+            byteOffset += size;
```
moratom, line 368: Maybe we can define some per message record struct that we can write.
For example we likely want additional metadata for ImgFrames, etc.

src/device/Device.cpp
moratom, line 408: Where does this limit come frome?

moratom (commented):
Seems like a good approach, neat!

I think we could add subset of original frame data for metadata saving - per message.
Then for the metadata files - we can future proof it a bit, to introduce a little bit of structure and preset the files with some metadata about the file that contains the metadata.

For example:
* Message size
* Number of messages

Then, even if we add some additional metadata, we will still be able to read the old metadata files and vice versa, as long as the new fields are at the end.

include/depthai/utility/RecordConfig.hpp
asahtik: A record config json is written to the final tar file

src/device/Device.cpp
asahtik: Good catch. I put this limit in as it's the lowest one that would let me run the pipeline (I believe I had the resolution set to 1080p). I will replace this with a value depending on the source isp resolution.

PR #950, opened 2024-01-17
Support for OAK-T

zrezke:
## Support for OAK-T
Screenshot from 2024-01-17 16-07-20


examples/Camera/thermal_cam.cpp
```diff
@@ -0,0 +1,88 @@
...
+    mouseY = y;
+}
+
+const cv::Scalar WHITE(255, 255, 255);
+
+int main() {
+    dai::Pipeline pipeline;
+    auto thermal = pipeline.create<dai::node::Camera>();
+    thermal->setBoardSocket(dai::CameraBoardSocket::CAM_E);
+    thermal->setFps(25);
```
themarpe, line 22: lets add a name and refer to this as "thermal" & `setCamera("thermal")`

Is set Fps needed? Lets add some comments to that places that are "required" / optional to configure, such that example is a bit more explanatory

examples/Camera/thermal_cam.cpp
```diff
@@ -0,0 +1,88 @@
...
+            }
+            double min, max;
+            cv::minMaxLoc(frameFp32, &min, &max);
+            auto textColor = WHITE;
+            // Draw crosshair
+            cv::line(colormapped, cv::Point(mouseX - 10, mouseY), cv::Point(mouseX + 10, mouseY), textColor, 1);
+            cv::line(colormapped, cv::Point(mouseX, mouseY - 10), cv::Point(mouseX, mouseY + 10), textColor, 1);
+            // Draw deg C
+            char text[32];
+            sprintf(text, "%.1f deg C", frameFp32.at<float>(mouseY, mouseX));
```
themarpe, line 72:
```suggestion
            snprintf(text, sizeof(text), "%.1f deg C", frameFp32.at<float>(mouseY, mouseX));
```

examples/Camera/thermal_cam.cpp
```diff
@@ -0,0 +1,88 @@
...
+const cv::Scalar WHITE(255, 255, 255);
+
+int main() {
+    dai::Pipeline pipeline;
+    auto thermal = pipeline.create<dai::node::Camera>();
+    thermal->setBoardSocket(dai::CameraBoardSocket::CAM_E);
+    thermal->setFps(25);
+    int width = 256;
+    int height = 192;
+    thermal->setPreviewSize(width, height);
```
themarpe, line 25: This might not always be the case (in future if we update the sensor on a new device for instance). Cannot the FW infer this and incoming ImgFrames carrying correct values?

themarpe (changes requested):

examples/Camera/thermal_cam.cpp
zrezke: Agree, I'll change it up, this is more or less a leftover from development, left on TODO. Will modify the behaviour of camera node to take into account the sensor resolution when selecting a preview width, height. Currently it's just a hardcoded default of 300x300. Other uses of width, height can be replaced by getWidth, getHeight.

moratom (changes requested):

moratom (approved):

themarpe (approved):
LGTM - we might change to `Thermal` (separate node) later on CC: @moratom

PR #951, opened 2024-01-17
Mock isp

zrezke:
Fixed and uncommented Camera node mockIsp, added mockIsp to ColorCamera and MonoCamera nodes.


zrezke:
cc: @ahrastnik for replay

moratom (commented):
Thanks Filip!

Was the non blocking behavior of the input queues here intentional?
I'm thinking if it would be better to have them blocking by default to (by default) always replay all frames, when replaying will be implemented.

Otherwise LGTM to merge either way after we add some test coverage.

moratom:
Merged in `v3_develop`

PR #954, opened 2024-01-24
Added ability to set the lens position via a float, to enable a more precies movement.

zrezke:


themarpe (approved):
LGTM on all. Just one additional thing - add means of being able to retrieve this value from ImgFrame as well (part of metadata)

PR #974, opened 2024-03-05
Default to larger timeouts

moratom:


themarpe (commented):
I think these should be made protocol specific (and within ETH, speed specific - after timing the initial FW upload "boot" time)

moratom:
> I think these should be made protocol specific (and within ETH, speed specific - after timing the initial FW upload "boot" time)
Makes sense yes.

I'll make sure those get in the next release to not rush this for this one.

PR #977, opened 2024-03-12
Nn archive

jakgra:


include/depthai/common/optional.hpp
```diff
@@ -25,6 +25,12 @@ struct adl_serializer<tl::optional<T>> {
                                // adl_serializer<T>::from_json
         }
     }
+    static tl::optional<T> from_json(const json& j) {
+        if(j.is_null())
+            return tl::make_optional<T>();
+        else
+            return tl::make_optional<T>(j.get<T>());
+    }
```
moratom, line 33: How come this is needed?
(if we bumped C++ to 17)

src/utility/ErrorMacros.hpp
```diff
@@ -0,0 +1,39 @@
+#include <stdexcept>
+
+#include "build/version.hpp"
+#include "utility/spdlog-fmt.hpp"
+
+// clang-format off
+// NOLINTBEGIN(cppcoreguidelines-macro-usage)
+
+// Only use this one for internal errors. Clearly invalid states that shouldn't happen.
+#define daiCheckIn(A) \
```
themarpe, line 10: Namespace "prefix" + ALL CAPS for defines
```suggestion
#define DEPTHAI_CHECK_IN(A) \
```

src/utility/ArchiveUtil.cpp
```diff
@@ -0,0 +1,98 @@
+#include "ArchiveUtil.hpp"
+
+#include "depthai/pipeline/node/DetectionNetwork.hpp"
+#include "utility/ErrorMacros.hpp"
+#include "utility/Logging.hpp"
+
+namespace dai::utility {
```
themarpe, line 7: C++17? Otherwise either retain one namespace per line or tweak others as well

include/depthai/pipeline/node/DetectionNetwork.hpp
```diff
@@ -17,8 +15,41 @@ namespace node {
...
+        TAR,
+        /**
+         * Force libarchive to treat the file as .tar.gz
+         */
+        TAR_GZ,
+        /**
+         * Force libarchive to treat the file as .tar.xz
+         */
+        TAR_XZ,
+    };
```
themarpe, line 42: Make `NNArchive` first class citizen of `dai::` instead of it being baked under DetectionNetwork

src/pipeline/node/DetectionNetwork.cpp
```diff
@@ -12,6 +22,78 @@ namespace node {
...
+                                                                              const NNArchiveFormat format,
+                                                                              bool& isJson,
+                                                                              std::string& blobPath) {
+    const auto filepath = path.string();
+#if defined(_WIN32) && defined(_MSC_VER)
+    const char separator = '\\';
+#else
+    const char separator = '/';
+#endif
+    const size_t lastSlashIndex = filepath.find_last_of(separator);
```
moratom, line 42: Could we use std::filesystem hereto avoid the `#if defined`?

themarpe (changes requested):
- `NNArchive` instead of `NnArchive` throughout
- Separate C++17 move into a separate PR IMO, to be more "clear" when & what changes happened for that

src/utility/ErrorMacros.hpp
jakgra: ok will change to all caps

moratom (approved):
Looks great, thanks @jakgra!

src/utility/ArchiveUtil.cpp
themarpe: I lean towards nested namespaces - if there will anytime be C++14/11 ingestion of eg headers, etc..., it'll be simpler to "tweak"/remain compatible with older standards.

---

So lets do nested for now & disable warning for nested

include/depthai/pipeline/node/DetectionNetwork.hpp
jakgra: NnArchive class is auto generated so it requires an additional tweak inside the generator to change it to NNArchive. That is why I left it as NnArchive. Should I add the additional tweak?

jakgra: I will rename NNArchiveFormat to ArchiveFromat and make it available directly in dai:: namespace

themarpe: Check OpenVINO::Blob for rough comparison. The NnArchive serialization, that is fine if its generated, but there will likely be logic surronding that ser/des representation. So lets hide the serdes under the "first class citizen" NNArchive, which is the only interface that user sees

src/pipeline/node/DetectionNetwork.cpp
jakgra: yes we could. Can I use std::filesystem (it will add another library...)?

include/depthai/common/optional.hpp
jakgra: I can change it everywhere. But should this be part of the nn_archive branch... . It's better to maybe create a cpp_17 branch and do all the fixes there...? (also the nested namespaces fix...)

moratom (approved):

PR #980, opened 2024-03-13
Fix finding jsoncpp

Serafadam:
Fixes multi-include bug that occurs for jsoncpp.
More information https://github.com/open-source-parsers/jsoncpp/issues/1356


moratom:
@Serafadam thanks! How did you hit into that issue? Including core into some project that already uses `jsoncpp`?

I'm thinking if we should do a bugfix release or not...

Serafadam:
This occurs when builting [RTABMap](https://github.com/introlab/rtabmap) standalone with DepthAI support

moratom (approved):

cmake/depthaiDependencies.cmake
```diff
@@ -76,8 +76,9 @@ endif()
 
 # OpenCV 4 - (optional, quiet always)
 find_package(OpenCV 4 QUIET CONFIG)
-
-find_package(jsoncpp QUIET)
+if(NOT TARGET JsonCpp::JsonCpp)
+    find_package(jsoncpp QUIET)
```
danzvara, line 80: @Serafadam I tried using this on OSX, but the problem persist. However I found that asking hunter to install the package does the job. So maybe this could help?

```suggestion
    hunter_add_package(jsoncpp)
    find_package(jsoncpp QUIET)
```

moratom: @Serafadam would you be able to check if the change from @danzvara works for you?

Using hunter to find it should be fine IMO. CC: @zrezke for v2.26

Issue #995, opened 2024-04-01
`DeviceBootloader::flashUserBootloader()` fails yet it returns success


diablodale:
On my oak-d-pro-poe, I have factory 0.0.17 bootloader.
I've written c++ code that uses the Bootloader apis to allow my users to update their firmware.
Using depthai-core 2.25.0 which has embedded bootloader 0.0.26.
The firmware update API returned `true` for success, yet the bootloader actually did not update to 0.0.26. It remained on 0.0.17.
Likely related to https://github.com/luxonis/depthai-core/issues/988

### Pseudocode

```c++
deviceBootloader = std::make_unique<dai::DeviceBootloader>(candidate, true);
if (deviceBootloader->isUserBootloaderSupported()) {   // this returns true
    std::tie(success, resultMessage) = deviceBootloader->flashUserBootloader(localFlashCallback); // this returns {true, "blahblah"}
}
if (!success)
    throw
deviceBootloader.reset();
// wait
deviceBootloader = std::make_unique<dai::DeviceBootloader>(candidate);
assert(deviceBootloader.getVersion() == {0, 0, 17});
```

Unplugging/plugging mains power to POE does not help. The bootloader continues to be 0.0.17.

### Suspicion

I suspect that the rules of https://docs.luxonis.com/projects/api/en/latest/components/bootloader/#danger-zoneg and bug #988 are interacting to result in the failure of this bug.
`isUserBootloaderSupported()` is coded to check against bootloader version 0.0.21. Since #988 returns the wrong version, the compare is 0.0.26 > 0.0.21 so Supported() returns true. But should it return false instead?

Also, as can be seen from the DEPTHAI debug log, the firmware update process did SOMETHING for 18 seconds and then declared it did the flash `Success flashing the configuration userBlSize to '3824072' and userBlChecksum to '2142766317'`
Below is the log, I add comments with `//` at the end of some lines

```
[2024-04-01 01:35:47.042] [depthai] [info] Search for 1 devices with criteria DeviceInfo(name=192.168.2.23, mxid=18443010318EF50800, X_LINK_BOOTLOADER, X_LINK_TCP_IP, X_LINK_MYRIAD_X, X_LINK_SUCCESS) + hint=false
[...]

Issue #988, opened 2024-03-20
`DeviceBootloader::getVersion()` returns wrong version when `DeviceBootloader(allowFlashingBootloader=true)`


diablodale:
When `allowFlashingBootloader=true`  --> getVersion() returns the embedded blob version (WRONG)
When `allowFlashingBootloader=false`  --> getVersion() returns the version on the device (CORRECT)

depthai-core v2.25.0 (reminder that this embeds bootloader v0.0.26)
OAK-D-PRO-POE with bootloader firmware v0.0.17

```
auto deviceBootloader = std::make_unique<dai::DeviceBootloader>(targetDeviceInfo, false);
const auto test_good = deviceBootloader->getVersion();
assert(test_good == v0.0.17)

auto deviceBootloader = std::make_unique<dai::DeviceBootloader>(targetDeviceInfo, true);
const auto test_bad = deviceBootloader->getVersion();
assert(test_bad == v0.0.26)  //  💣 this assert fails
```

diablodale:
Don't use `DeviceBootloader::getVersion()`. That's what I have learned.

After collect bit-and-pieces of information and some reverse engineering, this is an *incomplete* function I've written to get the actual version of a bootloader saved in the device's flash.

```c++
// DeviceBootloader::getVersion() is not reliable, use this instead
// BUGBUG https://github.com/luxonis/depthai-core/issues/988 errant versions when flash=true
std::optional<dai::DeviceBootloader::Version> getBootloaderVersionStoredOnDevice(const dai::DeviceBootloader& deviceBootloader) {
    // allowFlashingBootloader=true always dynamically boots the embedded blob bootloader which overwrites the information of the actual stored bootloader
    if (deviceBootloader.isAllowedFlashingBootloader()) {
        if ([[maybe_unused]] const auto memInfo = deviceBootloader.getMemoryInfo(dai::DeviceBootloader::Memory::FLASH); memInfo.available) {
            [[maybe_unused]] const auto appInfo = deviceBootloader.readApplicationInfo(dai::DeviceBootloader::Memory::FLASH);
            DPOST(std::format("Memory FLASH, size {}, info {}, app flashed {}, firmware version {}, app name {}\n", memInfo.size, memInfo.info, appInfo.hasApplication, appInfo.firmwareVersion, appInfo.applicationName));
        }
        else {
            DPOST("Memory FLASH not available\n");
        }
        assert(deviceBootloader.isEmbeddedVersion());
[...]

Issue #993, opened 2024-03-25
`getConnectedCameraFeatures()` returns invalid `CameraSensorConfig`


diablodale:
With OAK-D-LITE (one of them without an imu), a call to `getConnectedCameraFeatures()` returns invalid `CameraSensorConfig` entries.
There are two config entries for both mono cameras on this device that are impossible. Below is the `struct CameraFeatures` for one of these mono cameras. We can see this is a camera named "left" (which is a non-color camera since it is always named "color"). We can see its `supportedTypes` contains only `dai::CameraSensorType::MONO`. Yet, when we look at `dai::CameraFeatures::configs` we can see 4 entries...and 2 of them are impossible. These two configs have `dai::CameraSensorConfig::type = dai::CameraSensorType::COLOR`. That is impossible. We have already proven this "left" `dai::CameraSensorType::MONO` camera. It is not possible for a MONO camera to have COLOR configs.

I do see the depthai-shared comment
```c++
/* For some sensors it's not possible to determine if they are color or mono
* (e.g. OV9782 and OV9282), so this could return more than one entry
*/
std::vector<CameraSensorType> supportedTypes;
```

However `supportedTypes` contains only one (1) entry of MONO and therefore that comment is moot.

```json
{
    "additionalNames": [],
    "calibrationResolution": {
        "fov": {
            "height": 815.0,
            "width": 1295.0,
            "x": 0.0,
            "y": 0.0
        },
        "height": 800,
        "maxFps": 129.60000610351563,
        "minFps": 1.687000036239624,
        "type": 1,
        "width": 1280
    },
    "configs": [
        {
            "fov": {
                "height": 720.0,
                "width": 1280.0,
                "x": 0.0,
                "y": 40.0
            },
            "height": 720,
            "maxFps": 143.10000610351563,
            "minFps": 1.687000036239624,
            "type": 1,
            "width": 1280
        },
        {
            "fov": {
                "height": 720.0,
                "width": 1280.0,
                "x": 0.0,
                "y": 40.0
            },
            "height": 720,
            "maxFps": 143.10000610351563,
            "minFps": 1.687000036239624,
            "type": 0,
            "width": 1280
        },
        {
            "fov": {
                "height": 800.0,
                "width": 1280.0,
                "x": 0.0,
                "y": 0.0
            },
            "height": 800,
            "maxFps": 129.60000610351563,
            "minFps": 1.687000036239624,
            "type": 1,
            "width": 1280
        },
        {
            "fov": {
[...]

PR #997, opened 2024-04-04
Introduced `DeviceBootloader::getFlashedVersion` to retrieve the bootloader version flashed on the device. Fixed `isUserBootloaderSupported` - thereby fixed user bootloader flashing.

zrezke:
@diablodale reported in https://github.com/luxonis/depthai-core/issues/988 that `DeviceBootloader::getVersion()` returns the incorrect version when constructed with `allowFlashingBootloader=true`.

As you noticed - underneath depthai boots the latest embedded bootloader. As `getVersion()` returns the current running version of the bootloader - the reported version was unexpected (e.g. device with 0.0.17 bl flashed, would report bl version 0.0.26).. 

This version reporting also messes with `flashUserBootloader`, due to reliance on `getVersion()` to check if the "flashed" bootloader supports user bootloader (as reported in https://github.com/luxonis/depthai-core/issues/995).

To resolve this issue - this PR introduces a `DeviceBootloader::getFlashedVersion()` api to get the flashed bl version, `isUserBootloaderSupported()` now relies on `getFlashedVersion` to decide whether flashing a user bootloader is supported. `getVersion` is left unchanged and still reports the version of the booted bootloader.

@jakaskerl could we enhance the docs in this regard. We could include some comments about which version `getVersion` reports etc.

Thanks @diablodale for the thorough bug reports.

### DIFF
--- a/src/device/DeviceBootloader.cpp
+++ b/src/device/DeviceBootloader.cpp
@@ -416,6 +416,7 @@ void DeviceBootloader::init(bool embeddedMvcmd, const dai::Path& pathToMvcmd, tl
 
             // Retrieve bootloader version
             version = requestVersion();
+            flashedVersion = version;
             if(version >= Version(0, 0, 12)) {
                 // If version is adequate, do an in memory boot.
 
@@ -586,6 +587,10 @@ DeviceBootloader::Version DeviceBootloader::getVersion() const {
     return version;
 }
 
+tl::optional<DeviceBootloader::Version> DeviceBootloader::getFlashedVersion() const {
+    return flashedVersion;
+}
+
 DeviceBootloader::Version DeviceBootloader::requestVersion() {
     // Send request to retrieve bootloader version
     if(!sendRequest(Request::GetBootloaderVersion{})) {
@@ -697,7 +702,7 @@ bool DeviceBootloader::isUserBootloaderSupported() {
     }
 
     // Check if bootloader version is adequate
-    if(getVersion().getSemver() < Version(Request::IsUserBootloader::VERSION)) {
+    if(getFlashedVersion().value_or(DeviceBootloader::VERSION_USER_BL_UNSUPPORTED).getSemver() < Version(Request::IsUserBootloader::VERSION)) {
         return false;
     }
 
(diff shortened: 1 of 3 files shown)


src/device/DeviceBootloader.cpp
```diff
@@ -697,7 +702,7 @@ bool DeviceBootloader::isUserBootloaderSupported() {
     }
 
     // Check if bootloader version is adequate
-    if(getVersion().getSemver() < Version(Request::IsUserBootloader::VERSION)) {
+    if(getFlashedVersion().value_or(DeviceBootloader::VERSION_USER_BL_UNSUPPORTED).getSemver() < Version(Request::IsUserBootloader::VERSION)) {
```
diablodale, line 705: I recommend combining `VERSION_USER_BL_UNSUPPORTED` and `Request::IsUserBootloader::VERSION`. They represent the same idea and are only +/- 1. Having two constants for the same idea is generally unwanted and can lead to future mistakes of updating one but not the other.

moratom: Probably best to check this in two separate if statements.

src/device/DeviceBootloader.cpp
```diff
@@ -925,7 +933,7 @@ std::tuple<bool, std::string> DeviceBootloader::flashUserBootloader(std::functio
     // }
 
     // Check if bootloader version is adequate
-    if(getVersion().getSemver() < Version(Request::IsUserBootloader::VERSION)) {
+    if(getFlashedVersion().value_or(DeviceBootloader::VERSION_USER_BL_UNSUPPORTED).getSemver() < Version(Request::IsUserBootloader::VERSION)) {
         throw std::runtime_error("Current bootloader version doesn't support User Bootloader");
     }
```
moratom, line 938: Same comment as above.

In case getFlashedVersion is null, the exception message could point out that the device (likely) doesn't have a bootloader flashed at all.

zrezke: Yea agree on both.

src/device/DeviceBootloader.cpp
zrezke: Did as @moratom suggested - this approach gets rid of `VERSION_USER_BL_UNSUPORTED` completely.
Is this good to merge now?

diablodale:
Just remembered... where are the test cases?

zrezke:
@moratom / @diablodale  I will add the fix for https://github.com/luxonis/depthai-core/issues/993 in here too.

PR #1009, opened 2024-04-25
V3 develop sugared

Zimamazim:


bindings/python/src/pipeline/node/HostNodeBindings.cpp
```diff
@@ -43,10 +45,66 @@ void bind_hostnode(pybind11::module& m, void* pCallstack){
...
+                return self
+            cls.link_args = link_args
+
+            def __init__(self, *args):
+                node.HostNode.__init__(self)
+                self.link_args(*args)
+            cls.__init__ = __init__
+
+        node.HostNode.__init_subclass__ = classmethod(__init_subclass__)
+    )", m.attr("__dict__"));
```
moratom, line 109: When does this `py::exec` actually run?

bindings/python/src/pipeline/node/HostNodeBindings.cpp
```diff
@@ -43,10 +45,66 @@ void bind_hostnode(pybind11::module& m, void* pCallstack){
...
+                        for hierarchy in arg.getPossibleDatatypes():
+                            # I believe this check isn't sound nor complete
+                            # However, nether does the original in canConnect
+                            # I belive it would be more confusing to have two
+                            # different behaviours than one incorrect
+                            if type_enum == hierarchy.datatype: break
+                            if isDatatypeSubclassOf(type_enum, hierarchy.datatype): break
+                        else:
+                            raise TypeError(f"Input '{name}' cannot be linked due to incompatible message types. Input type: {type_enum} Output type: {hierarchy.datatype}")
+                    arg.link(self.inputs[name])
```
moratom, line 99: Do we have some guarantee here that `name` is unique?

bindings/python/src/pipeline/node/HostNodeBindings.cpp
Zimamazim: The `py::exec` is standard statement in the `bind_hostname` functions. So it is executed as any other statement in that function. I add the `__init_subclass__` through monkey patching.

bindings/python/src/pipeline/node/HostNodeBindings.cpp
Zimamazim: The name is unique as long as the parameters of the `process` method are unique. And if they are not, Python crashes first with a reasonable error message.

include/depthai/pipeline/node/host/HostNode.hpp
```diff
@@ -23,7 +23,7 @@ class HostNode : public ThreadedHostNode {
     InputMap& inputs = sync->inputs;
     // Output out{*this, "out", Output::Type::MSender, {{DatatypeEnum::Buffer, true}}};
     Output out{*this, {.name = "out", .types = {{DatatypeEnum::Buffer, true}}}};
-    virtual std::shared_ptr<Buffer> runOnce(std::shared_ptr<dai::MessageGroup> in) = 0;
+    virtual std::shared_ptr<Buffer> _process(std::shared_ptr<dai::MessageGroup> in) = 0;
```
moratom, line 26: Let's rename the method only in the bindings, starting methods with an `_` is not a standard practice in the C++.

moratom (changes requested):
Thanks @Zimamazim !

Looks great, let's just add some tests, specfically at least for:
* Checking if names are unique, as them not being unique would cause unexpected behavior
* Multi-threading - does it work as expected

Also please rename the `runOnce` function throughout the code base so the examples will still work.

include/depthai/pipeline/node/host/HostNode.hpp
Zimamazim: What about `processGroup`?

bindings/python/src/pipeline/node/HostNodeBindings.cpp
moratom: Perhaps still worth checking that the input doesn't already exist, to be defensive.

As one can still now manually add inputs in `self.inputs`.

Zimamazim: I wouldn't allow mixing `process` magic with user defining his own inputs through `self.inputs`. I suggest renaming `inputs` to `_inputs` to indicate that choice. If we wan't to be super defensive, I'd check that `self._inputs` is empty at the beginning of `link_args`.

moratom: I guess now the user must pass the inputs in a named form?

Would it make sense to lift that requirement, let the arguments be normal *args and generate unique names for the inputs?

Zimamazim: I'm not sure I'm following. Currently, user needs to pass the arguments (inputs) as positional arguments. Passing through key-word arguments (i.e. in a named form) is currently not allowed.

Would you like to pass them as kwargs? At first glance, I don't think this matches the rest of the codebase well as we wanted to use positional arguments for inputs and kwargs for parameters.

 I'm not sure why would you want to use generated names instead of parameter names from the `process` method.

bindings/python/tests/multithreaded_pipeline_creation.py
```diff
@@ -0,0 +1,18 @@
...
+def createPipeline():
+    with dai.Pipeline(createImplicitDevice=False) as pipeline:
+        X() 
+        time.sleep(1) # Yield to other threads
+        X()
+        assert sum(1 for node in pipeline.getAllNodes() if isinstance(node, dai.node.HostNode)) == 2
+
+for i in range(500):
+    thread = threading.Thread(target=createPipeline)
+    thread.start()
```
moratom, line 18: Does this wait for threads to stop implicitly?

bindings/python/src/pipeline/node/HostNodeBindings.cpp
moratom: Oh right, no it's fine like it is.

I got confused above as I thought the names are somehow gotten from the constructor parameters, which don't have any named arguments by default.

Looks good then, sorry for the confusion.

PR #1011, opened 2024-04-30
V3 develop build

moratom:


include/depthai/pipeline/node/DetectionNetwork.hpp
```diff
@@ -161,27 +164,23 @@ class DetectionNetwork : public NodeGroup {
...
+    std::shared_ptr<MobileNetDetectionNetwork> build();
 };
 
 /**
  * @brief YoloDetectionNetwork node. Parses Yolo results
  */
 class YoloDetectionNetwork : public DetectionNetwork {
    public:
-    void build();
+    std::shared_ptr<YoloDetectionNetwork> build();
```
moratom, line 181: Would it be better perhaps for the build methods on all detection network to require an input?

Maybe this would be best for all detection networks:
* build(Input, NNArchive) # For all 
* build(Input, model, inputSize) # for MobileNet
* build(Input, model, allOtherParams) # for Yolo

Zimamazim: Yeah, definitely. However, I just implemented the few nodes for testing - `Camera`, `DetectionNetwork`, `StereoDepth` and `VideoEncoder`. I plan to do others like Yolo and MobileNet with the rest after we settle on the API

include/depthai/pipeline/node/NeuralNetwork.hpp
```diff
@@ -26,6 +26,10 @@ class NeuralNetwork : public DeviceNodeCRTP<DeviceNode, NeuralNetwork, NeuralNet
     std::optional<OpenVINO::Version> networkOpenvinoVersion;
 
    public:
+
+    std::shared_ptr<NeuralNetwork> build() {
+        return std::static_pointer_cast<NeuralNetwork>(shared_from_this());
+    }
```
moratom, line 32: Due to multiple inputs it actually does make sense to have an empty one.

But the most common case is to pass in an input, so I'd add a `build(Input)` too.

include/depthai/pipeline/node/PointCloud.hpp
```diff
@@ -26,6 +26,11 @@ class PointCloud : public DeviceNodeCRTP<DeviceNode, PointCloud, PointCloudPrope
     using DeviceNodeCRTP::DeviceNodeCRTP;
 
    public:
+
+    std::shared_ptr<PointCloud> build() {
+        return std::static_pointer_cast<PointCloud>(shared_from_this());
+    }
+
```
moratom, line 33: Always needs depth as input

include/depthai/pipeline/node/Warp.hpp
```diff
@@ -17,6 +17,9 @@ class Warp : public DeviceNodeCRTP<DeviceNode, Warp, WarpProperties> {
    public:
     constexpr static const char* NAME = "Warp";
     using DeviceNodeCRTP::DeviceNodeCRTP;
+    std::shared_ptr<Warp> build() {
+        return std::static_pointer_cast<Warp>(shared_from_this());
+    }
```
moratom, line 22: Always needs something on the input

src/pipeline/node/Benchmark.cpp
```diff
@@ -16,5 +16,12 @@ void BenchmarkIn::setNumMessagesToGet(int num) {
     properties.numMessages = num;
 }
 
+std::shared_ptr<BenchmarkOut> BenchmarkOut::build() {
+        properties.numMessages = -1;  // By default send messages indefinitely
+        isBuild = true; 
+        return std::static_pointer_cast<BenchmarkOut>(shared_from_this());
+    }
```
moratom, line 23: Needed?

We can just change the default in properties IMO

src/pipeline/node/Camera.cpp
```diff
@@ -18,6 +18,8 @@ void Camera::build() {
     properties.previewWidth = 300;
     properties.fps = 30.0;
     properties.previewKeepAspectRatio = true;
```
moratom, line 20: I'd remove the overriding the properties here (I know this is very old and not your change, just noting it down here so it's in once place) and change the defaults in properties directly.

src/pipeline/node/ColorCamera.cpp
```diff
@@ -34,6 +34,9 @@ void ColorCamera::build() {
     properties.resolution = ColorCameraProperties::SensorResolution::THE_1080_P;
     properties.fps = 30.0;
     properties.previewKeepAspectRatio = true;
+
```
moratom, line 37: Same comment as above

src/pipeline/node/MonoCamera.cpp
```diff
@@ -5,10 +5,13 @@
 namespace dai {
 namespace node {
 
-void MonoCamera::build() {
+std::shared_ptr<MonoCamera> MonoCamera::build() {
     properties.boardSocket = CameraBoardSocket::AUTO;
     properties.resolution = MonoCameraProperties::SensorResolution::THE_720_P;
     properties.fps = 30.0;
```
moratom, line 11: Same comment as for other cameras

PR #1016, opened 2024-05-07
Depth align node

moratom:


include/depthai/pipeline/node/DepthAlign.hpp
```diff
@@ -0,0 +1,73 @@
...
+    /**
+     * Outputs ImgFrame message that carries aligned depth image.
+     */
+    Output outputAlignedDepth{*this, "outputAlignedDepth", Output::Type::MSender, {{DatatypeEnum::ImgFrame, false}}};
+
+    /**
+     * Passthrough message on which the calculation was performed.
+     * Suitable for when input queue is set to non-blocking behavior.
+     */
+    Output passthroughDepth{*this, "passthroughDepth", Output::Type::MSender, {{DatatypeEnum::ImgFrame, false}}};
```
moratom, line 58: Thoughts on changing the interface to:
* `input` the ImgFrame stream that gets aligned (can be `Depth` but also a normal `ImgFrame`)
* `inputAlignTo` stream to what the `input` should be aligned to. Used only to get the socket (and later on perhaps transformations) out of it.

Then in case the input is depth, we use it do perfect alignment, if the input is not depth we only align with the intrinsics and compensate for rotation. 

This prepares us for a bright future where we might also take into account any prior transformations and such interface wise.
For now we can warn `The Align node will align the image to the source image frame while the frame received on inputAlignTo has been transformed, so it might not work as expected`.


Thoughts @SzabolcsGergely ?

moratom: I think that would make sense yes.

Perhaps we leave the current API to be able to override though.

SzabolcsGergely: Interface updated

moratom: Thanks!

PR #1019, opened 2024-05-08
Cast node

moratom:


include/depthai/pipeline/node/Cast.hpp
```diff
@@ -0,0 +1,61 @@
...
+     * @param outputType Output frame type
+     */
+    Cast& setOutputFrameType(dai::RawImgFrame::Type outputType);
+
+    /**
+     * Set scale
+     * @param scale Scale
+     */
+    Cast& setScale(float scale);
+};
```
moratom, line 58: Is it trivial to add offset too?

Aka
`outImg[i][j] = offset + scale * NNDataLayer[i][j]` @SzabolcsGergely ?

PR #1020, opened 2024-05-10
Bootup/stability fixes

alex-luxonis:
- fix a potential bootup/reboot failure, that required power cycle to recover from (more likely to occur on PoE devices). Requires the **factory** bootloader to be upgraded
- fix a potential packet metadata corruption, due to some memory alignment in FW
- add more details in the exceptions thrown in case of a bad packet
- `setXLinkRateLimit` API - new feature for bandwidth throttling:
```
    /**
     * Sets the maximum transmission rate for the XLink connection on device side,
     * using a simple token bucket algorithm. Useful for bandwidth throttling
     *
     * @param maxRateBytesPerSecond Rate limit in Bytes/second
     * @param burstSize Size in Bytes for how much to attempt to send once, 0 = auto
     * @param waitUs Time in microseconds to wait for replenishing tokens, 0 = auto
     */
    void setXLinkRateLimit(int maxRateBytesPerSecond, int burstSize = 0, int waitUs = 0);
```

TODO:
- bootloader release before merge
- add also a fix for a different issue: C++ app not being able to recover after the automatic crashdump check - terminate called even with exceptions caught, possibly a missing thread join on recreating Device() internally


themarpe (commented):
>TODO:
> - bootloader release before merge
> - add also a fix for a different issue: C++ app not being able to recover after the automatic crashdump check - terminate called even with exceptions caught, possibly a missing thread join on recreating Device() internally


The crashdump was mitigated before - lets release the new BL and prepare for merge

moratom (approved):
LGTM!

PR #1029, opened 2024-05-26
Add input queues to the new API

lnotspotl:
This PR renames the old `createQueue` method to `createOutputQueue` for the `Output` class. The old function stayed in the code, however it was marked as deprecated.

Additionally, a new class, `InputQueue` was introduced, allowing for communication between host and the connected device. A new method called `createInputQueue` was added to the `Input` class. This method returns a shared pointer to an `InputQueue` working as a communication interface between the host and the device via its `send(...)` method.

Lastly, all the `createQueue` function calls were replaced with `createOutputQueue` function calls throughout the entire codebase.

The relevant clickup task is available here: https://app.clickup.com/t/86bym8w49

### DIFF
--- a/bindings/python/src/pipeline/node/NodeBindings.cpp
+++ b/bindings/python/src/pipeline/node/NodeBindings.cpp
@@ -5,6 +5,7 @@
 #include "Common.hpp"
 #include "depthai/pipeline/DeviceNode.hpp"
 #include "depthai/pipeline/Node.hpp"
+#include "depthai/pipeline/InputQueue.hpp"
 #include "depthai/pipeline/NodeGroup.hpp"
 #include "depthai/pipeline/Pipeline.hpp"
 #include "depthai/pipeline/ThreadedNode.hpp"
--- a/bindings/python/src/pipeline/node/NodeBindings.cpp
+++ b/bindings/python/src/pipeline/node/NodeBindings.cpp
@@ -292,7 +292,11 @@ void NodeBindings::bind(pybind11::module& m, void* pCallstack) {
         .def("getWaitForMessage", &Node::Input::getWaitForMessage, DOC(dai, Node, Input, getWaitForMessage))
         .def("setReusePreviousMessage", &Node::Input::setReusePreviousMessage, py::arg("reusePreviousMessage"), DOC(dai, Node, Input, setReusePreviousMessage))
         .def("getReusePreviousMessage", &Node::Input::getReusePreviousMessage, DOC(dai, Node, Input, getReusePreviousMessage))
-        .def("createInputQueue", &Node::Input::createInputQueue, py::arg("maxSize") = 16, py::arg("blocking") = true, DOC(dai, Node, Input, createInputQueue));
+        .def("createInputQueue",
+             &Node::Input::createInputQueue,
+             py::arg("maxSize") = Node::Input::INPUT_QUEUE_DEFAULT_MAX_SIZE,
+             py::arg("blocking") = Node::Input::INPUT_QUEUE_DEFAULT_BLOCKING,
+             DOC(dai, Node, Input, createInputQueue));
 
     // Node::Output bindings
     nodeOutputType.value("MSender", Node::Output::Type::MSender).value("SSender", Node::Output::Type::SSender);
@@ -316,8 +320,11 @@ void NodeBindings::bind(pybind11::module& m, void* pCallstack) {
              DOC(dai, Node, Output, getParent))
         .def("isSamePipeline", &Node::Output::isSamePipeline, py::arg("input"), DOC(dai, Node, Output, isSamePipeline))
         .def("canConnect", &Node::Output::canConnect, py::arg("input"), DOC(dai, Node, Output, canConnect))
-        .def("createQueue", &Node::Output::createQueue, py::arg("maxSize") = 16, py::arg("blocking") = true, DOC(dai, Node, Output, createQueue))
-        .def("createOutputQueue", &Node::Output::createOutputQueue, py::arg("maxSize") = 16, py::arg("blocking") = true, DOC(dai, Node, Output, createOutputQueue))
+        .def("createOutputQueue",
+             &Node::Output::createOutputQueue,
+             py::arg("maxSize") = Node::Output::OUTPUT_QUEUE_DEFAULT_MAX_SIZE,
+             py::arg("blocking") = Node::Output::OUTPUT_QUEUE_DEFAULT_BLOCKING,
+             DOC(dai, Node, Output, createOutputQueue))
         .def("link", static_cast<void (Node::Output::*)(Node::Input&)>(&Node::Output::link), py::arg("input"), DOC(dai, Node, Output, link))
         .def("unlink", static_cast<void (Node::Output::*)(Node::Input&)>(&Node::Output::unlink), py::arg("input"), DOC(dai, Node, Output, unlink))
         .def("send", &Node::Output::send, py::arg("msg"), DOC(dai, Node, Output, send), py::call_guard<py::gil_scoped_release>())
--- a/bindings/python/src/pipeline/node/NodeBindings.cpp
+++ b/bindings/python/src/pipeline/node/NodeBindings.cpp
@@ -209,6 +209,10 @@ void NodeBindings::bind(pybind11::module& m, void* pCallstack) {
     py::enum_<Node::Output::Type> nodeOutputType(pyOutput, "Type");
     py::class_<Properties, std::shared_ptr<Properties>> pyProperties(m, "Properties", DOC(dai, Properties));
     py::class_<Node::DatatypeHierarchy> nodeDatatypeHierarchy(pyNode, "DatatypeHierarchy", DOC(dai, Node, DatatypeHierarchy));
+    
+    
+    py::class_<InputQueue, std::shared_ptr<InputQueue>> pyInputQueue(m, "InputQueue", DOC(dai, InputQueue));
+    pyInputQueue.def("send", &InputQueue::send, py::arg("msg"), DOC(dai, InputQueue, send));
 
     // Node::Id bindings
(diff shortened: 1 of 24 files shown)


bindings/python/src/pipeline/node/NodeBindings.cpp
```diff
@@ -287,7 +291,8 @@ void NodeBindings::bind(pybind11::module& m, void* pCallstack) {
         .def("setWaitForMessage", &Node::Input::setWaitForMessage, py::arg("waitForMessage"), DOC(dai, Node, Input, setWaitForMessage))
         .def("getWaitForMessage", &Node::Input::getWaitForMessage, DOC(dai, Node, Input, getWaitForMessage))
         .def("setReusePreviousMessage", &Node::Input::setReusePreviousMessage, py::arg("reusePreviousMessage"), DOC(dai, Node, Input, setReusePreviousMessage))
-        .def("getReusePreviousMessage", &Node::Input::getReusePreviousMessage, DOC(dai, Node, Input, getReusePreviousMessage));
+        .def("getReusePreviousMessage", &Node::Input::getReusePreviousMessage, DOC(dai, Node, Input, getReusePreviousMessage))
+        .def("createInputQueue", &Node::Input::createInputQueue, py::arg("maxSize") = 16, py::arg("blocking") = true, DOC(dai, Node, Input, createInputQueue));
```
moratom, line 295: It would be better to share the defaults with C++ implementation here - you can use a `constexpr` and then use it both in C++ for the default and in Python.

bindings/python/src/pipeline/node/NodeBindings.cpp
```diff
@@ -312,6 +317,7 @@ void NodeBindings::bind(pybind11::module& m, void* pCallstack) {
         .def("isSamePipeline", &Node::Output::isSamePipeline, py::arg("input"), DOC(dai, Node, Output, isSamePipeline))
         .def("canConnect", &Node::Output::canConnect, py::arg("input"), DOC(dai, Node, Output, canConnect))
         .def("createQueue", &Node::Output::createQueue, py::arg("maxSize") = 16, py::arg("blocking") = true, DOC(dai, Node, Output, createQueue))
+        .def("createOutputQueue", &Node::Output::createOutputQueue, py::arg("maxSize") = 16, py::arg("blocking") = true, DOC(dai, Node, Output, createOutputQueue))
```
moratom, line 320: Let's remove the `createQueue` and same comment for the default as for the InputQueues.

include/depthai/pipeline/Node.hpp
```diff
@@ -229,7 +234,17 @@ class Node : public std::enable_shared_from_this<Node> {
             return queueConnections;
         }
 
-        std::shared_ptr<dai::MessageQueue> createQueue(unsigned int maxSize = 16, bool blocking = true);
+        [[deprecated("Use 'createOutputQueue()' instead")]] std::shared_ptr<MessageQueue> createQueue(unsigned int maxSize = 16, bool blocking = true);
```
moratom, line 237: Let's just remove it for now, no need to keep the old function as the only people using this used it in an experimental phase

include/depthai/pipeline/Node.hpp
```diff
@@ -614,6 +640,46 @@ class Node : public std::enable_shared_from_this<Node> {
...
+
+    JoiningThread inputQueueThread_;
+    std::unique_ptr<MessageQueue> queuePtr_;
+    std::atomic_bool stopThreadFlag_;
+
+    mutable std::mutex guard_;
+    std::condition_variable sendThreadCv_;
+    std::condition_variable inputQueueEmptiedCv_;
+};
+
```
moratom, line 682: Could we do with a "private/protected" inheritance here to be able to expose only the relevant API (`send`)?

src/pipeline/Node.cpp
```diff
@@ -251,6 +267,77 @@ bool Node::Input::getReusePreviousMessage() const {
...
+    if(queuePtr_->getBlocking() && queuePtr_->isFull()) {
+        sendThreadCv_.notify_all();
+        inputQueueEmptiedCv_.wait(lock);
+    }
+
+    // If queue is non-blocking and full, send will overwrite the oldest message
+    queuePtr_->send(msg);
+
+    // Let other threads know that there is a new message
+    sendThreadCv_.notify_all();
```
moratom, line 330: Is the extra syncronisation needed here?

Generally speaking the `MessageQueue` already provides a blocking `send` and `get`

moratom (changes requested):
Thanks a lot!

Left a few nitpicks, but looks good.

include/depthai/pipeline/Node.hpp
themarpe: When was the `_` postfix introduced for private members? 
IMO, the private designator + IDE handles this

bindings/python/src/pipeline/node/NodeBindings.cpp
lnotspotl: Fixed in d89c76e2a5f9c6944ddf53dd35361029ecdd41a5. Not sure about the default values for `maxSize` and `blocking` though.

bindings/python/src/pipeline/node/NodeBindings.cpp
lnotspotl: Fixed in d89c76e2a5f9c6944ddf53dd35361029ecdd41a5.

include/depthai/pipeline/Node.hpp
lnotspotl: I've built the habit of adding the `_` suffix to my private attributes after seeing it in many codebases, though it makes sense not to include them here as it would  be inconsistent with the already existing code. I will keep this in mind in the future.

I've changed the `InputQueue`'s class declaration, exposing only the important method, i.e. `send`. This change is part of b19b0945e08b18ac6ea277f5be17d7081f7023d9

include/depthai/pipeline/Node.hpp
```diff
@@ -614,6 +651,48 @@ class Node : public std::enable_shared_from_this<Node> {
...
+    JoiningThread inputQueueThread;
+    std::unique_ptr<MessageQueue> queuePtr;
+    std::atomic_bool stopThreadFlag;
+
+    mutable std::mutex guard;
+    std::condition_variable sendThreadCv;
+    std::condition_variable inputQueueEmptiedCv;
+};
+
```
moratom, line 695: Can we maybe move to a new file, to keep the file sizes more manageable.

Same for the implementation file.

include/depthai/pipeline/Node.hpp
```diff
@@ -614,6 +651,48 @@ class Node : public std::enable_shared_from_this<Node> {
...
+    void start() override;
+    void stop() override;
+    void wait() override;
+    bool runOnHost() const override;
+    const char* getName() const override;
+
+    Output output{*this, {.name = "output", .types = {{DatatypeEnum::Buffer, true}}}};
+
+    JoiningThread inputQueueThread;
+    std::unique_ptr<MessageQueue> queuePtr;
```
moratom, line 688: Maybe the queue can have just a regular input (similar to all other nodes)?

src/pipeline/Node.cpp
```diff
@@ -251,6 +263,77 @@ bool Node::Input::getReusePreviousMessage() const {
...
+        sendThreadCv.notify_all();
+        inputQueueEmptiedCv.wait(lock);
+    }
+
+    // If queue is non-blocking and full, send will overwrite the oldest message
+    queuePtr->send(msg);
+
+    // Let other threads know that there is a new message
+    sendThreadCv.notify_all();
+}
```
moratom, line 327: If the simplification can be done - this would only pass-through the send directly to the input:

```suggestion
void InputQueue::send(const std::shared_ptr<ADatatype>& msg) {
    input->send(msg);
}
```

src/pipeline/Node.cpp
```diff
@@ -251,6 +263,77 @@ bool Node::Input::getReusePreviousMessage() const {
...
+            }
+
+            inputQueueEmptiedCv.notify_all();
+        }
+    } catch(const MessageQueue::QueueException& ex) {
+        stopThreadFlag = true;
+    } catch(const std::runtime_error& ex) {
+        std::cout << "Node threw exception, stopping the node. Exception message: " << ex.what() << std::endl;
+    }
+}
```
moratom, line 310:
```suggestion
void InputQueue::run() {
    try {
        while(isRunning()) {
              output.send(input.get())
        }
    } catch(const MessageQueue::QueueException& ex) {
        stopThreadFlag = true;
    } catch(const std::runtime_error& ex) {
        std::cout << "Node threw exception, stopping the node. Exception message: " << ex.what() << std::endl;
    }
}
```

I was thinking if this could be simplified as such?

The `start` `stop` and `wait` functions could then be left at default implementations I believe.

include/depthai/pipeline/Node.hpp
moratom: Then in the constructor - it could just readjust the queue size in the input.

lnotspotl:
Thanks a lot, @moratom. Appreciate the feedback. I was reinventing the wheel originally and with your suggestions incorporated into this PR the code is way simpler.

include/depthai/pipeline/InputQueue.hpp
```diff
@@ -0,0 +1,40 @@
+#pragma once
+
+#include "depthai/pipeline/Node.hpp"
+#include "depthai/pipeline/ThreadedHostNode.hpp"
+
+namespace dai {
+
+class InputQueue : public node::ThreadedHostNode {
+    friend class Node::Input;
```
moratom, line 9: Can we make the inheritance here protected, so the public API from the node is not exposed to the queue as well>

include/depthai/pipeline/Node.hpp
```diff
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <algorithm>
+#include <condition_variable>
```
moratom, line 4: Not used anymore right?

themarpe (approved):

PR #1032, opened 2024-05-29
Renaming cast examples

jakaskerl:
- renamed cast examples
  - Edited image_align example to remove redundant nodes and rename variables


jakaskerl:
@moratom check image_align.cpp closely as I didn't test it.

moratom (approved):

PR #1033, opened 2024-06-05
Add modelzoo functionality

lnotspotl:
This PR adds the necessary functionality needed for seamless model downloading. More specifically, there are two functions and one class in particular, namely `getModelFromZoo`, `downloadModelsFromZoo` and `NNModelDescription` respectively, that are relevant to accomplishing this task.

---

`NNModelDescription` is a class containing information as to what model is to be downloaded.

`getModelFromZoo` downloads the relevant model based on the description specified in `NNModelDescription`. All the downloaded models are cached and so if a model is present on the user's computer, the cached model is used instead of downloading the same model again.

Finally `downloadModelsFromZoo` allows one to download several models at once, all specified in `yaml` files. This is especially useful if models are to be pre-downloaded and subsequently redistributed among other computers, some potentially without access to the internet.

---

There are three more things that need to be accomplished prior to this PR being merged:
- [x] Add python example script
- [x] Add cpp example script

These two should not be as hard to do, though I encountered an issue with the NeuralNetwork node causing a core dump as written [here](https://luxonis.slack.com/archives/C065PDYKNAK/p1718997113857789)

- [ ] Perform thorough testing 

Overall, the first code review can be performed, though small tweaks are still expected.


src/modelzoo/Zoo.cpp
```diff
@@ -0,0 +1,151 @@
...
+}
+
+void ZooManager::createCacheFolder() const {
+    std::string cacheFolderName = getCacheFolderName();
+    system(("mkdir -p " + cacheFolderName).c_str());
+}
+
+void ZooManager::removeCacheFolder() const {
+    std::string cacheFolderName = getCacheFolderName();
+    system(("rm -rf " + cacheFolderName).c_str());
```
moratom, line 41: Probably not cross platform

src/modelzoo/Zoo.cpp
```diff
@@ -0,0 +1,230 @@
...
+    foldername += "model_slug=" + modelDescription.getModelSlug() + separator;
+    foldername += "platform=" + modelDescription.getPlatform() + separator;
+
+    // OPTIONAL fields
+    if(modelDescription.getmodelVersionSlug().size() > 0) foldername += "model_version_slug=" + modelDescription.getmodelVersionSlug() + separator;
+
+    // Append hash
+    foldername += "hash=" + computeModelHash();
+
+    return foldername;
```
moratom, line 32: Could we name the folder as just the hash instead to not over complicate?

include/depthai/modelzoo/Zoo.hpp
```diff
@@ -0,0 +1,126 @@
...
+
+namespace dai {
+
+class ZooManager {
+   public:
+    /**
+     * @brief Construct a new Zoo Manager object
+     *
+     * @param modelDescription: Model description
+     * @param cacheDirectory: Cache directory, default is "."
```
moratom, line 13: Let's change the default to `./.depthai_cached_models`

src/modelzoo/Zoo.cpp
```diff
@@ -0,0 +1,230 @@
...
+}
+
+void ZooManager::removeModelCacheFolder() const {
+    std::string cacheFolderName = getModelCacheFolderPath(cacheDirectory);
+    std::filesystem::remove_all(cacheFolderName);
+}
+
+bool ZooManager::isCached() const {
+    return checkExists(getModelCacheFolderPath(cacheDirectory));
+}
```
moratom, line 55: This just checks if the cache exists right?

Is it needed?

include/depthai/modelzoo/Zoo.hpp
```diff
@@ -0,0 +1,126 @@
...
+     * @param folder: Folder path
+     * @return std::vector<std::string>: List of files in the folder
+     */
+    std::vector<std::string> getFilesInFolder(const std::string& folder) const;
+
+   private:
+    const NNModelDescription& modelDescription;
+    const std::string cacheDirectory;
+};
+
```
moratom, line 106: As a nitpick comment - I'd go with stateless functions here, as the amount of state needed is rather small and disconnected.

No need to change it though.

moratom (commented):
Thanks! Left a few comments, agree with the TODO list + we need to add a binary that will download all the models when we specify a directory.

Should be a 5 liner though as I see we already have a function that esentially does it.

bindings/python/src/modelzoo/ZooBindings.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+    m.def("getModelFromZoo",
+          dai::getModelFromZoo,
+          py::arg("modelDescription"),
+          py::arg("cacheDirectory") = ".",
+          py::arg("useCached") = true,
+          py::arg("verbose") = false,
+          DOC(dai, getModelFromZoo));
+
+    m.def("downloadModelsFromZoo",
```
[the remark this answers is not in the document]
lnotspotl: I don't think the two functions are too similar. Whereas the first function, `getModelFromZoo`, downloads a single model from zoo and returns a path to the cached file, the `downloadModelsFromZoo` takes as an input a folder full of yaml model descriptions and stores the files in a specified folder, it does not return anything.

include/depthai/modelzoo/NNModelDescription.hpp
```diff
@@ -0,0 +1,83 @@
...
+
+    /**
+     * @brief Initialize NNModelDescription in code using parameters
+     *
+     * @param modelSlug: Model slug
+     * @param platform: Model platform
+     * @param modelVersionSlug: Model instance slug
+     * @return NNModelDescription
+     */
+    static NNModelDescription fromParameters(const std::string& modelSlug, const Platform platform, const std::string& modelVersionSlug = "");
```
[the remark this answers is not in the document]
lnotspotl: Instead of adding a completely new struct, I decided to make `NNModelDescription` a struct with all the parameters and each assigned a default value of an empty string. This should do the trick.

include/depthai/modelzoo/Zoo.hpp
```diff
@@ -0,0 +1,126 @@
...
+    std::string getModelCacheFolderPath(const std::string& cacheDirectory = ".") const;
+
+    /**
+     * @brief Combine two paths
+     *
+     * @param path1: First path
+     * @param path2: Second path
+     * @return std::string: Combined path
+     */
+    std::string combinePaths(const std::string& path1, const std::string& path2) const;
```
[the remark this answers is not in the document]
lnotspotl: I have not found anything.

include/depthai/modelzoo/Zoo.hpp
lnotspotl: The idea I had in mind was to have a `ZooManager` class knowing what the cache directory is and what model it should take care of and then have specific functions (without any arguments) exposed to perform needed actions.

If needed, we can have a discussion internally, and I will change the interface as requested.

src/modelzoo/Zoo.cpp
lnotspotl: Do you think we should inline the function, i.e. replace the one occurence of `isCached()` with `checkExists(...)`?

examples/python/RVC2/ModelZoo/model_zoo.py
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/env python3
```
moratom, line 1: camelCase to keep examples consistent

src/modelzoo/zoo_helper.cpp
```diff
@@ -0,0 +1,21 @@
...
+
+    // Unpack arguments
+    const std::string yamlFolder = argv[1];
+    const std::string cacheFolder = argv[2];
+
+    // Download models
+    dai::downloadModelsFromZoo(yamlFolder, cacheFolder, true); // true for verbose
+
+    return EXIT_SUCCESS;
+}
```
moratom, line 21: This needs to accept platform - do we have a default to RVC2?

bindings/python/src/modelzoo/NNModelDescriptionBindings.cpp
```diff
@@ -0,0 +1,39 @@
...
+        .def("toString", &NNModelDescription::toString, DOC(dai, NNModelDescription, toString))
+        .def("__str__", &NNModelDescription::toString, DOC(dai, NNModelDescription, toString))
+        .def_readwrite("modelSlug", &NNModelDescription::modelSlug, DOC(dai, NNModelDescription, modelSlug))
+        .def_readwrite("platform", &NNModelDescription::platform, DOC(dai, NNModelDescription, platform))
+        .def_readwrite("modelVersionSlug", &NNModelDescription::modelVersionSlug, DOC(dai, NNModelDescription, modelVersionSlug))
+        .def_readwrite("modelInstanceSlug", &NNModelDescription::modelInstanceSlug, DOC(dai, NNModelDescription, modelInstanceSlug))
+        .def_readwrite("modelInstanceHash", &NNModelDescription::modelInstanceHash, DOC(dai, NNModelDescription, modelInstanceHash))
+        .def_readwrite("optimizationLevel", &NNModelDescription::optimizationLevel, DOC(dai, NNModelDescription, optimizationLevel))
+        .def_readwrite("compressionLevel", &NNModelDescription::compressionLevel, DOC(dai, NNModelDescription, compressionLevel))
+        ;
```
moratom, line 38: Let's add a constructor in python with kwargs for all the parameters (it can be nice in python, not so much in C++)

src/modelzoo/zoo_helper.cpp
lnotspotl: Each yaml file has has to contain two required parameters, namely `modelSlug` and `platform`. If not found, the program throws an error. Do you think we should make the `platform` parameter optional and set the value to `RVC2` if not specified? It would be reasonable as most devices will be RVC2s.

bindings/python/src/modelzoo/NNModelDescriptionBindings.cpp
lnotspotl: I initially though that `depthai` wanted to have as much 1:1 looking codebase, both in C++ and python. I agree with you that such a constructor with keyword arguments in python will definitely make working with the class a bit more convenient. Added

PR #1034, opened 2024-06-06
V3 human pose

LanSevcnikar:
I am worried that alongside the commits of things that needed to be added (primarily the human_pose_estimation.cpp, the blob for the model, and makeList.cmake), I also added a few things that were created when I built my code for the first time. In that case, only merge the files needed, the build process in respect to git practices is new to me


moratom:
You can use this as an example on how to make an app out of `depthai-core` repository.

https://github.com/luxonis/depthai-core-example

PR #1053, opened 2024-07-01
Develop zerocopy

TheMutta:
This update leverages the new FD protocol in XLink to avoid unecessary copying, allowing for a zerocopy runtime on localhost.

This is allowed by the new memory class, SharedMemory, which instread of containing the data itself, like VectorMemory, it only holds the FD and the mapping.


include/depthai/utility/SharedMemory.hpp
```diff
@@ -0,0 +1,93 @@
...
+	    return;
+	}
+
+	munmap(mapping, getSize());
+    }
+   public:
+    SharedMemory() {
+	kind = MemoryKinds::MEMORY_KIND_SHARED_MEMORY;
+	fd = -1;
+    }
```
moratom, line 43: Does it make sense to allow this?

I think it would make more sense to only allow to pass in a size and allocate the shared memory.

include/depthai/utility/SharedMemory.hpp
```diff
@@ -0,0 +1,93 @@
...
+	fstat(fd, &fileStats);
+
+        return fileStats.st_size;
+    }
+    std::size_t getOffset() const override {
+        return ftell(fdopen(fd, "r"));
+    }
+    void setSize(std::size_t size) override {
+	ftruncate(fd, size);
+    }
```
moratom, line 82: What if size is bigger than current size?

include/depthai/utility/VectorMemory.hpp
```diff
@@ -14,9 +14,9 @@ namespace dai {
 class VectorMemory : public std::vector<std::uint8_t>, public Memory {
    public:
     // using std::vector<std::uint8_t>::vector;
-    VectorMemory() = default;
-    VectorMemory(const std::vector<std::uint8_t>& d) : vector(std::move(d)) {}
-    VectorMemory(std::vector<std::uint8_t>&& d) : vector(std::move(d)) {}
+    VectorMemory() { kind = MemoryKinds::MEMORY_KIND_VECTOR_MEMORY; }
+    VectorMemory(const std::vector<std::uint8_t>& d) : vector(std::move(d)) { kind = MemoryKinds::MEMORY_KIND_VECTOR_MEMORY; }
+    VectorMemory(std::vector<std::uint8_t>&& d) : vector(std::move(d)) { kind = MemoryKinds::MEMORY_KIND_VECTOR_MEMORY; }
```
moratom, line 19: We can remove the kind I think and use casts instead.

moratom (commented):
Thanks Filippo!

I left a small nitpicks.
You can run clangformat on the code with `cmake --build build --target clangformat`.

* I think we should extend `SharedMemory` so it can be allocated on core side (conditionally for linux builds).
* And to extend `writingThread` in `DataQueue.cpp` to send out frames with shared memory when possible.

To allow for fast transfer in both directions.

include/depthai/utility/SharedMemory.hpp
TheMutta: Well, technically ftruncate extends it, but how that would interact with the various kinds of shared memory depends on the implementation

If it's for example memory created with shm_open, it will just expand it. If it's a DMA buffer, it depends on the implementation. For Gst, it seems to be the same as shm_open. 
This behavior should be overridden by implementations if it's different.

However, we probably need to unmap and map the FD so that we can actually have a mapping of the correct size.

include/depthai/utility/SharedMemory.hpp
TheMutta: Yeah probably useless, i removed it.

> I think it would make more sense to only allow to pass in a size and allocate the shared memory.

Well, yes, that makes sense. So both passing the FD externally or automatically allocating with a shared memory buffer.

Also, i think it would be worth it looking into [`memfd_create()`](https://man7.org/linux/man-pages/man2/memfd_create.2.html), since it is automatically closed when all references are released, which would be better than `shm_open`, which must be unlinked.

include/depthai/utility/VectorMemory.hpp
TheMutta: I think the kind is important, as for example in XLinkOut in DepthAI device, we need it to discern how to output. Unless we want to understand which it is by using sizeof(), which is imperfect, this system works better.

include/depthai/pipeline/datatype/ADatatype.hpp
```diff
@@ -6,6 +6,8 @@
 #include "depthai/utility/Memory.hpp"
 #include "depthai/utility/Serialization.hpp"
 #include "depthai/utility/VectorMemory.hpp"
+#include "depthai/utility/SharedMemory.hpp"
```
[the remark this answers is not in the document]
TheMutta: Yes, removed

include/depthai/utility/SharedMemory.hpp
moratom: Makes sense, I though truncate only downsizes.

include/depthai/utility/VectorMemory.hpp
moratom: Can we use `if(std::dynamic_ptr_cast<SharedMemory>(data))`?

include/depthai/utility/SharedMemory.hpp
moratom: Please look into `memfd_create()` in that case so we can send data zero copy in the other direction as well then yeah.

include/depthai/utility/VectorMemory.hpp
TheMutta: I didn't think about that...
That should work yes. I've tested it on a separate program and it works. I'm not sure about overhead, but it should be comparable anyway.

include/depthai/utility/SharedMemory.hpp
TheMutta: It is pretty nice. It's like `shm_open()`, but better. You can do sealing and it has no need for unlinking.
The only issue would be name generation, which is the same for `shm_open()`.
I'd wager we could use a static class variable to create a string like "shared_memory_" and then add the number of instances created. I think that should last us up until 2^64 instances :)

TheMutta: Oh btw, that constructor is actually needed because of the `GstBufferMemory` constructor in DepthAI device, as we set the FD in the constructor itself.

moratom: If we set it in the constructor, can we just initialize shared memory in the initializer list?

`GstBufferMemory(int fd): SharedMemory(fd) {}...`

moratom: Oh I see what you mean (checked the source).

Let's leave it then.

include/depthai/utility/VectorMemory.hpp
TheMutta: Seems like it doesn't work. I've reinstated the kind variables and they work now.

TheMutta:
Branch is ready for merge

lnotspotl:
There is a missing `fd` at the end of this [line](https://github.com/luxonis/depthai-core/blob/917e5ad8ab2f3de351e7880c467133152b164ba7/src/pipeline/datatype/StreamMessageParser.cpp#L216)

PR #1055, opened 2024-07-03
Add onStart, onStep , createInput and createOutput methods

lnotspotl:
This PR aims to solve two clickup tasks at the same time. 

Firstly, I added two methods, namely `onStart` and `onStop`, called in `ThreadedNode`'s `start` and `stop` methods respectivelly. It is two virtual functions so users can override these and do some useful work before starting the main thread or directly after stopping it (e.g. storing some variables to a file). Unless overriden, the two functions do nothing = it's just `{}`. The task on clickup is [here](https://app.clickup.com/t/86bymcrzd)

The second task was all about adding `createInput` and `createOutput` methods to the `ThreadedNode` class. As the `ThreadedNode` keeps pointers to all its inputs and outputs, it was necessary to make sure a created `Input` or `Output` does not get deallocated, resulting in a dangling pointer. To achieve that, `ThreadedNode` stores a vector of shared pointers to the created `Inputs` and `Outputs`. This ensures that the there's never a dangling pointer as long as a `Node` is alive (it uses these pointers internally)

The second task on clickup is [here](https://app.clickup.com/t/86bymavt6)


include/depthai/pipeline/ThreadedNode.hpp
```diff
@@ -17,11 +17,33 @@ class ThreadedNode : public Node {
     AtomicBool running{false};
     static inline std::shared_ptr<spdlog::details::thread_pool> threadPool = std::make_shared<spdlog::details::thread_pool>(8192, 1);
 
+    // Keep track of created inputs and outputs - we need to make sure they are kept
+    // alive so that `getInputRefs` and `getOutputRefs` don't use dangling Input/Output
+    // pointers
+    std::vector<std::shared_ptr<Input>> createdInputs;
+    std::vector<std::shared_ptr<Output>> createdOutputs;
+
```
moratom, line 25: Let's move this just to the python host nodes, not to all nodes.

src/pipeline/ThreadedNode.cpp
```diff
@@ -61,6 +63,16 @@ void ThreadedNode::stop() {
...
 
+std::shared_ptr<Node::Input> ThreadedNode::createInput() {
+    createdInputs.push_back(std::make_shared<Input>(*this, InputDescription{}));
+    return createdInputs.back();
+}
+
+std::shared_ptr<Node::Output> ThreadedNode::createOutput() {
+    createdOutputs.push_back(std::make_shared<Output>(*this, OutputDescription{}));
+    return createdOutputs.back();
+}
```
moratom, line 74: Accept descriptions structs as arguments and bind it as args in python, similar to how constructor is done now.

moratom (commented):
Let's just move the function to be python only, otherwise LGTM

bindings/python/src/pipeline/node/NodeBindings.cpp
```diff
@@ -405,5 +404,34 @@ void NodeBindings::bind(pybind11::module& m, void* pCallstack) {
...
+        .def(
+            "createOutput",
+            [](ThreadedNode& node, std::string name, std::string group, std::vector<Node::DatatypeHierarchy> types) {
+                static std::vector<std::shared_ptr<Node::Output>> createdOutputs;
+                createdOutputs.push_back(std::make_shared<Node::Output>(node, Node::OutputDescription{name, group, types}));
+                return createdOutputs.back();
+            },
+            py::arg("name") = Node::OutputDescription{}.name,
+            py::arg("group") = Node::OutputDescription{}.group,
+            py::arg("possibleDatatypes") = Node::OutputDescription{}.types);
```
moratom, line 436: This can be done in a similar manner than current binded constructors with keep_alive and moved to HostNodeBindings

lnotspotl: `keep_alive<2,1>()` is better, cleaner. :+1:

include/depthai/pipeline/ThreadedNode.hpp
moratom: Still needed?

bindings/python/src/pipeline/node/NodeBindings.cpp
```diff
@@ -405,5 +404,32 @@ void NodeBindings::bind(pybind11::module& m, void* pCallstack) {
...
+            py::keep_alive<2, 1>())
+        .def(
+            "createOutput",
+            [](ThreadedNode& node, std::string name, std::string group, std::vector<Node::DatatypeHierarchy> types) {
+                return std::make_unique<Node::Output>(node, Node::OutputDescription{name, group, types});
+            },
+            py::arg("name") = Node::OutputDescription{}.name,
+            py::arg("group") = Node::OutputDescription{}.group,
+            py::arg("possibleDatatypes") = Node::OutputDescription{}.types,
+            py::keep_alive<2, 1>());
```
moratom, line 434: I think it should be py::keep_alive<1,0>, but in any case let's add a test for it.

lnotspotl: I think you are right, the output (0) should be kept alive at least as long as arg 1 (parent).  In that case, isn't there a bug in the bindings for the constructor for `Input` and `Output`?

include/depthai/pipeline/ThreadedNode.hpp
lnotspotl: Removed.

PR #1068, opened 2024-07-19
RVC2: Add v3 camera node example

SzabolcsGergely:


moratom:
Let's go, nice!

I think we should move the `Camera` folder to the "root" examples directory, since it's not platform dependent anymore.

moratom (approved):
Thanks @SzabolcsGergely, LGTM!

PR #1074, opened 2024-07-24
Remove build, update exampled, replace build with buildInternal

lnotspotl:
Q: Do we want to remove `build` completely or deprecate it and print out a warning if it is used?

Q: I don't really like the fact that `DeviceNode`'s `buildInternal` virtual method is public. Is there a better way to go about this?


bindings/python/src/pipeline/node/CameraBindings.cpp
```diff
@@ -28,14 +28,14 @@ void bind_camera(pybind11::module& m, void* pCallstack) {
...
-                self.build();
+                self.buildInternal();
                 CAMERA_CODE(.)
                 return std::static_pointer_cast<Camera>(self.shared_from_this());
             },
             CAMERA_PYARGS)
         .def(py::init([](CAMERA_ARGS) {
                  auto self = getImplicitPipeline()->create<Camera>();
-                 self->build();
+                 self->buildInternal();
```
moratom, line 38: isn't called by `create` already?

bindings/python/src/pipeline/node/ColorCameraBindings.cpp
```diff
@@ -98,8 +98,12 @@ void bind_colorcamera(pybind11::module& m, void* pCallstack){
     ;
 
     // ColorCamera node
-    colorCamera.def(py::init([]() { return getImplicitPipeline()->create<ColorCamera>()->build(); }))
-        .def("build", &ColorCamera::build, DOC(dai, node, ColorCamera, build))
+    colorCamera
+        .def(py::init([]() {
+            auto camera = getImplicitPipeline()->create<ColorCamera>();
+            camera->buildInternal();
```
moratom, line 104: Called by create?

moratom:
>Q: Do we want to remove build completely or deprecate it and print out a warning if it is used?
Let's remove it, we're still early in the alpha release, just write in the internal channels that we've done it.

>Q: I don't really like the fact that DeviceNode's buildInternal virtual method is public. Is there a better way to go about this?
What happens if made private?

bindings/python/src/pipeline/node/ColorCameraBindings.cpp
lnotspotl: Good catch.

bindings/python/src/pipeline/node/CameraBindings.cpp
lnotspotl: Good catch. :100:

lnotspotl:
@moratom Designating the `buildInternal` method as private would not help here. After the shared pointer `nodePtr` gets created, calling the `buildInternal` method is not possible due to it being private.

```C++
template <typename... Args>
[[nodiscard]] static std::shared_ptr<Derived> create(std::shared_ptr<Device> device, Args&&... args) {
    auto nodePtr = std::shared_ptr<Derived>(new Derived(device, std::forward<Args>(args)...));
    nodePtr->buildInternal();
    return nodePtr;
}
```

 This can be fixed by marking `DeviceNodeCRTP<Base, Derived, Prop>` as a friend for each of the classes (ImageManip, XLinkOut, SystemLogger, ...). I am not a big fan of this approach, as you would have two places with e.g. `DeviceNodeCRTP<DeviceNode, ImageManip, ImageManipConfig>`.
 
 ```C++
class SystemLogger : public DeviceNodeCRTP<DeviceNode, SystemLogger, SystemLoggerProperties> {

...
  protected:
    friend class DeviceNodeCRTP<DeviceNode, SystemLogger, SystemLoggerProperties>;
    void buildInternal();

...

};
```
 I don't see any other way of doing that. Does something come to your mind?

**Comment by moratom** [comment]
@lnotspotl let's keep it public for now then, not the cleanest but it was public in `rvc3_develop` already and it does have `Internal` in the name.

File: include/depthai/pipeline/node/DetectionNetwork.hpp [hunk 3]
```diff
@@ -173,15 +169,15 @@ class MobileNetDetectionNetwork : public DetectionNetwork {
...
 };
 
 /**
  * @brief YoloDetectionNetwork node. Parses Yolo results
  */
 class YoloDetectionNetwork : public DetectionNetwork {
    public:
-    std::shared_ptr<YoloDetectionNetwork> build();
     [[nodiscard]] static std::shared_ptr<YoloDetectionNetwork> create() {
         return std::make_shared<YoloDetectionNetwork>();
```
moratom, line 182: We need to be call it in the `create()` here.

Likewise for all the detection & spatial detection networks.

moratom:
We should check that the python examples don't call the .build function any more and then we can merge.

lnotspotl:
No `build()` function calls anywhere throughout the codebase. The only exception is the `Pipeline` class, but this PR is not about that, and we want to keep it the way it is anyway. This PR is ready.

moratom (approved):

PR #1078, opened 2024-07-29
Auto reconnect

borbrudar:
added automatic reconnect functionality

### DIFF
--- a/include/depthai/device/DeviceBase.hpp
+++ b/include/depthai/device/DeviceBase.hpp
@@ -48,6 +48,8 @@ class PipelineImpl;
  * The core of depthai device for RAII, connects to device and maintains watchdog, timesync, ...
  */
 class DeviceBase {
+    friend class PipelineImpl;  // Needed for reconnections
+
    public:
     // constants
 
@@ -893,9 +895,9 @@ class DeviceBase {
         return connection;
     }
     /**
-    * Sets max number of automatic reconnection attempts
-    * @param maxAttempts 
-    */
+     * Sets max number of automatic reconnection attempts
+     * @param maxAttempts
+     */
     void setMaxReconnectionAttempts(int maxAttempts);
 
    protected:
@@ -923,7 +925,6 @@ class DeviceBase {
      * @note Remember to call this function in the overload to setup the communication properly
      */
     virtual void closeImpl();
-
 
    protected:
     // protected functions
@@ -1002,9 +1003,8 @@ class DeviceBase {
     // Started pipeline
     std::optional<PipelineSchema> pipelineSchema;
 
-    // Reconnection attempts
-    int maxReconnectionAttempts=0;
-public:
+    // Reconnection attempts and pointer to reset connections
+    int maxReconnectionAttempts = 0;
     std::weak_ptr<PipelineImpl> pipeline_ptr;
 };
 }  // namespace dai
--- a/include/depthai/device/DeviceBase.hpp
+++ b/include/depthai/device/DeviceBase.hpp
@@ -43,12 +43,15 @@ namespace dai {
 
 // Forward declare Pipeline
 class Pipeline;
-
+class PipelineImpl;
 /**
  * The core of depthai device for RAII, connects to device and maintains watchdog, timesync, ...
  */
 class DeviceBase {
    public:
+    int maxReconnectionAttempts = 5;
+    void setMaxReconnectionAttempts(int maxReconnectionAttempts);
+
     // constants
 
     /// Default search time for constructors which discover devices
@@ -941,7 +944,7 @@ class DeviceBase {
 
    private:
     // private functions
-    void init2(Config cfg, const dai::Path& pathToMvcmd, bool hasPipeline);
+    void init2(Config cfg, const dai::Path& pathToMvcmd, bool hasPipeline, bool reconnect = 0);
     void tryGetDevice();
 
     DeviceInfo deviceInfo = {};
@@ -995,5 +998,8 @@ class DeviceBase {
 
     // Started pipeline
     std::optional<PipelineSchema> pipelineSchema;
+
+public:
+    std::weak_ptr<PipelineImpl> pipeline_ptr;
 };
 }  // namespace dai
--- a/include/depthai/device/DeviceBase.hpp
+++ b/include/depthai/device/DeviceBase.hpp
@@ -950,7 +950,13 @@ class DeviceBase {
     // private functions
     void init2(Config cfg, const dai::Path& pathToMvcmd, bool hasPipeline, bool reconnect = 0);
     void tryGetDevice();
-
+    struct PrevInfo {
+        DeviceInfo deviceInfo;
+        Config cfg;
+        dai::Path pathToMvcmd;
+        bool hasPipeline;
+    };
+    void monitorCallback(std::chrono::milliseconds watchdogTimeout, PrevInfo prev);
     DeviceInfo deviceInfo = {};
     std::optional<Version> bootloaderVersion;
 
--- a/include/depthai/device/DeviceBase.hpp
+++ b/include/depthai/device/DeviceBase.hpp
@@ -49,9 +49,6 @@ class PipelineImpl;
  */
 class DeviceBase {
    public:
-    int maxReconnectionAttempts = 5;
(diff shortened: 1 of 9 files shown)


include/depthai/device/DeviceBase.hpp
```diff
@@ -995,5 +1001,10 @@ class DeviceBase {
 
     // Started pipeline
     std::optional<PipelineSchema> pipelineSchema;
+
+    // Reconnection attempts
+    int maxReconnectionAttempts=0;
+public:
+    std::weak_ptr<PipelineImpl> pipeline_ptr;
```
moratom, line 1008: does it need to be public?

include/depthai/pipeline/Pipeline.hpp
```diff
@@ -202,7 +202,9 @@ class PipelineImpl : public std::enable_shared_from_this<PipelineImpl> {
     void wait();
     void stop();
     void run();
-
+public:
+    void resetConnections();
```
moratom, line 206: needs to be public? Can we make `Device` a friend class instead?

src/device/DeviceBase.cpp
```diff
@@ -573,7 +574,7 @@ void DeviceBase::closeImpl() {
     // keep the shared_ptr valid (even if closed). Otherwise leads to using null pointers,
     // invalid memory, etc. which hard crashes main app
     connection->close();
-
+
```
moratom, line 577: Let's run clangformat to avoid whitespace changes.

`cmake --build build --target clangformat`

src/device/DeviceBase.cpp
```diff
@@ -816,7 +844,8 @@ void DeviceBase::init2(Config cfg, const dai::Path& pathToMvcmd, bool hasPipelin
 
         // Boot and connect with XLinkConnection constructor
         connection = std::make_shared<XLinkConnection>(deviceInfo, fwWithConfig, expectedBootState);
-
+        if(connection != nullptr) std::cout<<"connection created 1\n";
+        else std::cout<<"connection is null 1\n";
```
moratom, line 848: Remove debug info (or even better use a logger and keep it :))

src/pipeline/Pipeline.cpp
```diff
@@ -851,6 +849,28 @@ void PipelineImpl::start() {
...
+    // Add pointer to the pipeline to the device
+    std::shared_ptr<PipelineImpl> shared = shared_from_this();
+    const auto weak = std::weak_ptr<PipelineImpl>(shared);
+    defaultDevice->pipeline_ptr = weak;
+}
+
+void PipelineImpl::resetConnections(){
+    // reset connection on all nod<es
+    for(auto node : getAllNodes()){
+        if(defaultDevice->getConnection() == nullptr) continue;
```
moratom, line 862: why is this in the for loop?

src/device/DeviceBase.cpp
```diff
@@ -936,8 +968,54 @@ void DeviceBase::init2(Config cfg, const dai::Path& pathToMvcmd, bool hasPipelin
...
+                    }
+                }
+                break;
+            }
+            //std::cout<<"Reconnection successful, resetting connections on pipeline\n";
+            auto shared= pipeline_ptr.lock();
+            shared->resetConnections();
+            //std::cout<<"\"Connections reset, restarting monitor thread\"\n";
+            std::cout<<"Reconnection successful\n";
+            goto startMonitor;
```
moratom, line 1015: Can we just pot everything in a loop instead of a goto?
Or is goto explicitly needed here?

src/device/DeviceBase.cpp
```diff
@@ -682,10 +688,25 @@ void DeviceBase::init(Config config, UsbSpeed maxUsbSpeed, const dai::Path& path
...
+        dai::Path pathToMvcmd;
+        bool hasPipeline, loggingRunning, profilingRunning,timesyncRunning;
+    } prev;
+    prev.deviceInfo = deviceInfo;
+    prev.cfg = cfg;
+    prev.pathToMvcmd = dai::Path(pathToMvcmd);
+    prev.hasPipeline = hasPipeline;
+    prev.loggingRunning = loggingRunning;
+    prev.profilingRunning = profilingRunning;
+    prev.timesyncRunning = timesyncRunning;
```
moratom, line 708: do we need to save those?

moratom (commented):
Thanks Bor!

Generally looks good!
Let's clean up the debug logs (put them to a logger).

Also have we tested if this works when a NN model is used?

src/device/DeviceBase.cpp
```diff
@@ -936,8 +968,54 @@ void DeviceBase::init2(Config cfg, const dai::Path& pathToMvcmd, bool hasPipelin
...
+            if(timesyncThread.joinable()) timesyncThread.join();
+            if(loggingThread.joinable()) loggingThread.join();
+            if(profilingThread.joinable()) profilingThread.join();
+            pimpl->rpcStream = nullptr;
+            pimpl->rpcClient = nullptr;
+            if(!connection->isClosed()) connection->close();
+            connection = nullptr;
+            std::cout<<"Closed connection\nAttempting to reconnect in 1s\n";
+            deviceInfo = prev.deviceInfo;
+            sleep(1);
```
moratom, line 987: Let's expose this as an ENV variable, so the timings here can be overridden.

src/device/DeviceBase.cpp
borbrudar: Put into a while loop.

src/device/DeviceBase.cpp
borbrudar: Removed.

src/device/DeviceBase.cpp
borbrudar: Done.

borbrudar:
Put debug logs into the logger, tested NNs with gaze estimation, which I recently ported. Curiously enough the sync node version of the same experiment doesn't work, but it's functionally different, so there's probably an unrelated bug somewhere.

src/pipeline/Pipeline.cpp
borbrudar: good point, moved it outside the loop

include/depthai/pipeline/Pipeline.hpp
borbrudar: Making `Device` friend doesn't compile, but we can make `DeviceBase` a friend class, which is what I've done.

src/device/DeviceBase.cpp
borbrudar: Exposed as DEPTHAI_RECONNECT_TIMEOUT (value in seconds).

include/depthai/device/DeviceBase.hpp
borbrudar: Made it private and added `PipelineImpl` as a friend class.

include/depthai/pipeline/Pipeline.hpp
moratom: Let's do that yeah.

src/pipeline/Pipeline.cpp
moratom: generally, what is the expected scenario where this returns a nullptr?


If it's unexpected and/or an error a throw might be better?

borbrudar: It'd happen if there's a disconnect between the reconnect and pipeline connection reset. Throwing seems more reasonable I suppose.

src/device/DeviceBase.cpp
```diff
@@ -816,7 +840,6 @@ void DeviceBase::init2(Config cfg, const dai::Path& pathToMvcmd, bool hasPipelin
 
         // Boot and connect with XLinkConnection constructor
         connection = std::make_shared<XLinkConnection>(deviceInfo, fwWithConfig, expectedBootState);
-
     } else if(deviceInfo.state == X_LINK_BOOTED) {
```
moratom, line 843: Nitpick: best to remove whitespace changes

borbrudar: Noted.

src/pipeline/Pipeline.cpp
borbrudar: Done.

PR #1081, opened 2024-07-30
Auto reconnect #2

borbrudar:
fixed unwanted reconnect


src/device/DeviceBase.cpp
```diff
@@ -1074,6 +1072,92 @@ void DeviceBase::init2(Config cfg, const dai::Path& pathToMvcmd, bool hasPipelin
...
+                // Try parsing the string as a number
+                try {
+                    reconnectTimeout = std::stoi(timeoutStr);
+                } catch(const std::invalid_argument& e) {
+                    pimpl->logger.warn("DEPTHAI_RECONNECT_TIMEOUT value invalid: {}, should be a number (non-zero to enable)", e.what());
+                }
+            }
+            pimpl->logger.warn("Closed connection\nAttempting to reconnect in {}s\n", reconnectTimeout);
+            deviceInfo = prev.deviceInfo;
+            sleep(reconnectTimeout);
```
moratom, line 1127: Let's change to a float and the logic to keep trying for the set amount of time instead of sleeping and then only trying once.

So if someone sets a large timeout he won't always have to wait as long, just wait as long in the worst case.

themarpe: `reconnectTimeout` should become std::chrono::duration<float> and sleep should become `std::this_thread::sleep_for(reconnectTimeout)`

This encapsulate time in correct units as well + portable et al

src/device/DeviceBase.cpp
```diff
@@ -778,6 +791,13 @@ void DeviceBase::init2(Config cfg, const dai::Path& pathToMvcmd, bool hasPipelin
         pimpl->logger.debug("Device - BoardConfig: {} \nlibnop:{}", jBoardConfig.dump(), spdlog::to_hex(utility::serialize(config.board)));
     }
 
+    // Check if the device actually exists
+    try {
+        tryGetDevice();
+    } catch(std::exception&) {
+        throw;
+    }
+
```
moratom, line 800: Move out of the `init2` function and only call this part on the reconnect event.

There use `getAnyAvailableDevice` with a custom timeout that's configurable through the env variable and is specific for the reconnect event

src/device/DeviceBase.cpp
borbrudar: Done.

src/device/DeviceBase.cpp
borbrudar: Removed.

moratom:
@themarpe thoughts on what should be default?

I think reconnect by default makes sense, it would be breaking behavior on `v2`, but might worth taking the opportunity of the new major version and reconnecting by default?

It would make a lot of rare bugs less painful and one can still disable it explicitly and/or provide a callback to catch the re-connection events in case the application is doing something critical where disconnects have to be logged.

themarpe:
@moratom default on LGTM - note, we still target this happening close to never or never (should be logged as "crash" from report perspective)

PR #1082, opened 2024-07-31
Python type mismatch

borbrudar:
Added an extra argument to `addTensor` to allow users to specify the underlying tensor data type.


moratom (commented):
I think this will still behave a bit unexpected due to how C++ API works.

The int version I believe gets collapsed to uint8_t for example.
 
Let's add tests (see nndata_tensor_test.py).
To catch all various edge cases.

borbrudar:
True, but that's due to the internals of how `addTensor` works, specifically:

``` // Convert data to u8 or fp16 and write to data
        if(std::is_integral<_Ty>::value) {
            for(uint32_t i = 0; i < tensor.size(); i++) {
                vecData->data()[i + offset] = (uint8_t)tensor.data()[i];
            }
        } else {
            for(uint32_t i = 0; i < tensor.size(); i++) {
                *(uint16_t*)(&vecData->data()[2 * i + offset]) = fp32_to_fp16(tensor.data()[i]);
            }
        }

Tests from `nndata_tensor_test.py` seem to pass with the modified signatures, although the optional parameter doesn't appear to be optional.

File: include/depthai/pipeline/datatype/NNData.hpp [hunk 1]
```diff
@@ -223,7 +265,10 @@ class NNData : public Buffer {
         auto vecData = std::dynamic_pointer_cast<VectorMemory>(data);
 
         // Get size in bytes of the converted tensor data, u8 for integral and fp16 for floating point
-        const size_t sConvertedData = std::is_integral<_Ty>::value ? tensor.size() : 2 * tensor.size();
+        //const size_t sConvertedData = std::is_integral<_Ty>::value ? tensor.size() : 2 * tensor.size();
+        size_t sConvertedData = tensor.size();
+        if(dataType == dai::TensorInfo::DataType::FP16) sConvertedData *= 2;
+        else if(dataType == dai::TensorInfo::DataType::FP32 || dataType == dai::TensorInfo::DataType::INT) sConvertedData *= 4;
```
moratom, line 271: Let's do a switch here and cover all cases, so we visually see that nothing is missing

moratom (changes requested):
Let's remove all of the specialized functions and rather have it as such:
* By default we try to infer the type from the tensor
* We allow to override the type with a function argument

We can do so both in C++ and Python I believe.

And please add tests as well.

include/depthai/pipeline/datatype/NNData.hpp
borbrudar: amended

borbrudar:
Reworked, however there are limitations to type conversions from python to c++ and vice versa, e.g. fp16 is not supported in pybind, so we have to cast it to a different type before we can pass it to c++ (i've chosen to upcast it to fp32). Similarly fp64 is not supported in c++ so i downcast it to fp32.

include/depthai/pipeline/datatype/NNData.hpp
```diff
@@ -518,7 +617,7 @@ class NNData : public Buffer {
 
         return {};
     }
-#endif
+//#endif
```
moratom, line 620: Intentional?

include/depthai/pipeline/datatype/NNData.hpp
```diff
@@ -198,7 +198,7 @@ class NNData : public Buffer {
      */
     span<std::uint8_t> emplaceTensor(TensorInfo& tensor);
 
-#ifdef DEPTHAI_XTENSOR_SUPPORT
+//#ifdef DEPTHAI_XTENSOR_SUPPORT
```
moratom, line 201: I think this is still needed?

include/depthai/pipeline/datatype/NNData.hpp
```diff
@@ -208,8 +208,47 @@ class NNData : public Buffer {
...
+    };
+    NNData& addTensor(const std::string& name, const xt::xarray<double>& tensor) {
+        return addTensor<double>(name, tensor, dai::TensorInfo::DataType::FP64);
+    };
+    NNData& addTensor(const std::string& name, const xt::xarray<std::int8_t>& tensor) {
+        return addTensor<std::int8_t>(name, tensor, dai::TensorInfo::DataType::I8);
+    };
+    NNData& addTensor(const std::string& name, const xt::xarray<std::uint8_t>& tensor) {
+        return addTensor<std::uint8_t>(name, tensor, dai::TensorInfo::DataType::U8F);
     };
```
moratom, line 252: Let's add tests for these too

moratom (changes requested):

include/depthai/pipeline/datatype/NNData.hpp
borbrudar: ;)

include/depthai/pipeline/datatype/NNData.hpp
borbrudar: Added.

PR #1086, opened 2024-08-01
Experimental/device subnodes ref

lnotspotl:


include/depthai/pipeline/Subnode.hpp
```diff
@@ -0,0 +1,45 @@
...
+            auto device = dynamic_cast<DeviceNode*>(&parent)->getDevice();
+            std::dynamic_pointer_cast<DeviceNode>(node)->setDevice(device);
+        }
+
+        // Now that node is created, call buildInternal
+        node->buildInternal();
+
+        // Add reference
+        parent.nodeRefs.push_back(&node);
+    }
```
moratom, line 36: Note that this will crash if configure mode is on.

`node->buildInternal` will be on a nullptr

lnotspotl: Fixed.

include/depthai/pipeline/Subnode.hpp
```diff
@@ -0,0 +1,45 @@
...
+class Subnode {
+    std::shared_ptr<Node> node;
+
+   public:
+    Subnode(Node& parent, std::string alias) {
+        if(!parent.configureMode) {
+            // Create node as well
+            node = std::make_shared<T>();
+            node->setAlias(alias);
+            node->parentNode = node;
```
lnotspotl, line 20: This is just a copy-past from the original `Subnode` and the parentNode is not being set correctly here.

Q: Do we still use the parentNode somewhere else other than `getPipelineSchema`? What is a parentNode anyway? Can't we replace `parentNode` with `parentNodeId` for it's use in `getPipelineSchema`?

Note: We can't do  `shared_from_this()` as it throws a `bad_weak_ptr` error.

moratom: I think this breaks the `getAllNodes()` function though, which does get used.

include/depthai/pipeline/node/NeuralNetwork.hpp
```diff
@@ -80,6 +81,14 @@ class NeuralNetwork : public DeviceNodeCRTP<DeviceNode, NeuralNetwork, NeuralNet
      */
     void setNNArchive(const NNArchive& nnArchive, int numShaves);
 
+    /**
+     * @brief Download model from zoo and set it for this Node
+     *
+     * @param description: Model description to download
+     * @param useCached: Use cached model if available
+     */
+    void setFromModelzoo(NNModelDescription description, bool useCached = true);
```
moratom, line 90: setFromModelZoo

include/depthai/pipeline/node/DetectionParser.hpp
```diff
@@ -60,12 +61,12 @@ class DetectionParser : public DeviceNodeCRTP<DeviceNode, DetectionParser, Detec
...
-     * @brief Set NNArchive for this Node, throws if the archive's type is not SUPERBLOB
+     * @brief Download model from zoo and set it for this Node
      *
-     * @param nnArchive: NNArchive to set
-     * @param numShaves: Number of shaves to use
+     * @param description: Model description to download
+     * @param useCached: Use cached model if available
      */
-    void setNNArchive(const NNArchive& nnArchive, int numShaves);
+    void setFromModelzoo(NNModelDescription description, bool useCached = true);
```
moratom, line 70: Let's remove this function from the `DetectionParser` unlikely to ever be used.

Instead the `DetectionNetwork` can pass in the `NNArchive` directly.

include/depthai/pipeline/DeviceNode.hpp
```diff
@@ -23,6 +23,20 @@ class DeviceNode : public ThreadedNode {
         return false;
     }
 
+    /**
+     * @brief Set device for this node
+     *
+     * @param device: shared pointer to device
+     */
+    void setDevice(std::shared_ptr<Device> device);
+
```
moratom, line 32: Can we make it private and the class that uses it a friend?

lnotspotl: Done. I think the `getDevice` method should be kept as it is, i.e. public. I can see applications for that.

src/pipeline/Pipeline.cpp
```diff
@@ -185,7 +186,7 @@ PipelineSchema PipelineImpl::getPipelineSchema(SerializationType type) const {
     // Loop over all nodes, and add them to schema
     for(const auto& node : getAllNodes()) {
         // const auto& node = kv.second;
-        if(std::string(node->getName()) == std::string("NodeGroup")) {
+        if(std::string(node->getName()) == std::string("NodeGroup") || std::string(node->getName()) == std::string("DeviceNodeGroup")) {
```
lnotspotl, line 189: Why don't we add a virtual method `isDummyNode()` returning true for all "normal" nodes and false for these dummy nodes such as `DetectionNetwork`.

include/depthai/pipeline/node/DetectionParser.hpp
lnotspotl: Removed. Should be better now.

lnotspotl:
Github action: https://github.com/luxonis/depthai-core/actions/runs/10223566899

src/pipeline/Pipeline.cpp
moratom: We can yes.

IMO the cleanest is multiinheritance, similar to how we handle `SourceNode`s and we use `dynamic_cast` here.

src/pipeline/node/DetectionNetwork.cpp
```diff
@@ -22,24 +24,21 @@ namespace node {
...
+      input{neuralNetwork->input},
+      passthrough{neuralNetwork->passthrough} {};
 
 // -------------------------------------------------------------------
 // Neural Network API
 // -------------------------------------------------------------------
 
 void DetectionNetwork::buildInternal() {
+    // Call parent's buildInternal
+    DeviceNodeGroup::buildInternal();
```
moratom, line 40: This doesn't do anything right?

src/pipeline/DeviceNode.cpp
```diff
@@ -17,6 +17,14 @@ DeviceNode::DeviceNode(const std::shared_ptr<Device>& device, std::unique_ptr<Pr
     configureMode = conf;
 }
 
+void DeviceNode::setDevice(std::shared_ptr<Device> device) {
+    this->device = device;
+}
+
+const std::shared_ptr<Device> DeviceNode::getDevice() const {
```
moratom, line 24: What was the idea with the first `const` here?

The device pointer we have saved is not const.

moratom (commented):
Thanks Jakub!

I think this needs a few minor tweaks (left some comments) before the merge but it is the right approach.

src/pipeline/node/DetectionNetwork.cpp
lnotspotl: No, it does now. I removed it, it was an old piece of code.

src/nn_archive/NNArchive.cpp
```diff
@@ -17,91 +18,96 @@ NNArchive::NNArchive(const std::string& archivePath, NNArchiveOptions options) :
...
+    switch(modelType) {
+        case model::ModelType::OTHER:
+        case model::ModelType::DLC:
             return unpackedModelPath;
             break;
-        case NNArchiveType::BLOB:
-        case NNArchiveType::SUPERBLOB:
+        case model::ModelType::BLOB:
+        case model::ModelType::SUPERBLOB:
             return std::nullopt;
```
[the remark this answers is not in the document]
lnotspotl: Totally. It makes a lot more sense to return a path to the model regardless of it's type.

src/pipeline/DeviceNode.cpp
lnotspotl: The `const` specifier prevents you from calling `reset()` or anything. The user does have non-const access to the underlying Device object though.

PR #1089, opened 2024-08-05
Change initialConfig python bindings

SzabolcsGergely:


moratom (commented):
```
stereo.initialConfig.setConfidenceThreshold(200)
print("Confidence threshold is {}", stereo.initialConfig.getConfidenceThreshold())
```
returns the default value `245` not the one that has been set, as it's changing a copy.

Likely we'll need to change the `initialConfig`s to be managed.

PR #1092, opened 2024-08-09
Replace custom python implementations for cvFrame with bindings

moratom:
Added bindings instead of custom python implementations for the `getCvFrame` and `getFrame` functions.

Should be easier to maintain with a single source of truth.

Another thing we get for free is `setCvFrame` from python  - that one is quite neat :)

TODO before merge:
*  Test with numpy2
*  Add tests

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -10,3 +10,6 @@
 [submodule "bindings/python/external/xtensor-python"]
 	path = bindings/python/external/xtensor-python
 	url = https://github.com/xtensor-stack/xtensor-python.git
+[submodule "3rdparty/pybind11_opencv_numpy"]
+	path = bindings/python/external/pybind11_opencv_numpy
+	url = git@github.com:edmBernard/pybind11_opencv_numpy.git
--- a/bindings/python/src/pipeline/datatype/ImgFrameBindings.cpp
+++ b/bindings/python/src/pipeline/datatype/ImgFrameBindings.cpp
@@ -2,10 +2,9 @@
 #include "pipeline/CommonBindings.hpp"
 #include <unordered_map>
 #include <memory>
-
 // depthai
 #include "depthai/pipeline/datatype/ImgFrame.hpp"
-
+#include "ndarray_converter.h"
 //pybind
 #include <pybind11/chrono.h>
 #include <pybind11/numpy.h>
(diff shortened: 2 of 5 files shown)


.gitmodules
```diff
@@ -10,3 +10,6 @@
 [submodule "bindings/python/external/xtensor-python"]
 	path = bindings/python/external/xtensor-python
 	url = https://github.com/xtensor-stack/xtensor-python.git
+[submodule "3rdparty/pybind11_opencv_numpy"]
+	path = bindings/python/external/pybind11_opencv_numpy
+	url = git@github.com:edmBernard/pybind11_opencv_numpy.git
```
asahtik, line 15: Do ssh submodule links work when a user is not logged in to github? Maybe it'd be better to replace it with https?

themarpe:
Neat! Does `getFrame` remain zerocopy?

src/pipeline/datatype/ImgFrame.cpp
```diff
@@ -63,8 +63,11 @@ unsigned int ImgFrame::getCategory() const {
 unsigned int ImgFrame::getWidth() const {
     return fb.width;
 }
+
 unsigned int ImgFrame::getStride() const {
-    if(fb.stride == 0) return getWidth();
+    if(fb.stride == 0) {
+        return static_cast<unsigned>(std::round(static_cast<float>(getWidth()) * getBytesPerPixel()));
```
asahtik, line 69: Does this work for YUV type frames (NV12)? If I remember correctly getBytesPerPixel returns 1.5, even though the actual stride is equal to width (if there's no alignment).

asahtik (approved):

.gitmodules
moratom: Good catch, thanks, will replace.

src/pipeline/datatype/ImgFrame.cpp
moratom: I returns `1` for NV12, but definitely still problematic, since stride is really per plane in NV12.

I don't think it's worse than before though.

moratom:
> Neat! Does `getFrame` remain zerocopy?
I don't think so, we can revisit though, the conversions seems of "manageable complexity" - 350 LOC.

Existing zero-copy is unsafe, so it would be nice if we somehow managed to tie the lifetimes now.
Right now this crashes:
```
imgFrame = queue.get()
myNpFrame = imgFrame.getFrame()
imgFrame = None
myPixel = myNpFrame[3][5] #  A likely crash as memory was already released :(
```

src/pipeline/datatype/ImgFrame.cpp
asahtik: Yeah sorry I re-checked and it does return 1. The way we store strides might be an issue with yuv420 though, but this is not in the scope of this PR. (the stride of the 2nd and 3rd planes is different than the 1st plane)

bindings/python/src/pipeline/datatype/ImgFrameBindings.cpp
```diff
@@ -138,215 +139,34 @@ void bind_imgframe(pybind11::module& m, void* pCallstack){
...
-                    break;
-            }
-
-            // Default case
-            return frame.attr("copy")();
-
-        }, "Returns BGR or grayscale frame compatible with use in other opencv functions")
-
+#ifdef DEPTHAI_HAVE_OPENCV_SUPPORT
+        .def("getFrame", &ImgFrame::getFrame, py::arg("copy") = true, DOC(dai, ImgFrame, getFrame))
```
moratom, line 143: A copy likely happens twice here now, once by `cpp` impl and then once more on the opencv -> numpy conversion.
Likewise for `getCvFrame`.

TODO - fix before merge.

moratom: Done

PR #1094, opened 2024-08-11
Update to use latest xlink repo and added Clang support

onthegrid007:
Update to use latest xlink repo and added Clang support

### DIFF
--- a/.github/workflows/main.workflow.yml
+++ b/.github/workflows/main.workflow.yml
@@ -96,24 +96,24 @@ jobs:
     - name: Install dependencies
       if: matrix.os == 'macos-latest'
       run: |
-        brew install opencv
+        brew install opencv clang ninja
 
     - name: Install dependencies
       if: matrix.os == 'ubuntu-latest'
       run: |
         sudo apt-get update
-        sudo apt-get install libopencv-dev
+        sudo apt-get install libopencv-dev clang ninja-build
 
     - name: Install dependencies
       if: matrix.os == 'windows-latest'
       run: |
-        choco install opencv
+        choco install opencv llvm ninja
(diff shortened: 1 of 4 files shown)


moratom:
Thanks @onthegrid007 - any particular reason for the switch to ninja?

moratom:
CI fails on MacOS: https://github.com/luxonis/depthai-core/actions/runs/10337442673/job/28646392973

Let's either leave the CI builders as is or extend with a "matrix" to build with both GCC and Clang.

onthegrid007:
> Thanks @onthegrid007 - any particular reason for the switch to ninja?

Ninja is required for Clang on windows to overwrite the use of MSVS compiler in Cmake...

Thought it would be easier overall if everything just switched over at once.

onthegrid007:
Unfortunately unless I rebuild opencv in the ci I have to use clang with a target for mingw and mingw be the library target on windows, Ninja is still required for it not to default to MSVS, and since it's only a single binary for installation I'm not too worried...

Question, @moratom should I rebuild opencv cleanly in the ci so clang can work without mingw (see windows latest build for the error I'm mentioning) or is there a way like it kind of mentions to just specify opencv_dir and hope there's no binary discrepancies between the mingw prebuilt opencv binaries and clang compiling with msvs as the target libraries?

moratom:
@onthegrid007 we don't necessarily want to switch the CI to clang, rather we should extend it, so we run both with GCC and clang in the CI.

You can use the "matrix" for that in github actions.


Note that we'll be mainlining the updated XLink in this PR https://github.com/luxonis/depthai-core/pull/1098 and keep the this one to expend the CI to clang.

onthegrid007:
Is this good enough?

PR #1107, opened 2024-08-21
Updating getCameraExtrinsics

jakaskerl:
Update getCameraExtrinsics() so any two sockets can be used regardless of how they are linked. 

Calculation changed to:
1. src -> origin 
2. dst -> origin 
3. src->origin->dst

Due to some inaccuracies in calculations, the results might be slightly different than the old method. 
Tested for cyclic links as well. 

Excuse my cpp inexperience :)


moratom (commented):
Let's retarget this to v3_develop, please. v2 is now in maintenance mode.

jakaskerl:
#1142

PR #1109, opened 2024-08-22
Make Pipeline's run method catch signals

lnotspotl:
This PR  resolves the issue with the `Pipeline's` `run` method, where it would not catch signals and therefore make it impossible to stop some example programs using `Ctrl+C`.

The relevant clickup task is available [here](https://app.clickup.com/t/86bzv46rz)

Github [action](https://github.com/luxonis/depthai-core/actions/runs/10529898107)

Q: I've made all `PipelineImpl` methods public. Is there a better way to go about this?
Q: Instead of using the negative default value for the `timeoutSeconds` argument of the `processTasks` function, does it make sense to set it to `std::numeric_limits<double>::max()` by default, removing the need for the inner if statement?

### DIFF
--- a/bindings/python/src/pipeline/PipelineBindings.cpp
+++ b/bindings/python/src/pipeline/PipelineBindings.cpp
@@ -249,21 +249,21 @@ void PipelineBindings::bind(pybind11::module& m, void* pCallstack){
              [](Pipeline& p) {
                  {
                      py::gil_scoped_release release;
-                     p.impl()->start();
+                     p.start();
                  }
                  constexpr double timeoutSeconds = 0.1;
                  while(p.isRunning()) {
                      {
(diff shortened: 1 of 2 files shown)


moratom (commented):
Thanks Jakub!

Actually instead of making it public, can we use `Pipeline` object directly?

(I see I missed exposing the input variables to `processTasks` so we have to add the exist one and the new one you introduced.

moratom:
>Instead of using the negative default value for the timeoutSeconds argument of the processTasks function, does it make sense to set it to std::numeric_limits<double>::max() by default, removing the need for the inner if statement?

IMO no, while the max is probably so long that the universe will cool down before the function returns, the intent is to run for infinity.

PR #1114, opened 2024-08-24
Props, config and different image types

lnotspotl:


include/depthai/pipeline/datatype/AprilTagConfig.hpp
```diff
@@ -65,7 +65,7 @@ class AprilTagConfig : public Buffer {
      * lighting conditions or low light conditions.
      * The default value is 0.25.
      */
-    float decodeSharpening = 0.25f;
```
moratom, line 68: What's the reason for the change here? To match the underlying library?

src/pipeline/node/AprilTag.cpp
```diff
@@ -51,44 +56,102 @@ void AprilTag::buildInternal() {
...
         inFrame = inputImage.get<ImgFrame>();
-        // TODO: This only works for types that have a grayscale image at the beginning
-        uint8_t* src = inFrame->data->getData().data() + inFrame->fb.p1Offset;
-        auto width = static_cast<int32_t>(inFrame->getWidth());
-        auto height = static_cast<int32_t>(inFrame->getHeight());
-        auto stride = static_cast<int32_t>(inFrame->getStride());
-        image_u8_t aprilImg = {.width = width, .height = height, .stride = stride, .buf = src};
+
+        cv::Mat img;
+        cv::cvtColor(inFrame->getCvFrame(), img, cv::COLOR_BGR2GRAY);
```
moratom, line 152: That works, but it's quite inefficient, especially for very large frames.
For example for the case of NV12 it goes like this:
* getCvFrame creates a copy and converts NV12 to BGR888i to match opencv
* OpenCV does another copy to convert BGR888i to GRAY

Whilst we get the exact same results by just setting the pointer to the LUMA (grayscale) plane of the NV12.


Can we instead keep what we've had before for NV12 and GRAY (it was "view only").
Then I believe there is some other april tag datatype for color types that have three planes.
And then only for everything else we fallback to opencv (if depthai is compiled with opencv support).

src/pipeline/node/AprilTag.cpp
```diff
@@ -51,44 +56,102 @@ void AprilTag::buildInternal() {
...
+
+        image_u8_t aprilImg = {.width = img.cols, .height = img.rows, .stride = img.cols, .buf = img.data};
```
moratom, line 154: If we do the optimization above, we need to change stride back to stride in case `getCvFrame` is not used.

moratom (commented):
Thanks Jakub!

As I see it this is what's left:
* Disable apriltags for Windows builds
* Add back in the optimization to not use `getCvFrame` if possible whilst keeping the node working with all types.

include/depthai/pipeline/datatype/AprilTagConfig.hpp
lnotspotl: Yeah, the compiler was complaining about using a `float` as a `double`. Sure, we are sending a few more bits every message but in exchange we remove the compiler warnings. We could also perform casting. There must have been a reason why you had opted for floats, so please, let me know if you'd like to see the `float`s back.

themarpe:
> Internally, the detection algorithm does not care much about the pixel values. Instead, it produces its detections based on pixel intensities, i.e. how "shiny" a pixel appears to be. Therefore, in case of an RGB or BGR image, we can simply take one its channels. 

Due to "correctness" I think we'll have to convert to Grayscale. Picking just any R/G/B channel on its own, won't produce good results when tags are colored, tinted, not uniformly lit, etc...

moratom:
>Due to "correctness" I think we'll have to convert to Grayscale. Picking just any R/G/B channel on its own, won't produce good results when tags are colored, tinted, not uniformly lit, etc...

There is also `image_u8x3_t` in the AprilTag library which we can use for planar RGB types @lnotspotl

moratom (approved):

PR #1115, opened 2024-08-26
Develop Audio Nodes

TheMutta:
Adding audio managment capabilities to depthai with the creation of various audio nodes for audio capture, output, mixing, encoding and replay.

- AudioIn and AudioOut as endpoints for raw audio passing through alsa devices.
- AudioEncoder recieves data from one endpoint and converts it to some other format.
- AudioReplay, which plays a specific audio file on host.
- AudioMixer, tasked with multiplexing audio inputs and outputs


CMakeLists.txt
```diff
@@ -647,6 +654,9 @@ target_link_libraries(${TARGET_CORE_NAME}
     INTERFACE
         XLinkPublic
     PRIVATE
+        sndfile
+	samplerate
+	asound
```
asahtik, line 659: Please fix indentation

asahtik:
It'd probably be good to add comments to node / datatype getters and setters for docs generation.

Also might be good to run clangformat, but I'm not sure if `v3_develop` is properly formatted. If clangformat causes a lot of new changes, skip - we'll make a separate merge just for formatting.

moratom (commented):
Thanks Filippo!

Writing here a brief TODO list as a summary:
* Address comments from the PR
* Try to make all used libraries included by Hunter and/or make them optional so DAI can be compiled without audio support
* Regardless of the option chosen above, they should be included with cmake
* Make sure the PR builds successfully on all platforms in CI
* TBD on the AudioMixer node

PR #1117, opened 2024-09-02
Record & Replay V2

asahtik:


src/pipeline/node/Camera.cpp
```diff
@@ -178,6 +179,46 @@ NodeRecordParams Camera::getNodeRecordParams() const {
...
+            } else {
+                throw std::runtime_error("Unsupported size value");
+            }
+        }
+    }
+    if(maxWidth == 0 || maxHeight == 0) {
+        throw std::runtime_error("Invalid max width or height");
+    }
+    return std::make_pair(maxWidth, maxHeight);
+};
```
moratom, line 203: I think we should actually record maxSize, that **is** full FOV.

moratom (commented):
Thanks Sebastjan!

Let's add an example or two in python as well and move the cpp one to be cross platform, but generally LGTM.

examples/cpp/RVC2/Camera/rgb_video2.cpp
moratom, line 1: Let's remove before merge IMO

moratom (changes requested):

moratom:
Played around with it a bit, looks promising!

A TODO list from my side:
- [ ] Running the `holistic_record.py` example and the `holistic_replay.py` doesn't work out of the box, since the filename is different than the default (.tar.gz -> .tar)
- [ ] To add new env variables to README.md
- [ ] If you record with `camera_all.py` and then try to use holistic replay, you get:
  ```
  DEPTHAI_REPLAY=../Camera/recordings/recording.tar python detection_network.py 
  [2025-02-13 13:04:52.572] [warning] Replay disabled: Recording does not match the pipeline configuration.
  [2025-02-13 13:04:52.572] [depthai] [warning] Could not set up holistic replay. Record and replay disabled.
  ```
  whilst this should work IMO, the extra cameras shouldn't "bother" the example.
- [ ] Add tests
- [ ] The record&replay resolution should be picked as _the smallest nice full FoV that's still bigger than the requested outputs_.
       For example on IMX586 (4000x3000), if 300x300 is requested, 640x400 should be recorded, if 1000x1000 is requested 1920x1440...


In a later PR we will also need to mock the device side (.getConnectedCameraFeatures(), calibration, ...).

examples/cpp/CMakeLists.txt
moratom, line 1: Let's add the examples to the test suite IMO

src/opencv/HolisticRecordReplay.cpp
```diff
@@ -40,9 +58,57 @@ bool setupHolisticRecord(Pipeline& pipeline,
             outFilenames[nodeName] = filePath;
```
moratom, line 58: Should we maybe split this up into multiple functions? This has too many levels of indentation to digest easily

moratom (approved):

PR #1120, opened 2024-09-04
V3 thermal node

zrezke:
- Moved thermal to it's own node.
- Should be expanded later on to include runtime controls for tiny1c.
- And also it might be worth to create a ThermalFrame dtype with methods to convert between temperature units and get temp at X,Y.
- Right now you can get a FP16 and/or YUV422i ImgFrame from the temperature and color outputs respectively.


examples/python/RVC2/Thermal/thermal.py
moratom, line 1: Let's change to camelCase to keep examples consistent.

moratom (commented):
Thanks! I'd say we first implement the ThermalConfig on device side so it's correctly used and we can merge the PR.

examples/python/RVC2/Thermal/thermal.py
zrezke: Oh, sorry, I always forget we use camelCase here...

include/depthai/pipeline/datatype/ThermalConfig.hpp
```diff
@@ -0,0 +1,138 @@
...
+    std::optional<uint8_t> brightnessLevel;
+
+    /// Image contrast level, 0-255.
+    std::optional<uint8_t> contrastLevel;
+
+    /// Orientation of the image. Computed on the sensor.
+    std::optional<ThermalImageOrientation> orientation;
+    DEPTHAI_SERIALIZE(
+        ThermalImageParams, timeNoiseFilterLevel, spatialNoiseFilterLevel, digitalDetailEnhanceLevel, brightnessLevel, contrastLevel, orientation);
+};
```
moratom, line 108: Doesn't need to inherit from Buffer

zrezke:
This is ready now. @moratom

zrezke:
@moratom , I added the crosshair example and fixed the autoFFC not working + in theory scaled up the displayed images, but it seems like the opencv resizeWindow function doesn't work like it used to, reported the issue here: https://github.com/opencv/opencv/issues/26180

So for opencv-python<4.10.0.82 the image resize will work, for newer versions not

PR #1122, opened 2024-09-05
Update install_requirements so it pins numpu below version 2

Erol444:


examples/python/install_requirements.py
```diff
@@ -53,10 +53,7 @@ def hasWhitespace(string):
         requireOpenCv = True
 
 if requireOpenCv:
-    if args.install_rerun:
-        DEPENDENCIES.append('numpy==1.24.4') # rerun doesn't work with numpy 2.0 for now
-    else:
-        DEPENDENCIES.append('numpy')
+    DEPENDENCIES.append('numpy==1.24.4') # rerun doesn't work with numpy 2.0 for now
```
moratom, line 56: Let's change to <2 instead IMO.

Erol444:
LGTM

PR #1127, opened 2024-09-16
Refactor ImageManipV2 for RVC2 impl

asahtik:
\+ Some bugfixes


include/depthai/utility/ImageManipV2Impl.hpp
```diff
@@ -288,7 +291,7 @@ static inline void RGBfromYUV(float& R, float& G, float& B, float Y, float U, fl
 template <template <typename T> typename ImageManipBuffer, typename ImageManipData>
 bool ColorChange<ImageManipBuffer, ImageManipData>::colorConvertToRGB888p(
     const dai::span<const uint8_t> inputFrame, dai::span<uint8_t> outputFrame, FrameSpecs srcSpecs, FrameSpecs dstSpecs, dai::ImgFrame::Type from) {
-    dai::ImgFrame::Type to = dai::ImgFrame::Type::RGB888p;
+    // dai::ImgFrame::Type to = dai::ImgFrame::Type::RGB888p;
```
moratom, line 294: Why are these commented out now?

moratom (commented):
Thanks!

Can we add/move examples that are now cross platform I assume?

moratom (commented):
`.setBackgroundColor` has no effect on RVC2.


Let's address before merge.


Also I would split the ImageManipConfigV2 object API into operations and everything else that's not meant to be called more than once to avoid confusion.
Maybe we can only namespace them to ::operations?

PR #1128, opened 2024-09-19
Created keypoints parser node

NicikD:
Added keypoints parser device node. For now not working on device (only on host) and without python bindings. Also added keypoints parser example.


examples/cpp/parsers/keypoints_device.cpp
moratom, line 1: Let's move the example to RVC2 if it's expected to only work on RVC2

include/depthai/pipeline/datatype/creators/KeypointsCreator.hpp
```diff
@@ -0,0 +1,42 @@
...
+ * @param keypoints detected 2D keypoints
+ * @param scores confidence scores for each keypoint
+ * @param confidence_threshold confidence threshold
+ *
+ * @returns keypoints message
+ */
+std::shared_ptr<Keypoints> createKeypointsMessage(const std::vector<Point2f> keypoints);
+std::shared_ptr<Keypoints> createKeypointsMessage(const std::vector<Point2f> keypoints, const std::vector<float> scores, float confidence_threshold);
+
+}  // namespace dai
```
moratom, line 42: Can we rather just have different constructors?

include/depthai/pipeline/node/KeypointsParser.hpp
```diff
@@ -0,0 +1,97 @@
...
+
+    /**
+     * Set the number of keypoints.
+     */
+    void setNumKeypoints(int numKeypoints);
+    /**
+     * Get the number of keypoints.
+     */
+    int getNumKeypoints() const;
+
```
moratom, line 81: Is this apart of the model?

If so it might be better to just ingest the NNArchive itself in the parser and we can query it out from there.

src/pipeline/datatype/creators/KeypointsCreator.cpp
```diff
@@ -0,0 +1,67 @@
...
+    for (size_t i = 0; i < keypoints.size(); i++) {
+        points3d[i].x = keypoints[i].x;
+        points3d[i].y = keypoints[i].y;
+        points3d[i].z = 0;
+    }
+
+    return createKeypointsMessage(points3d, scores, confidence_threshold);
+}
+
+}  // namespace dai
```
moratom, line 67: IMO best to just have this in the constructors

src/pipeline/node/KeypointsParser.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+        msg->setTimestamp(inputData->getTimestamp());
+        msg->setTimestampDevice(inputData->getTimestampDevice());
+        msg->setSequenceNum(inputData->getSequenceNum());
+        out.send(msg);
+    }
+}
+
+}  // namespace node
+}  // namespace dai
```
moratom, line 85: Let's make it conditional to only have the node working if xtensor is available.

moratom (commented):
Thanks!

Let's add bindings, make sure it compiles/runs if xtensor is disabled and we'll add the node on Device then as well.

PR #1129, opened 2024-09-19
Add .build(nnArchive, config) to NeuralNetwork node 

borbrudar:
Added the function as well as python binding.


moratom:
LGTM when tested, thanks!

PR #1131, opened 2024-09-23
Image align

borbrudar:
Added ImageAlign class to v3.


moratom:
@borbrudar please add the bindings as well

PR #1133, opened 2024-09-23
Rename NNArchiveConfig to NNArchiveVersionedConfig, add NNArchive config getters

lnotspotl:
This PR renames the `NNArchiveConfig` class to `NNArchiveVersionedConfig` to reflect the fact that the class itself can hold many different `NNArchive` config versions (just `v1` at the time of creating this pull request).

Internally, the different configs types are encapsulated within a `variant`, a new C++17 data-structure introduced for such intents. The benefit of this is warning and errors issued by the compiler in case a new config version is added in the future but not explicitly handled.

Clickup [task](https://app.clickup.com/t/86c0532x0).  CI/CD [action](https://github.com/luxonis/depthai-core/actions/runs/11015182529).


jakgra (approved):
LGTM.

In the future when we add other config versions the getConfig() python version will have the autocomplete for the whole union of versions and getConfigV1() will retain the autocomplete for only version 1 of config. Something to be aware of when writing the examples for the documentation. @Erol444

PR #1134, opened 2024-09-23
Add method to custom ThreadedHostNode node to create a subnode

lnotspotl:
This PR adds the possibility of creating "sub-nodes" as part user defined python nodes. Though user-defined nodes are host-only, i.e. they cannot run on camera, a sub-node running on a camera can be created.

Clickup [task](https://app.clickup.com/t/86byk5a7x).


moratom (approved):
LGTM, thanks!

PR #1135, opened 2024-09-24
Replay IMU set latest timestamp to message

asahtik:


moratom (approved):
Let's add sequence number as well and merge, thanks!

Erol444:
@asahtik @moratom I still don't think this works as expected:
```
    while pipeline.isRunning():
        imudata: dai.IMUData = imuQ.get()
        frame = q.get()
        print(f"Seq: {imudata.getSequenceNum()}, TS: {imudata.getTimestamp()}, TS Device: {imudata.getTimestampDevice()}, Data: {imudata.packets[0].acceleroMeter.x}, {imudata.packets[0].acceleroMeter.y}, {imudata.packets[0].acceleroMeter.z}")
  ```
Will produce
```
Seq: 0, TS: 0:00:00, TS Device: 0:00:00, Data: -11.118672370910645, -1.1013327836990356, -0.9672574996948242
```
And TS is correctly set in MCAP:
image

PR #1139, opened 2024-09-26
Tarless holistic replay

asahtik:
Added ability to replay from folder instead of tar


moratom (approved):
Thanks Sebastjan!

Let's just add some tests before we merge for the CI please.

PR #1140, opened 2024-09-30
Add PointCloudData protobuf message definition and serialization routine

lnotspotl:


moratom (approved):
Thanks @lnotspotl !

Let's add an example and merge, thanks!

PR #1141, opened 2024-10-01
ImgTransformationsV2

asahtik:
Branched from `rvc2_image_manip_v2`


src/pipeline/datatype/ImgFrame.cpp
```diff
@@ -321,40 +323,6 @@ float ImgFrame::getSourceVFov() const {
     return verticalFovDegrees;
 }
```
moratom, line 324: Should be removed or gotten directly from the intrinsics IMO

src/pipeline/datatype/ImgTransformations.cpp
```diff
@@ -1,293 +1,822 @@
...
-
-    auto lastWidth = transformations[0].beforeTransformWidth;
-    auto lastHeight = transformations[0].beforeTransformHeight;
-    for(auto& transformation : transformations) {
-        if(transformation.beforeTransformWidth != lastWidth || transformation.beforeTransformHeight != lastHeight) {
-            return false;
-        }
-        lastWidth = transformation.afterTransformWidth;
-        lastHeight = transformation.afterTransformHeight;
+    srcMaskValid = dstMaskValid = srcBorderValid = dstBorderValid = false;
```
moratom, line 395: What is the intetion here? (trying to understand a bit the logic)

src/utility/ImageManipV2Impl.cpp
moratom, line 1: It doesn't keep track of the new transformations?

moratom (commented):
Thanks a bunch @asahtik!

Very glad to see transformations working on RVC4.



A few questions:
* Any reason why we use hfov/dfov for transformations still in ImgFrame? I think this should all be handled by ImgTransformation object
* Have we tested for performance - is the matrix first collapsed into one before remapping points?

src/utility/ImageManipV2Impl.cpp
asahtik: Transformations are kept track of in `ImageManipV2.cpp:50` and similarly on device side

src/pipeline/datatype/ImgFrame.cpp
```diff
@@ -245,12 +251,22 @@ ImgFrame& ImgFrame::setSourceHFov(float degrees) {
...
 
 float ImgFrame::getSourceHFov() const {
-    return HFovDegrees;
+    auto intrMat = transformation.getSourceIntrinsicMatrix();
+    if(HFovDegrees > 0.0f && intrMat[0][0] == 1.0f && intrMat[0][1] == 0.0f && intrMat[0][2] == 0.0f && intrMat[1][0] == 0.0f && intrMat[1][1] == 1.0f
+       && intrMat[1][2] == 0.0f && intrMat[2][0] == 0.0f && intrMat[2][1] == 0 && intrMat[2][2] == 1) {
+        return HFovDegrees;
+    } else {
```
moratom, line 258: Let's just remove it as a variable instead IMO, more trouble than worth to keep both.

src/pipeline/datatype/ImgFrame.cpp
```diff
@@ -287,155 +303,77 @@ float ImgFrame::getSourceDFov() const {
...
-        throw std::runtime_error("Source image has invalid horizontal FOV - horizontal FOV needs to be set before remapping");
-    }
+        // Calculate vertical FoV (in radians)
+        float verticalFoV = 2 * atan(getHeight() / (2.0f * fy));
 
-    if(!(destImage.getSourceHFov() > 0)) {
-        throw std::runtime_error("Destination image has invalid horizontal FOV - horizontal FOV needs to be set before remapping");
+        // Convert radians to degrees
+        return verticalFoV * 180.0f / (float)M_PI;
     }
```
moratom, line 347: Let's always take them from transformation, otherwise throw/error out.

include/depthai/common/ImgTransformations.hpp
```diff
@@ -1,114 +1,223 @@
...
+ */
+struct ImgTransformation {
    private:
-    static dai::Point2f applyMatrixTransformation(dai::Point2f point, std::vector<std::vector<float>>& matrix);
-};
+    std::array<std::array<float, 3>, 3> transformationMatrix = {{{1, 0, 0}, {0, 1, 0}, {0, 0, 1}}};
+    std::array<std::array<float, 3>, 3> transformationMatrixInv = {{{1, 0, 0}, {0, 1, 0}, {0, 0, 1}}};  // Precomputed inverse matrix
+    std::array<std::array<float, 3>, 3> sourceIntrinsicMatrix = {{{1, 0, 0}, {0, 1, 0}, {0, 0, 1}}};
+    std::array<std::array<float, 3>, 3> sourceIntrinsicMatrixInv = {{{1, 0, 0}, {0, 1, 0}, {0, 0, 1}}};
+    CameraModel distortionModel;
```
moratom, line 22: Should we have it as optional?

include/depthai/common/ImgTransformations.hpp
```diff
@@ -1,114 +1,223 @@
...
+     * Retrieve the inverse intrinsic matrix of the source sensor.
+     * @return Inverse intrinsic matrix
+     */
+    std::array<std::array<float, 3>, 3> getSourceIntrinsicMatrixInv() const;
+    std::vector<dai::RotatedRect> getSrcCrops() const;
+
+    bool getSrcMaskPt(size_t x, size_t y);
+    bool getDstMaskPt(size_t x, size_t y);
+    const std::vector<uint8_t>& getSrcMask(size_t srcWidth, size_t srcHeight);
+    const std::vector<uint8_t>& getDstMask();
```
moratom, line 134: Add comments on what the functions do&return please

src/pipeline/datatype/ImgTransformations.cpp
```diff
@@ -1,293 +1,409 @@
...
-        default:
-            break;
+    srcMask.resize(srcWidth * srcHeight);
+    memset(srcMask.data(), 0, srcMask.size());
+    for(auto i = srcMinY; i < srcMaxY; ++i) {
+        for(auto j = srcMinX; j < srcMaxX; ++j) {
+            srcMask[i * srcWidth + j] = 255;
+            for(auto crop : srcCrops) {
+                if(!isPointInRotatedRectangle({(float)j, (float)i}, crop)) {
+                    srcMask[i * srcWidth + j] = 0;
```
moratom, line 263: Transformations will be often copied from frame to frame, so having that optimization to store the mask is quite expensive in the copy constructor.


I think it would be better to remove the optimiziation and only keep the function per point.

moratom (commented):
Can you add a coment that explains a bit how the border calculation works?

```calcDstBorder()```

etc.

I can't seem to be able to wrap my head around it.

moratom (approved):
Thansk @asahtik !

Please just a add a few tests that make sure that the transformations arrive to host and are not nullptr.

That spatial detections have the imgTransformations, etc) before merging.

PR #1160, opened 2024-10-29
Add bindings to ImgFrame for the new transformations API

moratom:


src/pipeline/datatype/ImgTransformations.cpp
```diff
@@ -197,6 +197,62 @@ std::array<std::array<float, 3>, 3> ImgTransformation::getSourceIntrinsicMatrix(
...
+}
+float ImgTransformation::getVFov() const {
+    float fy = getIntrinsicMatrix()[1][1];
+
+    // Calculate vertical FoV (in radians)
+    float verticalFoV = 2 * atan(height / (2.0f * fy));
+
+    // Convert radians to degrees
+    return verticalFoV * 180.0f / (float)M_PI;
+}
```
moratom, line 255: Let's also add implementations for source FOVs and then have ImgFrame just passthrough the calls to the ImgTransformations to have the functionality more centralized

include/depthai/common/ImgTransformations.hpp
```diff
@@ -172,29 +200,33 @@ struct ImgTransformation {
...
      * intrinsics to remap the rect.
      * @param from Transformation to remap from
      * @param point RotatedRect to remap
+     * @param normalized Set to true if the rect is normalized
      */
-    dai::RotatedRect remapRectFrom(const ImgTransformation& from, dai::RotatedRect rect) const;
+    dai::RotatedRect remapRectFrom(const ImgTransformation& from, dai::RotatedRect rect, bool normalized = false) const;
 
     /**
      * Check if the transformations are valid. The transformations are valid if the source frame size and the current frame size are set.
```
moratom, line 232: I would add the normalized flag to dai::RotatedRect instead

moratom: or is there a specific reason to keep to not add it?

asahtik (approved):

PR #1161, opened 2024-10-30
Implement sending EncodedFrames from VideoEncoder on RVC4

asahtik:


moratom:
Nice, thanks!

* Update examples to use .out instead of .bitstream
* Add a basic test for the video encoder (we can do default values for all three codecs)

moratom (approved):
Neat, thanks!

PR #1163, opened 2024-11-07
Rect normalization improvements

asahtik:


moratom (approved):
Neat!

Can we also now make the necessary changes for the remapping example & update the example so normalization is not needed?

(We can split in a separate PR, but it might uncover missing details)

PR #1169, opened 2024-11-11
ImageManipV2 output size rounding and RVC2 crash fix

asahtik:


moratom (commented):
@asahtik can we split into two PRs please?

PR #1172, opened 2024-11-13
Add support for custom services to RemoteConnection

moratom:


examples/python/Visualizer/custom_services.py
```diff
@@ -0,0 +1,27 @@
...
+def stop():
+    global isRunning
+    isRunning = False
+
+def signal_handler(sig, frame):
+    print('You pressed Ctrl+C!')
+    stop()
+    sys.exit(0)
+
+signal.signal(signal.SIGINT, signal_handler)
```
asahtik, line 17: Is this needed? It'd probably be cleaner to put time.sleep in a try except KeyboardInterrupt block inside the while loop (if it works)

asahtik (approved):
I have a nitpick about the python example, otherwise LGTM

PR #1173, opened 2024-11-14
Expanded RotatedRect object definition to accept floats

aljazkonec1:
This PR adds a new definition to initialize RotatedRect object with just floats.

### DIFF
--- a/bindings/python/src/pipeline/CommonBindings.cpp
+++ b/bindings/python/src/pipeline/CommonBindings.cpp
@@ -97,7 +97,7 @@ void CommonBindings::bind(pybind11::module& m, void* pCallstack){
         .def(py::init<>())
         .def(py::init<Point2f, Size2f, float>())
         .def(py::init<Rect, float>())
-        .def(py::init<float float float float float>(), py::arg("center_x"), py::arg("center_y"), py::arg("size_width"), py::arg("size_height"), py::arg("angle"))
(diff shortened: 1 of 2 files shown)


bindings/python/src/pipeline/CommonBindings.cpp
```diff
@@ -97,6 +97,7 @@ void CommonBindings::bind(pybind11::module& m, void* pCallstack){
         .def(py::init<>())
         .def(py::init<Point2f, Size2f, float>())
         .def(py::init<Rect, float>())
+        .def(py::init<float float float float float>(), py::arg("center_x"), py::arg("center_y"), py::arg("size_width"), py::arg("size_height"), py::arg("angle"))
```
asahtik, line 100: I think you need commas between the floats

asahtik (approved):
This is just a subjective opinion, which you can feel free to disregard, but I generally don't like it when there are many arguments of the same type together like this, because it's easy to forget what each argument means. 
I'd like to offer an alternative:
Use `std::tuple<float, float>` to define center and size so in Python the constructor can be called with `dai.RotatedRect((x, y), (width, height), angle)`.

moratom:
Agree with @asahtik - as an example:
```dai.RotatedRect(center=dai.Point2f(200.0, 300.0), size=dai.Size2f(100, 200),angle=0.7)```
is a lot more readable than
```dai.RotatedRect(200.0, 300.0, 100, 200, 0.7)```

PR #1185, opened 2024-11-25
Move MacOS building to self hosted to as github runner is broken

moratom:


moratom:
Seems like fix will come soon enough on the official runners:
https://github.com/actions/runner-images/pull/11031

PR #1186, opened 2024-12-06
RVC2 img transformations

asahtik:


moratom (approved):
Thanks - let's just add the examples so they are tested -https://github.com/luxonis/depthai-core/blob/v3_develop/examples/python/CMakeLists.txt.

PR #1188, opened 2024-12-10
V3 snaps error handling

Serafadam:
* Fix CMake options for EventsManager
* Throw with a message when API key is empty
* Fix python example
* Properly resolve url
* Fix headers


src/utility/EventsManager.cpp
```diff
@@ -109,6 +111,9 @@ EventsManager::EventsManager(std::string url, bool uploadCachedOnStart, float pu
     sourceAppId = utility::getEnv("AGENT_APP_ID");
     sourceAppIdentifier = utility::getEnv("AGENT_APP_IDENTIFIER");
     token = utility::getEnv("DEPTHAI_HUB_API_KEY");
+    if(token.empty()) {
```
moratom, line 114: Let's add a way to set that in the constructor and document the ENV variables in README.md

examples/python/HostNodes/events.py
moratom, line 1: I think best to comment out things that fail and only leave in the "positive examples".

+ move the examples to a separate SnapsEvents folder and let's add a comment on top on how can someone use the example (what's needed on the Hub side)

moratom (approved):
Thanks!

PR #1189, opened 2024-12-10
Add missing API to script node on RVC4 + tests and examples

moratom:
Adds the missing API to script node on RVC4 + some tests and examples.


asahtik (changes requested):
`script_switch_left_right_cameras` and `script_switch_all_cameras` seem to do the same thing

moratom:
> `script_switch_left_right_cameras` and `script_switch_all_cameras` seem to do the same thing

Almost, the idea was that one was easier to understand and the other more generic.
Which one would you leave in?

asahtik (approved):

PR #1192, opened 2024-12-11
SpatialDetection network: align depth to input image

SzabolcsGergely:


examples/python/SpatialDetectionNetwork/spatial_detection.py
```diff
@@ -82,7 +82,9 @@ def drawDetections(self, frame, detection, frameWidth, frameHeight):
     # setting node configs
     stereo.setDefaultProfilePreset(dai.node.StereoDepth.PresetMode.HIGH_DENSITY)
     stereo.setExtendedDisparity(True)
-    stereo.setOutputSize(640, 400)
+    platform = p.getDefaultDevice().getPlatform()
+    if platform == dai.Platform.RVC2:
+        stereo.setOutputSize(640, 400)
```
moratom, line 87: Needed even with `.inputAlignTo`?

moratom: Maybe good to add a comment in that case

moratom (approved):

PR #1198, opened 2024-12-18
V3 RGBD Node

Serafadam:
* Host RGBD Node
* Basalt fixes
* Using stock TBB for Basalt
* Added optional Kompute support for performing calculations on GPU devices with Vulkan support


src/pipeline/node/host/RGBD.cpp
```diff
@@ -0,0 +1,320 @@
...
+    auto calibHandler = getParentPipeline().getDefaultDevice()->readCalibration();
+    auto camID = static_cast<CameraBoardSocket>(colorFrame->getInstanceNum());
+    auto width = colorFrame->getWidth();
+    auto height = colorFrame->getHeight();
+    auto intrinsics = calibHandler.getCameraIntrinsics(camID, width, height);
+    float fx = intrinsics[0][0];
+    float fy = intrinsics[1][1];
+    float cx = intrinsics[0][2];
+    float cy = intrinsics[1][2];
+    pimpl->setIntrinsics(fx, fy, cx, cy, width, height);
```
moratom, line 264: Let's get the intrinsics from the first frame instead (after rvc2_img_transformations_v2 is merged which will likely be before this PR)

src/pipeline/node/host/RGBD.cpp
```diff
@@ -0,0 +1,320 @@
...
+    }
+    auto pipeline = getParentPipeline();
+    auto colorCam = pipeline.create<node::Camera>()->build();
+    auto depth = pipeline.create<node::StereoDepth>()->build(true);
+    auto* out = colorCam->requestOutput(size, dai::ImgFrame::Type::RGB888i);
+    out->link(inColor);
+    out->link(depth->inputAlignTo);
+    depth->depth.link(inDepth);
+    return build();
+}
```
moratom, line 242: Let's add a warning that this API is "global" and that if used the cameras can't be used for other things.
And we can add the same warning in StereoDepth too.

moratom (commented):
Thanks Adam! Look forward to getting this in :)

include/depthai/common/Point3fRGB.hpp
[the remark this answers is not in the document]
Serafadam: Foxglove expects rgb pointclouds size to be 15 bytes when deserializing, though we could add alpha channel here to make this 16

src/pipeline/node/host/RGBD.cpp
```diff
@@ -0,0 +1,320 @@
...
+
+        std::vector<Point3fRGB> points;
+        // Fill the point cloud
+        auto* depthData = depthFrame->getData().data();
+        auto* colorData = colorFrame->getData().data();
+        // Use GPU to compute point cloud
+        pimpl->computePointCloud(depthData, colorData, points);
+
+        pc->setPointsRGB(points);
+        pcl.send(pc);
```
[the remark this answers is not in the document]
Serafadam: Width & height are checked in initialize method. For aligment checking, we should just check frame id, right?

src/basalt/BasaltVIO.cpp
```diff
@@ -218,6 +220,7 @@ void BasaltVIO::initialize(std::vector<std::shared_ptr<ImgFrame>> frames) {
     vio->out_state_queue = pimpl->outStateQueue;
     vio->opt_flow_depth_guess_queue = optFlowPtr->input_depth_queue;
     vio->opt_flow_state_queue = optFlowPtr->input_state_queue;
+    vio->opt_flow_lm_bundle_queue = optFlowPtr->input_lm_bundle_queue;
```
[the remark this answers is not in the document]
Serafadam: Partly, yes, though main improvements are in basalt lib itself

src/pipeline/node/host/RGBD.cpp
```diff
@@ -0,0 +1,357 @@
...
+        auto rgbdData = std::make_shared<RGBDData>();
+        rgbdData->depthFrame = *depthFrame;
+        rgbdData->rgbFrame = *colorFrame;
+        rgbd.send(rgbdData);
+    }
+}
```
moratom, line 339: Maybe best to detect for each output RGBD and PCL if it is connected anywhere and only make the PCL / RGBD message if it will be used to save resources.

include/depthai/pipeline/datatype/RGBDData.hpp
```diff
@@ -0,0 +1,30 @@
...
+   public:
+    /**
+     * Construct RGBD message.
+     */
+    RGBDData() = default;
+
+    virtual ~RGBDData() = default;
+
+    ImgFrame rgbFrame;
+    ImgFrame depthFrame;
```
moratom, line 21: I think that it would be best to make this in a similar way as MessageGroup is done for performance reasons.
(so these two would be pointers and serialization would require a specialization)

examples/python/RGBD/rgbd.py
moratom, line 1: Playing around with it, at least for me, 640x400 provides a better overall experience (the PCLs are dense enough but everything runs at 30 FPS)
Same goes for other examples.

include/depthai/pipeline/node/host/RGBD.hpp
```diff
@@ -0,0 +1,77 @@
...
+    Input& inDepth = inputs[depthInputName];
+
+    /**
+     * Output point cloud.
+     */
+    Output pcl{*this, {"pcl", DEFAULT_GROUP, {{DatatypeEnum::PointCloudData, true}}}};
+    /**
+     * Output RGBD frames.
+     */
+    Output rgbd{*this, {"rgbd", DEFAULT_GROUP, {{DatatypeEnum::RGBDData, true}}}};
```
moratom, line 39: The bindings are missing for this one

bindings/python/src/pipeline/node/StereoDepthBindings.cpp
```diff
@@ -122,9 +122,10 @@ void bind_stereodepth(pybind11::module& m, void* pCallstack) {
              py::arg("presetMode") = StereoDepth::PresetMode::HIGH_DENSITY,
              DOC(dai, node, StereoDepth, build))
         .def("build",
-             static_cast<std::shared_ptr<StereoDepth> (StereoDepth::*)(bool autoCreate, StereoDepth::PresetMode)>(&StereoDepth::build),
+             static_cast<std::shared_ptr<StereoDepth> (StereoDepth::*)(bool autoCreate, StereoDepth::PresetMode, std::pair<int, int> size)>(&StereoDepth::build),
              py::arg("autoCreateCameras"),
-             py::arg("presetMode") = StereoDepth::PresetMode::HIGH_DENSITY)
+             py::arg("presetMode") = StereoDepth::PresetMode::HIGH_DENSITY,
+             py::arg("size") = std::pair<int, int>{1280, 720})
```
moratom, line 128: Let's default to 1280x800 instead, it's the most common resolution for mono cameras.

examples/python/RGBD/rgbd_o3d.py
moratom, line 1: The pointcloud seems to be upside down for me at the beginning, is it upright on your side?

tests/src/ondevice_tests/rgbd_test.cpp
```diff
@@ -0,0 +1,76 @@
...
+        REQUIRE(pcl->getMinY() <= pcl->getMaxY());
+        REQUIRE(pcl->getMinZ() <= pcl->getMaxZ());
+    }
+}
+TEST_CASE("rgbd autocreate") {
+    // Create pipeline
+    dai::Pipeline pipeline;
+    auto rgbd = pipeline.create<dai::node::RGBD>()->build(true);
+
+    auto outQ = rgbd->pcl.createOutputQueue();
```
moratom, line 63: Maybe we can add a test for rgbd as well

moratom (commented):
Thanks!

Tested it out and it seems to work well:) 

I left a few comments for the new additions, but LGTM otherwise.

examples/python/RGBD/rgbd_o3d.py
Serafadam: It's in RDF frame convention, maybe that's why?
image

src/pipeline/node/host/RGBD.cpp
```diff
@@ -0,0 +1,368 @@
...
+std::shared_ptr<RGBD> RGBD::build() {
+    sync->out.link(inSync);
+    inColor.setBlocking(false);
+    inColor.setMaxSize(1);
+    inDepth.setBlocking(false);
+    inDepth.setMaxSize(1);
+    inSync.setBlocking(false);
+    inSync.setMaxSize(1);
+    return std::static_pointer_cast<RGBD>(shared_from_this());
+}
```
lnotspotl, line 248: Can we move this to `buildInternal`?

examples/cpp/RGBD/rgbd.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+rerun::Collection<rerun::TensorDimension> tensorShape(const cv::Mat& img) {
+    return {size_t(img.rows), size_t(img.cols), size_t(img.channels())};
+};
+class RerunNode : public dai::NodeCRTP<dai::node::ThreadedHostNode, RerunNode> {
+   public:
+    constexpr static const char* NAME = "RerunNode";
+
+   public:
+    void build() {}
```
lnotspotl, line 13: If there is some setup, we should put it into `buildInternal` ?

bindings/python/src/pipeline/node/StereoDepthBindings.cpp
moratom: Bump on this

moratom (approved):

PR #1200, opened 2024-12-19
Add modelzoo metadata, improve integrity

lnotspotl:
Clickup task: https://app.clickup.com/t/86c1bmx1r

This PR improves our functions for downloading models from the hub. A new `metadata.yaml` file is generated for each and every downloaded model for easier checking and hash validation.


src/modelzoo/Zoo.cpp
```diff
@@ -80,7 +80,7 @@ class ZooManager {
     /**
      * @brief Download model from model zoo
      */
-    void downloadModel();
+    void downloadModel(const nlohmann::json& responseJson);
```
moratom, line 83: Perhaps nicer to return the object instead since the function is now void anyhow.

src/modelzoo/Zoo.cpp
```diff
@@ -351,6 +425,21 @@ void downloadModelsFromZoo(const std::string& path, const std::string& cacheDire
...
+bool ZooManager::internetIsAvailable() const {
+    constexpr int timeout_ms = 5000;
+    constexpr std::string_view host = "http://example.com";
+    try {
+        cpr::Response r = cpr::Get(cpr::Url{host}, cpr::Timeout{timeout_ms});
+        return r.status_code == cpr::status::HTTP_OK;
+    } catch(const std::exception& e) {
+        return false;
+    }
+}
```
moratom, line 437: In case of no internet this will slow down startup by 5 seconds, I think we should reduce this to ~500ms instead.


Would it maybe make sense to try pinging the URL we actually need instead of `http://example.com`?
Some customers might setup the network in a way where only the Zoo is accessible for the model updates, etc.

moratom (commented):
Thanks Jakub!

Generally looks good, left just a couple of comments.

src/modelzoo/Zoo.cpp
lnotspotl: Are you suggesting `downloadModel` returns a path to the downloaded file?

src/modelzoo/Zoo.cpp
lnotspotl: I performed some empirical tests and 500 milliseconds turn out not to be enough. At times, `depthai` would not receive a response from the hub fast enough and thus conclude that there is no internet connection despite my PC being connected to the network. A timeout of 1 second turns out to be enough.

lnotspotl: Turns out maybe even that 1 second is not sufficient.

image

lnotspotl:
Update:


I have just introduced two new environment variables:
-  **DEPTHAI_ZOO_INTERNET_CHECK** ... whether or not to perform model zoo internet check
- **DEPTHAI_ZOO_INTERNET_CHECK_TIMEOUT** ... timeout for internet check in milliseconds

This should fix all the issues and concerns that were raised

src/modelzoo/Zoo.cpp
```diff
@@ -351,6 +430,35 @@ void downloadModelsFromZoo(const std::string& path, const std::string& cacheDire
...
+    constexpr std::string_view host = MODEL_ZOO_HEALTH_ENDPOINT;
+
+    // Check if internet is available
+    bool connected = false;
+    try {
+        cpr::Response r = cpr::Get(cpr::Url{host}, cpr::Timeout{timeout_ms});
+        connected = r.status_code == cpr::status::HTTP_OK;
+    } catch(const std::exception& e) {
+        // pass
+    }
```
moratom, line 449: Maybe better to only catch the expected exception here

src/modelzoo/Zoo.cpp
```diff
@@ -351,6 +430,35 @@ void downloadModelsFromZoo(const std::string& path, const std::string& cacheDire
     }
 }
 
+bool ZooManager::internetIsAvailable() const {
```
moratom, line 433: This function can be made static (doesn't access any object state) & perhaps better to rename to something more specific like `connectionToZooAvailable` since we test that specifically.

src/modelzoo/Zoo.cpp
```diff
@@ -295,26 +345,55 @@ std::string getModelFromZoo(const NNModelDescription& modelDescription, bool use
...
+        Logging::getInstance().logger.warn("Cached model hash does not match response hash, downloading anew ...");
     }
 
     // Remove cached model if present
     if(modelIsCached) {
         zooManager.removeModelCacheFolder();
     }
 
+    if(!internetIsAvailable) {
+        throw std::runtime_error("No internet connection available. Please check your network settings and try again.");
```
moratom, line 388: Maybe we can add the URL in the error message that we use for checking the connection (so users can debug easier)

src/modelzoo/Zoo.cpp
moratom: Never mind, I misunderstood the function signature, missed the `const` and thought that that the input reference was used for the return.

moratom (commented):
Added a few more nitpicks, but looks good!

src/modelzoo/Zoo.cpp
lnotspotl: Yeah, I've changed it so that the we just catch `cpr::Error`-derived errors.

src/modelzoo/Zoo.cpp
lnotspotl: Added a better logging messages.

moratom (approved):
Thanks!

PR #1204, opened 2025-01-02
Enable implicit conversion between str and NNModelDescription for python bindings in build methods

lnotspotl:
Clickup task: https://app.clickup.com/t/86c191hr8


moratom (approved):
Thanks!

PR #1206, opened 2025-01-07
RVC4 manual focus

jakgra:


tests/src/ondevice_tests/manual_focus_test.cpp
```diff
@@ -0,0 +1,29 @@
+#include <catch2/catch_all.hpp>
+
+#include "depthai/depthai.hpp"
+
+TEST_CASE("Test manual focus") {
+    dai::Pipeline pipeline;
+    auto fps = 30.0f;
+    auto camRgb = pipeline.create<dai::node::Camera>()->build();
```
zrezke, line 8: Will this always default to the color cam? Or should we discover the color cam first and then set the socket?

moratom: This will always try to default to the "main camera" - it can stay in the test iMO.

Right now it auto-selects the first socket out of CAM_A, CAM_B, CAM_C that is available.

PR #1207, opened 2025-01-07
Deprecate/remove mxId and replace with deviceId

moratom:
This PR renames mxId to deviceId as mxId is not accurate anymore with the addition of RVC4 lineup.


asahtik (commented):
Shouldn't `getDeviceByMxId` also be renamed to `getDeviceById` or something like that? Or does this function not work with RVC4 devices?

moratom:
> Shouldn't `getDeviceByMxId` also be renamed to `getDeviceById` or something like that? Or does this function not work with RVC4 devices?
Good catch! Will replace/deprecate.

PR #1208, opened 2025-01-08
getEnv refactor -> getEnvAs and getEnvAsChecked

lnotspotl:
Based on our discussion here: https://luxonis.slack.com/archives/C065PDYKNAK/p1735943335766909

TLDR: remove older `getEnv` utility function for retrieving environment variables and replace it with `getEnvAs` and `getEnvAsChecked`, two template functions allowing one to specify the return type and performing necessary parsing internally.


src/utility/Environment.hpp
```diff
@@ -1,14 +1,147 @@
...
+    // Return immediately if cached
+    if(cache.count(var) > 0) {
+        return cache[var];
+    }
+
+    // Get environment variable
+    std::string value = spdlog::details::os::getenv(var.c_str());
+
+    // Check if we should use the default
+    bool useDefault = value.empty();
```
Serafadam, line 41: I think we should also log message if the env variable is not found so that we can see that there is no typo & what default value will be substituted here

Serafadam (changes requested):
Overall LGTM, but I would add tests for both functions before merging

src/utility/Environment.hpp
lnotspotl: I changed it to `INFO`

Serafadam (approved):
LGTM, thanks for the changes, we can merge after CI passes

PR #1212, opened 2025-01-15
ImageManipV2 bugfixes

asahtik:
Fix ImageManipV2 rebuilding and handling 2B grayscale images


moratom (approved):
Let's bump the RVC4 FW before the merge, thanks!

PR #1214, opened 2025-01-20
V3 Replay node in NN build methods

Serafadam:
Adds new overloads for build methods in NN, Detections and SpatialDetection nodes. Also some cosmetic changes


moratom (approved):
LGTM - let's just add a few basic tests before merging

bindings/python/src/pipeline/node/NeuralNetworkBindings.cpp
```diff
@@ -85,6 +85,24 @@ void bind_neuralnetwork(pybind11::module& m, void* pCallstack) {
             py::arg("model"),
             py::arg("fps") = 30.0f,
             DOC(dai, node, NeuralNetwork, build))
+        .def(
+            "build",
+            [](NeuralNetwork& self, const std::shared_ptr<ReplayVideo>& input, const dai::NNArchive& nnArchive, float fps) {
```
asahtik, line 90: Use `py::overload_cast<const std::shared_ptr<ReplayVideo>&, const dai::NNArchive&, float>(&NeuralNetwork::build)` instead of lambda for consistency

examples/python/RVC2/RecordReplay/replay_yolo.py
moratom, line 1: The examples should actually be cross platform, so it can be moved to "DetectionNetwork" folder IMO

Serafadam: And for CPP to HostNodes?

moratom: I think we can make the same one in DetectionNetwork (note I renamed the example a bit to be consistent with the other ones in the folder and added it to tests).

PR #1218, opened 2025-01-22
Updated imageManip examples by adding full hd frame as well, 

Erol444:
so it's more clear (for docs as well)


moratom (approved):
LGTM

Erol444:
@moratom some check didn't pass, but seems like everything is ok?

PR #1225, opened 2025-01-29
V3 Yolo model type deduction

Serafadam:
Yolo models are first deduced from NNArchive now


include/depthai/common/DetectionParserOptions.hpp
moratom, line 1: Should we maybe go with an enum here?

I think we should also add a way to specify the subtype with setters.

Serafadam: IIRC in NNArchive spec it is defined as a string, I think this way is a bit simpler if we want to still have the option to use tensor layer names

moratom (approved):

PR #1229, opened 2025-02-01
Added queue.getFps() function and tests for this function

Erol444:
Tests fail anyways due to other tests:
```
./message_queue_test
Randomness seeded to: 2974482811
message_queue_test(41900,0x16bb27000) malloc: double free for ptr 0x14a009200
message_queue_test(41900,0x16bb27000) malloc: *** set a breakpoint in malloc_error_break to debug

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
message_queue_test is a Catch2 v3.7.1 host application.
Run with -? for options

-------------------------------------------------------------------------------
MessageQueue - Multiple producers and consumers
-------------------------------------------------------------------------------
/Users/erik/Documents/luxonis/depthai-core/tests/src/onhost_tests/message_queue_test.cpp:58
...............................................................................

/Users/erik/Documents/luxonis/depthai-core/tests/src/onhost_tests/message_queue_test.cpp:58: FAILED:
due to a fatal error condition:
  SIGABRT - Abort (abnormal termination) signal

===============================================================================
test cases: 4 | 3 passed | 1 failed
assertions: 7 | 6 passed | 1 failed

zsh: abort      ./message_queue_test
```

### DIFF
--- a/src/pipeline/MessageQueue.cpp
+++ b/src/pipeline/MessageQueue.cpp
@@ -2,6 +2,7 @@
 
 // std
 #include <chrono>
+#include <deque>
 #include <iostream>
 
 // project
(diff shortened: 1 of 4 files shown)


src/pipeline/MessageQueue.cpp
```diff
@@ -111,6 +143,15 @@ void MessageQueue::send(const std::shared_ptr<ADatatype>& msg) {
...
+
+    // Record the timestamp for FPS calculation
+    {
+        auto now = std::chrono::steady_clock::now();
+        fpsQueue.push_back(now);
+        if(fpsQueue.size() > FPS_QUEUE_MAX_SIZE) {
+            fpsQueue.pop_front();
+        }
+    }
 }
```
moratom, line 155: The function must be thread safe, so the fpsQueue needs to be protected with a mutex.

src/pipeline/MessageQueue.cpp
```diff
@@ -70,6 +71,37 @@ unsigned int MessageQueue::isFull() const {
     return queue.isFull();
 }
 
+float MessageQueue::getFps() {
+    std::unique_lock<std::mutex> lock(callbacksMtx);
```
moratom, line 75: A mutex for the FPS queue should be separate and acquired here

src/pipeline/MessageQueue.cpp
```diff
@@ -70,6 +71,37 @@ unsigned int MessageQueue::isFull() const {
...
+
+    // Get current time
+    auto now = std::chrono::steady_clock::now();
+    auto threshold = now - std::chrono::seconds(2);
```
moratom, line 79: What's the reason behind the treshold? I would set a max queue size instead so it's more consistent for low and high FPS

moratom (commented):
Thanks!

Left a few comments about thread safety, but other than that looks good to me.

Might be worth considering a LockingQueue (the same as is used for the messages) since it's already thread safe and "battle tested".

SzabolcsGergely:
How about getMessagesPerSecond instead getFPS?

Erol444:
@SzabolcsGergely while it's more accurate, I think most people are used to FPS, even if it's actually Inference per second / Messages per second. Pehraps we could go with `getQueueHz()`, which is also quite widely used (aka hz == fps)?

SzabolcsGergely:
> @SzabolcsGergely while it's more accurate, I think most people are used to FPS, even if it's actually Inference per second / Messages per second. Pehraps we could go with `getQueueHz()`, which is also quite widely used (aka hz == fps)?

I don't see getQueueHz fit/good choice, personally.
This measurement refers to queue message rate per second.

PR #1230, opened 2025-02-02
Add missing serialize method for Tracklets

lnotspotl:
The `serialize` method would not be defined for `Tracklets` and thus we would silently fall back to `Buffer`'s serialization method... 

This PR aims to fix this silent bug.


moratom (approved):
Thanks!

Please check if the same thing is missing on `v2` as well, since the code was addopoted from there.

lnotspotl:
I ran a quick test and this was just a `v3` issue, the equivalent MRE code in `v2` works as expected. Therefore, no changes are needed on the `v2` side.

PR #1237, opened 2025-02-11
Script Node: Do not override defaults in ScriptProperties

SzabolcsGergely:
Not a very nice way to override defaults in Properties. Misleading.


moratom (approved):
Agreed.

PR #1240, opened 2025-02-14
Stereo rvc4 config extension

SzabolcsGergely:
Added optional SW confidence invalidation and left right check.


moratom (approved):
Thanks!

PR #1244, opened 2025-02-18
Bom integration

danilo-pejovic:
## Purpose
Adding bom integration to depthai-core pipeline

v3_develop (and other branches from it unless changed) will be able to have multiple workflows active at once - potentially swarming the HiL. If it HiL setup starts struggling I will look for better solution later on.


moratom:
Any reason why we didn't go with a separate workflow (since it's two workflows in one anyways now)?

.github/workflows/test.workflow.yml
moratom, line 1: These changes can now be removed right?

PR #1245, opened 2025-02-18
Bugfix/device subnodes

lnotspotl:
## Purpose
When we have a `HostRunnable` device node that is a subnode of a `ThreadedNode`, this node's device does not get set properly. This PR patches that by setting any `DeviceNode`'s device if it was not set in it's constructor

## Testing & Validation
Added a test to `subnode_tests.cpp`


Serafadam (approved):
LGTM

PR #1263, opened 2025-03-10
Set plane alignment to 1 

asahtik:
## Purpose
Set plane alignment to 1 for RVC2 NN compatibility. Fix `getCvFrame` for strided NV12 frames.

## Specification
Plane alignment is now set to 1 (not aligned)

None / not applicable

## Deployment Plan
None / not applicable

## Testing & Validation
None / not applicable


moratom:
Can we also fix the `getCvFrame()` function in case it's not handling the striding correctly?

src/opencv/ImgFrame.cpp
```diff
@@ -187,16 +187,12 @@ cv::Mat ImgFrame::getCvFrame(cv::MatAllocator* allocator) {
...
-                cv::Mat frameY(s, type, getData().data(), step);
-                cv::Mat frameUV(s / 2, type, getData().data() + getPlaneStride(), step);
-                cv::cvtColorTwoPlane(frameY, frameUV, output, code);
-            }
+            cv::Size s(getWidth(), getHeight());
+            int type = CV_8UC1;
+            int step = getStride();
+            cv::Mat frameY(s, type, getData().data(), step);
+            cv::Mat frameUV(s / 2, type, getData().data() + getPlaneStride(), step);
+            cv::cvtColorTwoPlane(frameY, frameUV, output, code);
```
moratom, line 195: Are we now always creating another copy of the data or not in this case?

moratom (approved):

src/opencv/ImgFrame.cpp
moratom: I measured the performance and it's a negligible difference.

PR #1285, opened 2025-03-31
use an env variable to specify where new zoo models are to be cached

lnotspotl:
## Purpose
Feature request from xxx team.

I have also decided to move `NNModelDescription` declarations to `Zoo.hpp` to allow for easier code reuse. Moreover, I think the two headers, `NNModelDescription.hpp` and `Zoo.hpp` were usually included together anyways.


moratom (approved):
Thanks!

PR #1287, opened 2025-04-01
Adding a test wrapper script

jakaskerl:
Adding another wrapper to example tests. Point was to make sure the device comes back online after running the test. Only then is the test considered "Passed".

**Current workflow:**
run_tests.py --> ctest --> ExecuteTestTimeout --> wrapper --> test(cpp/python)

**Timeouts enforced by:**
- ctest: 1000s (including all possible tests that are run)
- ExecuteTestTimeout: 300s for python (in case device is dead after the test) and 60s for cpp examples (seems to work ok atm)
- wrapper: 30s for both python and cpp examples 

Will likely need refactoring in the future since standard tests keep failing.


examples/cpp/CMakeLists.txt
```diff
@@ -53,9 +53,14 @@ function(dai_add_example example_name example_src enable_test use_pcl)
 
     # If 'DEPTHAI_TEST_EXAMPLES' is ON, then examples will be part of the test suite
     if(${enable_test} AND DEPTHAI_TEST_EXAMPLES)
-        add_test(NAME ${example_name} COMMAND ${CMAKE_COMMAND}
-            -DTIMEOUT_SECONDS=10
-            -P ${CMAKE_CURRENT_LIST_DIR}/cmake/ExecuteTestTimeout.cmake $<TARGET_FILE:${example_name}> ${arguments}
+        add_test(NAME ${example_name} COMMAND 
+            ${CMAKE_COMMAND}
+            -DTIMEOUT_SECONDS=60
```
moratom, line 58: Shouldn't we bump this up to 300 to match Python too?

moratom (changes requested):
Thanks @jakaskerl !! Looking forward to getting this merged.

PR #1302, opened 2025-04-21
Fixes for double

phodina:
Minor tweaks around floating point numbers that gave warnings during compilation.

### DIFF
--- a/include/depthai/common/Color.hpp
+++ b/include/depthai/common/Color.hpp
@@ -26,7 +26,8 @@ struct Color {
     Color(float r, float g, float b, float a) {
         // r,g,b,a values should be in range [0.0, 1.0]
         auto check = [](float val) -> float {
-            if(val < 0.0 || val > 1.0) {
+        const float epsilon = 1e-6f;
+            if(val < (0.0f - epsilon) || val > (1.0f + epsilon)) {
                 throw std::invalid_argument("Color values should be in range [0.0, 1.0]");
             }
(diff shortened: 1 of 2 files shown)


include/depthai/common/Color.hpp
```diff
@@ -26,7 +26,8 @@ struct Color {
     Color(float r, float g, float b, float a) {
         // r,g,b,a values should be in range [0.0, 1.0]
         auto check = [](float val) -> float {
-            if(val < 0.0 || val > 1.0) {
+        const float epsilon = 1e-6f;
+            if(val < (0.0f - epsilon) || val > (1.0f + epsilon)) {
```
Pandapip1, line 30: Comparisons to 0 and 1 should be fine since they can be represented exactly as an IEEE float, no?

hacker1024: As far as I'm aware, C++ doesn't specify a binary float representation format.

Pandapip1: It does, IIUC. C++ uses C types, and C specifies IEEE 754 for floating point integers.

moratom (approved):

include/depthai/common/Color.hpp
phodina: Well there's the `fpclassify` in C++ that could be used instead

https://en.cppreference.com/w/cpp/numeric/math/fpclassify

PR #1303, opened 2025-04-21
cmake: Add option to enable Hunter to fetch data

phodina:
This patch keeps the Hunter package manager by default enabled but allows to turn it of and bring this repo to offline snadboxed built.


CMakeLists.txt
```diff
@@ -1,6 +1,7 @@
 cmake_minimum_required(VERSION 3.20)
 # Early options
 option(DEPTHAI_ENABLE_LIBUSB "Enable usage of libusb and interaction with USB devices" ON)
+option(DEPTHAI_HUNTER_ENABLE "Enable Hunter to fetch all dependencies" ON)
```
moratom, line 4: We should rename this, since it's not Hunter who's downloading the files, we've "ported" that private_data function directly into our codebase.

Perhaps DEPTHAI_FETCH_ARTIFACTS or something along those lines.

moratom (changes requested):
Thanks! Left a comment but otherwise LGTM.

CMakeLists.txt
phodina: Sure, just put there a variable that would enable/disable the download part.

phodina: Renamed the option and comments to just fetching data

moratom (approved):

phodina:
For the C++ examples there's redefinition of Lenna images to download.

lnotspotl:
When compiled with cmake flags
```
-DDEPTHAI_TEST_EXAMPLES=ON -DDEPTHAI_FETCH_ARTIFACTS=OFF
```
a handful of examples, specifically those that depend on those external artifacts, won't run. I think those examples should only be built if their artifacts are present, i.e. they too should be in the `if(DEPTHAI_FETCH_ARTIFACTS)` block.

Otherwise, LGTM.

lnotspotl (commented):
When compiled with cmake flags
```
-DDEPTHAI_TEST_EXAMPLES=ON -DDEPTHAI_FETCH_ARTIFACTS=OFF
```
a handful of examples, specifically those that depend on those external artifacts, won't run. I think those examples should only be built if their artifacts are present, i.e. they too should be in the `if(DEPTHAI_FETCH_ARTIFACTS)` blocks.

Otherwise, LGTM.

phodina:
Okay, will look at those and add the check around as well

PR #1304, opened 2025-04-23
fix zoo_helper ci for linux-x86_64

lnotspotl:
## Purpose
Pin `almalinux` docker image version. The last update broke a couple of dependencies. Moreover, `nasm.us` has been down for quite a while now, thus I've decided to pull one of the dependencies, `nasm`, from what seems to be a fork of thereof.


moratom (approved):
LGTM, thanks!

PR #1308, opened 2025-04-28
ToF node enhancements, host-runnable ToF and image filters

lnotspotl:
The goal of this PR is to introduce a new host-runnable node for general image filtering (`ImageFilters`) as well as a new node called `ToFDepthConfidenceFilter` designed specifically for depth frame filtering based on data coming from the `ToF` node.

Additionally, the `ToF` node was renamed to `ToFBase` and a new `ToF` node took its place, combining `ToFBase`, `ToFDepthConfidenceFilter` and `ImageFilters` into a single node. This gives power-users the opportunity to fiddle with the low-level stuff (they can still use the origin `ToF` node implementation by instantiating the `ToFBase` class). For an average user, however,  the new `ToF` node with a predefined set of filters will likely suffice.

image

The image filters were taken from the FW codebase and the ToF filter was implemented based on a reference python implementation provided by the CV team. Both the `ToFDepthConfidenceFilter` node and the `ImageFilters` node allow for run-time parameter updates thanks to the possibility of sending in input configs. In addition to that, the `ImageFilters` node need not have a fixed filter pipeline, it is possible to change that filter pipeline completely at runtime.


include/depthai/properties/DepthFiltersProperties.hpp
```diff
@@ -0,0 +1,48 @@
...
+struct MedianFilterParams {
+    bool enable = false;
+    dai::StereoDepthConfig::MedianFilter median = dai::StereoDepthConfig::MedianFilter::KERNEL_3x3;
+};
+DEPTHAI_SERIALIZE_EXT(MedianFilterParams, enable, median);
+
+typedef dai::StereoDepthConfig::PostProcessing::SpatialFilter SpatialFilterParams;
+typedef dai::StereoDepthConfig::PostProcessing::SpeckleFilter SpeckleFilterParams;
+typedef dai::StereoDepthConfig::PostProcessing::TemporalFilter TemporalFilterParams;
+typedef std::variant<MedianFilterParams, SpatialFilterParams, SpeckleFilterParams, TemporalFilterParams> FilterParams;
```
moratom, line 22: Probbably better to move to the filters to a common location and then use `using ...` in both StereoDepth and here

include/depthai/pipeline/datatype/DepthFiltersConfig.hpp
```diff
@@ -0,0 +1,39 @@
...
+    /**
+     * Index of the filter to be applied
+     */
+    std::int32_t filterIndex;
+
+    /**
+     * Parameters of the filter to be applied
+     */
+    FilterParams filterParams;
+
```
moratom, line 22: Any reason why we didn't go with std::vector<FilterParams> here?

include/depthai/properties/DepthFiltersProperties.hpp
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * Properties for StereoDepthFilterPipeline node
+ */
+struct SequentialDepthFiltersProperties : PropertiesSerializable<Properties, SequentialDepthFiltersProperties> {
+    /**
+     * List of filters (the type of which is determined by the filter parameters) to apply to the input frame
+     */
+    std::vector<FilterParams> filters;
+};
+
```
moratom, line 33: Let's rather use an initialConfig, to not duplicate parameters.

include/depthai/properties/DepthFiltersProperties.hpp
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * Properties for DepthConfidenceFilter node
+ */
+struct DepthConfidenceFilterProperties : PropertiesSerializable<Properties, DepthConfidenceFilterProperties> {
+    /**
+     * Threshold for the confidence filter
+     */
+    float confidenceThreshold = 0.0f;
+};
```
moratom, line 44: Same as for DepthFilter properties

src/pipeline/node/host/DepthFilters.cpp
```diff
@@ -0,0 +1,904 @@
...
+    uint8_t currFrameIdx = params->currFrameIdx;
+    uint8_t* persistenceMap = params->persistenceMap;
+    static_assert((std::is_arithmetic<T>::value), "temporal filter assumes numeric types");
+
+    const bool fp = (std::is_floating_point<T>::value);
+
+    T delta_z = static_cast<T>(delta);
+
+    auto frame = reinterpret_cast<T*>(currentFrame->data->getData().data());
+    auto _last_frame = reinterpret_cast<T*>(accumulatorFrame);
```
moratom, line 369: Let's use camelCase throughout the implementation for consistency

moratom (commented):
Thanks Jakub!


Left some comments&suggestions, generally looks good.
Let's also add an example (can be similar to `stereo_depth_from_host.py`) and tests before merging.

src/pipeline/node/host/DepthFilters.cpp
```diff
@@ -0,0 +1,904 @@
...
+
+    for(int i = 0; i < params.iterationNr; i++) {
+        if(disparityFrame->getType() == dai::ImgFrame::Type::RAW16) {
+            recursive_filter_horizontal<uint16_t>(&params);
+            recursive_filter_vertical<uint16_t>(&params);
+        } else if(disparityFrame->getType() == dai::ImgFrame::Type::RAW8) {
+            recursive_filter_horizontal<uint8_t>(&params);
+            recursive_filter_vertical<uint8_t>(&params);
+        } else {
+            throw std::runtime_error("Unsupported disparity frame type");
```
SzabolcsGergely, line 319: remove disparity

src/pipeline/node/host/DepthFilters.cpp
```diff
@@ -0,0 +1,904 @@
...
+SpeckleFilter::~SpeckleFilter() {}
+
+void SpeckleFilter::process(std::shared_ptr<dai::ImgFrame>& disparityFrame, int speckleRange, int maxDiff) {
+    int cvtype;
+    if(disparityFrame->getType() == dai::ImgFrame::Type::RAW16) {
+        cvtype = CV_16SC1;
+    } else if(disparityFrame->getType() == dai::ImgFrame::Type::RAW8) {
+        cvtype = CV_8UC1;
+    } else {
+        throw std::runtime_error("Unsupported disparity frame type. Supported types are RAW16 and RAW8.");
```
SzabolcsGergely, line 343: remove disparity

src/pipeline/node/host/DepthFilters.cpp
```diff
@@ -0,0 +1,904 @@
...
+    SpeckleFilter();
+    ~SpeckleFilter();
+    int Init();
+
+    void process(std::shared_ptr<dai::ImgFrame>& disparityFrame, int speckleRange, int maxDiff);
+};
+
+constexpr const size_t PERSISTENCY_LUT_SIZE = 256;
+
+struct TemporalFilterParams {
```
SzabolcsGergely, line 59: why not break out each implementation in their own .cpp ?

src/pipeline/node/host/DepthFilters.cpp
```diff
@@ -0,0 +1,904 @@
+#include "depthai/pipeline/node/host/DepthFilters.hpp"
```
SzabolcsGergely, line 1: i'd rather call these ImageFilters, they can be used for general purposes, if one wants to.
Only depthConfidence filter is ToF specific?
E.g. temporal filter could be used on input images to stereo depth to reduce noise. Same for median.

include/depthai/pipeline/datatype/DepthFiltersConfig.hpp
lnotspotl: The idea is as follows: Say we have filters, applied sequentially, (F1 | F2 | F3 | F4 | F5). When one wants to modify the properties any one filter, say F3, one can simply specify its index and properties without modifying or having to worry about all the other filters. This will throw an error if the param's type does not match the expected type of the filter at the given index.

include/depthai/pipeline/datatype/ImageFiltersConfig.hpp
```diff
@@ -0,0 +1,39 @@
...
+
+class ToFDepthConfidenceFilterConfig : public Buffer {
+   public:
+    ToFDepthConfidenceFilterConfig() = default;
+    virtual ~ToFDepthConfidenceFilterConfig() = default;
+
+    /**
+     * Threshold for the confidence filter
+     */
+    float confidenceThreshold;
```
moratom: We should initialize this to some default

include/depthai/properties/ImageFiltersProperties.hpp
```diff
@@ -0,0 +1,44 @@
...
+typedef std::variant<MedianFilterParams, SpatialFilterParams, SpeckleFilterParams, TemporalFilterParams> FilterParams;
+
+/**
+ * Properties for StereoDepthFilterPipeline node
+ */
+struct ImageFiltersProperties : PropertiesSerializable<Properties, ImageFiltersProperties> {
+    /**
+     * List of filters (the type of which is determined by the filter parameters) to apply to the input frame
+     */
+    std::vector<FilterParams> filters;
```
moratom: We should have an initialConfig for that instead (see ImageManip, StereoNode) to avoid duplication.

include/depthai/properties/ImageFiltersProperties.hpp
```diff
@@ -0,0 +1,44 @@
...
+DEPTHAI_SERIALIZE_EXT(ImageFiltersProperties, filters);
+
+/**
+ * Properties for DepthConfidenceFilter node
+ */
+struct ToFDepthConfidenceFilterProperties : PropertiesSerializable<Properties, ToFDepthConfidenceFilterProperties> {
+    /**
+     * Threshold for the confidence filter
+     */
+    float confidenceThreshold = 0.0f;
```
moratom: We should have an initialConfig for that instead (see ImageManip, StereoNode) to avoid duplication.

src/pipeline/node/ImageFilters.cpp
moratom, line 1: Maybe it would be worth splitting this into files per filter and removing redundant constructors and other warnings brought up by clangd.

src/pipeline/node/ImageFilters.cpp
```diff
@@ -0,0 +1,1049 @@
...
+        // Create empty output frames
+        auto filteredDepthFrame = std::make_shared<dai::ImgFrame>();
+        auto confidenceFrame = std::make_shared<dai::ImgFrame>();
+
+        // Copy metadata from input frames
+        filteredDepthFrame->setMetadata(depthFrame);
+        confidenceFrame->setMetadata(depthFrame);
+
+        // Allocate memory for output frames
+        filteredDepthFrame->data->setSize(depthFrame->getData().size() * sizeof(std::uint16_t));
```
SzabolcsGergely: rather use depthFrame->getBytesPerPixel?

bindings/python/src/pipeline/node/ImageFiltersBindings.cpp
```diff
@@ -0,0 +1,101 @@
...
+        .def_readonly("filteredDepth", &ToFDepthConfidenceFilter::filteredDepth, DOC(dai, node, ToFDepthConfidenceFilter, filteredDepth))
+        .def_readonly("confidence", &ToFDepthConfidenceFilter::confidence, DOC(dai, node, ToFDepthConfidenceFilter, confidence))
+        .def_readonly("config", &ToFDepthConfidenceFilter::config, DOC(dai, node, ToFDepthConfidenceFilter, config))
+        .def_readonly("initialConfig", &ToFDepthConfidenceFilter::initialConfig, DOC(dai, node, ToFDepthConfidenceFilter, initialConfig))
+        .def("setRunOnHost", &ToFDepthConfidenceFilter::setRunOnHost, py::arg("runOnHost"), DOC(dai, node, ToFDepthConfidenceFilter, setRunOnHost))
+        .def("runOnHost", &ToFDepthConfidenceFilter::runOnHost, DOC(dai, node, ToFDepthConfidenceFilter, runOnHost))
+        .def("setConfidenceThreshold",
+             &ToFDepthConfidenceFilter::setConfidenceThreshold,
+             py::arg("threshold"),
+             DOC(dai, node, ToFDepthConfidenceFilter, setConfidenceThreshold))
```
moratom: We had those in StereoDepth because we first had the settings global, before they were configurable through the initialConfig.

Now that we have the initialConfig from the get go, no need to have it duplicate, so I would remove `setConfidenceThreshold` directly on the node and leave it as initialConfig setter only (likewise for the getter)

bindings/python/src/pipeline/node/ImageFiltersBindings.cpp
```diff
@@ -0,0 +1,101 @@
...
+        .def("getConfidenceThreshold", &ToFDepthConfidenceFilter::getConfidenceThreshold, DOC(dai, node, ToFDepthConfidenceFilter, getConfidenceThreshold))
+        .def("build",
+             py::overload_cast<Node::Output&, Node::Output&, ImageFiltersPresetMode>(&ToFDepthConfidenceFilter::build),
+             py::arg("depth"),
+             py::arg("amplitude"),
+             py::arg("presetMode") = ImageFiltersPresetMode::DEFAULT,
+             DOC(dai, node, ToFDepthConfidenceFilter, build))
+        .def("build",
+             py::overload_cast<ImageFiltersPresetMode>(&ToFDepthConfidenceFilter::build),
+             py::arg("presetMode"),
```
moratom: Should we default to DEFAULT preset here too?

bindings/python/src/pipeline/node/ToFBindings.cpp
```diff
@@ -32,19 +33,63 @@ void bind_tof(pybind11::module& m, void* pCallstack) {
...
+            [](const ToF& self) -> const dai::node::ToFDepthConfidenceFilter& { return self.tofDepthConfidenceFilterNode; },
+            DOC(dai, node, ToF, tofDepthConfidenceFilterNode))
+        .def_property_readonly(
+            "imageFiltersNodes", [](const ToF& self) -> const dai::node::ImageFilters& { return self.imageFiltersNode; }, DOC(dai, node, ToF, imageFiltersNode))
+        .def_static("create", &ToF::create, "device"_a, DOC(dai, node, ToF, create))
+        .def("build",
+             &ToF::build,
+             "boardSocket"_a = CameraBoardSocket::AUTO,
+             "presetMode"_a = ImageFiltersPresetMode::DEFAULT,
+             "fps"_a = float(30),
```
moratom: Any reason to default on 30 here?
On Cameras we let this be FW driven (std::nullopt by default) so it can vary based on what makes sense for a given sensor.

examples/python/RVC2/ToF/tof.py
```diff
@@ -26,19 +27,24 @@ def colorizeDepth(frameDepth):
...
     return depthFrameColor
 
+
 # Create pipeline
 pipeline = dai.Pipeline()
 # Define source and output
-tof = pipeline.create(dai.node.ToF).build()
+tof = pipeline.create(dai.node.ToF).build(
+    dai.CameraBoardSocket.AUTO, dai.ImageFiltersPresetMode.DEFAULT
+)
```
moratom: Are these already the default?

examples/python/RVC2/ToF/tof.py
moratom, line 1: Tested out the example, works well!

examples/python/StereoDepth/stereo_depth_filters.py
```diff
@@ -0,0 +1,134 @@
+#!/usr/bin/env python3
```
moratom, line 1: I got
```
[2025-07-14 17:45:10.853] [ThreadedNode] [error] Node threw exception, stopping the node. Exception message: Invalid filter params. Expected MedianFilterParams, got 3
```

when running the example.

While it's a very neat example, we might potentially want to remove this one altogether or at least move it out from StereoDepth node to `Misc` since the default in StereoDepth should be to use the included filters.

include/depthai/pipeline/node/ImageFilters.hpp
```diff
@@ -0,0 +1,217 @@
...
+
+    /**
+     * Filtered frame
+     */
+    Node::Output output{*this, {"output", Node::DEFAULT_GROUP, {{{DatatypeEnum::ImgFrame, false}}}}};
+
+    /**
+     * Config to be set for a specific filter
+     */
+    Node::Input config{*this,
```
moratom: I think we have `inputConfig` on other nodes (ImageManip, StereoDepth)

include/depthai/pipeline/node/ImageFilters.hpp
```diff
@@ -0,0 +1,217 @@
...
+        virtual void setParams(const FilterParams& params) = 0;
+        virtual std::string getName() const = 0;
+        virtual ~Filter() = default;
+    };
+
+    /**
+     * Add a filter to the node pipeline. The type of filter is determined based on the type of the filter parameter.
+     * @param filter The filter to add
+     */
+    void addFilter(const FilterParams& filter);
```
moratom: Not used anywhere / no function definition?

include/depthai/pipeline/node/ImageFilters.hpp
```diff
@@ -0,0 +1,217 @@
...
+
+    /**
+     * Get the confidence threshold
+     */
+    float getConfidenceThreshold() const;
+
+    /**
+     * Set the confidence threshold
+     */
+    void setConfidenceThreshold(float threshold);
```
moratom: (same comment as in bindings, if duplicate of initialConfig we should remove it)

include/depthai/pipeline/node/ImageFilters.hpp
```diff
@@ -0,0 +1,217 @@
...
+
+    void run() override;
+
+    class Filter {
+       public:
+        virtual void process(std::shared_ptr<dai::ImgFrame>& frame) = 0;
+        virtual void setParams(const FilterParams& params) = 0;
+        virtual std::string getName() const = 0;
+        virtual ~Filter() = default;
+    };
```
moratom: I think this is only used internally, so no need to have it in the public interface (if not too painful to hide)

src/pipeline/node/ImageFilters.cpp
```diff
@@ -0,0 +1,1093 @@
...
+        } break;
+        case ImageFiltersPresetMode::HIGH_RANGE: {
+            // TODO: Add a sensible value here
+            initialConfig->confidenceThreshold = 0.0f;
+        } break;
+    }
+
+    properties.initialConfig = *initialConfig;
+}
+
```
moratom: I think we should go ahead and make the preset mode setting directly on the initial config, so it can be used in runtime too.

bindings/python/src/pipeline/node/ToFBindings.cpp
lnotspotl: Should I add this feature to the ToF node on the FW side? I have just checked and we don't have any notion of a AUTO fps value that would automatically pick the fps value based on the ToF sensor. I used the value of 30fps as the default for it had been used in the previous implementation.

include/depthai/pipeline/node/ImageFilters.hpp
lnotspotl: Moved to the unnamed namespace in `ImageFilters.cpp` [f88f7f2](https://github.com/luxonis/depthai-core/pull/1308/commits/f88f7f245f65607a46894c58222fcb109dcad176)

include/depthai/pipeline/node/ImageFilters.hpp
lnotspotl: Removed in [f88f7f2](https://github.com/luxonis/depthai-core/pull/1308/commits/f88f7f245f65607a46894c58222fcb109dcad176)

include/depthai/pipeline/node/ImageFilters.hpp
lnotspotl: Removed, I overlooked this one. Thanks! [f88f7f2](https://github.com/luxonis/depthai-core/pull/1308/commits/f88f7f245f65607a46894c58222fcb109dcad176)

include/depthai/pipeline/node/ImageFilters.hpp
lnotspotl: Renamed to `inputConfig`  in [f88f7f2](https://github.com/luxonis/depthai-core/pull/1308/commits/f88f7f245f65607a46894c58222fcb109dcad176)

examples/python/StereoDepth/stereo_depth_filters.py
lnotspotl: Fixed: [f88f7f2](https://github.com/luxonis/depthai-core/pull/1308/commits/f88f7f245f65607a46894c58222fcb109dcad176)

bindings/python/src/pipeline/node/ImageFiltersBindings.cpp
lnotspotl: Removed in [f88f7f2](https://github.com/luxonis/depthai-core/pull/1308/commits/f88f7f245f65607a46894c58222fcb109dcad176)

src/pipeline/node/ImageFilters.cpp
lnotspotl: I am not sure I understand what you mean by that. Should the user be able to pass in a reference of a config and this method would then make the modifications on that reference?

examples/python/RVC2/ToF/tof.py
lnotspotl: yes, these are the default values.

bindings/python/src/pipeline/node/ToFBindings.cpp
moratom: I think AUTO and then making the default on FW makes sense, yeah.
 
 Mainly for future proofing the API if we'll have a different ToF sensor where a different default will make sense.

src/pipeline/node/ImageFilters.cpp
moratom: Sorry, I meant to have the the `setDefaultProfilePreset` method (maybe could be renamed to just `setProfilePreset`) directly on the config.

So `ImageFilters.initialConfig.setProfilePreset()`

lnotspotl:
@moratom @SzabolcsGergely Me and @DavidZah won't be adding any more features so this PR is feature-complete. Let's now focus on making sure that these features nicely fit into `depthai-core`

examples/python/RVC2/ToF/tof_filter_demo.py
moratom, line 1: @lnotspotl thoughts on moving this to oak-examples instead? It's quite extensive for the context of examples here.

moratom (approved):
Looks good to merge from my side.

lnotspotl:
Fixed a minor error. The new tof demo example shall be merged into `oak-examples`. We have a PR open for that: https://github.com/luxonis/oak-examples/pull/723

PR #1309, opened 2025-04-30
cmake: Install shared libs

phodina:
## Purpose
Install depthai shared libraries and other build artifacts into the system using CMake

Also sets correctly the RPATH otherwise it would point to the build dir.

## Testing & Validation
Tested on Linux and MacOSX


phodina:
Set and propagate RPATH settings across targets and Python components

- Ensured BUILD_WITH_INSTALL_RPATH is enabled for correct runtime linking.
- Added proper RPATH handling to Python bindings and test modules for
  macOS and Linux to ensure correct runtime library resolution.

phodina:
@moratom so removed the code regarding the dependencies which is no longer needed. But rest is needed for system install and correct RPATH

PR #1310, opened 2025-05-05
Bom fixes

danilo-pejovic:
Improving BOM tests and adding disaptching said tests to the end of python workflow. Tests will run in parallel and should be fair bit quicker, also will test depthai-experiments


.github/workflows/python-main.yml
```diff
@@ -10,6 +10,7 @@ on:
   push:
     branches:
     - v3_develop*
+    - bom_fixes
```
moratom, line 13: Needed still?

moratom (approved):
Thanks - I'd remove to run this on PRs to bom_fixes, since the branch will be deleted after it's merged, but otherwise LGTM.

PR #1305, opened 2025-04-23
Zoo features, agent team

lnotspotl:
## Purpose
Add `zoo_helper`-related features requested by the agent team.

This PR should only be merged after #1304

PR #1313, opened 2025-05-07
Feature/zoo manager filesystem lock

lnotspotl:
## Purpose
Add filesystem lock to prevent race-conditions when loading the same model from two different processes at the same time.

This PR should only be merged after https://github.com/luxonis/depthai-core/pull/1305

Clickup task: https://app.clickup.com/t/86c34ugy5


moratom (commented):
Thanks!

We should add tests for this and make sure everything works as expected on Windows  MacOS & Linux.

Fcntl locks according to the documentation on the linux man-pages are not POSIX, so not sure if MacOS works?

lnotspotl:
@moratom Good point there, I did not realize part of man-page can be POSIX-compliant and part of it can be non-POSIX-compliant.

I've checked the macOS dev page [here](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man2/fcntl.2.html) and it seems that there is no difference between how BSD and POSIX-compliant systems implement file locking. I will test everything in a VM regardless, both on windows and macOS.

lnotspotl:
All tests have been performed, filesystem locks work on operating systems supported by `depthai-core`, i.e. Windows10/11, Linux (tested on Ubuntu), macOS13/14. This PR is ready to be reviewed one last time and merged subsequently.

tests/src/onhost_tests/utility/platform_test.cpp
```diff
@@ -0,0 +1,103 @@
...
+    pid_t pid2 = fork();
+    if(pid2 == 0) {
+        // Child process 2
+        auto lock = FileLock::lock(tempFile.string());
+        REQUIRE(lock->holding());
+        std::cout << "Process 2 holding lock" << lock->holding() << std::endl;
+        std::this_thread::sleep_for(std::chrono::seconds(1));
+        lock->unlock();
+        exit(0);
+    }
```
moratom: Can we use std::thread here to make this easier to read? (and I assume it should afterwards pass on Windows too)

src/modelzoo/Zoo.cpp
```diff
@@ -27,6 +27,11 @@ static std::string MODEL_ZOO_DOWNLOAD_ENDPOINT = "https://easyml.cloud.luxonis.c
 static std::string MODEL_ZOO_DEFAULT_CACHE_PATH = ".depthai_cached_models";  // hidden cache folder
 static std::string MODEL_ZOO_DEFAULT_MODELS_PATH = "depthai_models";         // folder
 
+// Take paths and combine them: this function is OS-agnostic
+static std::string combinePaths(const std::string& first, const std::string& second) {
+    return (std::filesystem::path(first) / std::filesystem::path(second)).string();
+}
```
moratom: Let's merge the filesystem PR first, so we can skip this

src/modelzoo/Zoo.cpp
```diff
@@ -57,6 +62,17 @@ class ZooManager {
...
             this->cacheDirectory = utility::getEnvAs<std::string>("DEPTHAI_ZOO_CACHE_PATH", dai::modelzoo::getDefaultCachePath(), false);
         }
+
+        if(this->cacheDirectory.empty()) {
+            throw std::runtime_error("Cache directory is not set");
+        }
+
+        // Lock the cache directory
+        createFolder(".locks");
+        const std::string modelLockFilePath = combinePaths(combinePaths(this->cacheDirectory, ".locks"), getModelCacheFolderName() + ".lock");
```
moratom: should get switched to fs::path after the merge

moratom (commented):
Went over quickly, looks good, but we should first merge the std::filesystem PR and update this one to reflect using fs::path everywhere where applicable.

tests/src/onhost_tests/utility/platform_test.cpp
lnotspotl: Thanks for pointing this out. Initially, I though making this cross-process safe was enough but then realized that making this thread-safe in addition to it being cross-process safe would be way better. I added a test that tests both file locking on a thread as well as process level.

That being said, creating a new process on Windows is not as easy as calling a single function, that's why I decided to disable it on Windows. Not really sure how one can spawn a new process with the same memory content as the parent process on that OS. ¯\\_(ツ)_/¯

moratom: Ah right, I made a quick comment since I quickly checked and thought we're using the POSIX API directly to create threads.

For processes, I think we could get around this by using the subprocess library (header only in `tests/include/subprocess.hpp`), but we'd have to compile a separate independant binary that would get executed by it ,it's not like fork().


I think making it thread safe too might be good anyways for some exotic multi-pipeline usecases in the same process.

lnotspotl: Let's create a  dummy executable then that will try to access a certain file.

Added: 68122325b370b64d0ff891767c236be2b860dde2

moratom (approved):
Thanks!

LGTM if tests pass.

PR #1314, opened 2025-05-07
ImgDetection text label

asahtik:
## Purpose
ImgDetections now also hold text label (name) if class names are available

## Specification

Add name parameter to ImgDetection

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable


include/depthai/pipeline/datatype/ImgDetections.hpp
```diff
@@ -11,14 +11,15 @@ namespace dai {
 
 struct ImgDetection {
     uint32_t label = 0;
+    std::string name = "";
```
moratom, line 14: nit: no need for an initializer here, std::string is default initialized

PR #1316, opened 2025-05-11
Fix method responsible for creating python subnodes (createSubnode)

lnotspotl:
## Purpose
Previously, a sub-node added to a node via the `createSubnode` python method would not be registered by the parent node. As a result of that, when deleting the parent node, the subnode would still remain in the pipeline. This PR fixes that. 

The way we go about doing it is by creating a node using the already established pipeline create method, we then remove that node from the pipeline and finally add it as a child node to the parent node. This is necessary because otherwise `getAllNodes` contains the same node twice, thereby breaking many other methods.

Clickup task: https://app.clickup.com/t/86c3cb67q


moratom (approved):
Thanks!

PR #1323, opened 2025-05-18
Lnotspotl/internal xlink

lnotspotl:
This PR moves `XLink` nodes and `XLink` properties to a newly introduced `dai::internal` namespace. This shall be a namespace with necessary public implementation that is not meant to be used by the user due to its potentially unstable API. This PR should be reviewed alongside the other opened FW PRs on Gitlab. Changes necessary were minimal.


themarpe:
Great to see, having this `internal` scaffolding in place for non public/non stable API

bindings/python/CMakeLists.txt
```diff
@@ -71,8 +71,8 @@ set(SOURCE_LIST
 
     src/pipeline/node/NodeBindings.cpp
 
-    src/pipeline/node/XLinkInBindings.cpp
-    src/pipeline/node/XLinkOutBindings.cpp
+    src/pipeline/node/internal/XLinkInBindings.cpp
```
moratom, line 74: I don't think we need the bindings for these two nodes anymore.

moratom (approved):
LGTM!
Two comments:
* We should remove the `v2_examples` (they are not in a working state anyways)
* We don't need the bindings for XLink nodes anymore

bindings/python/src/pipeline/node/Common.hpp
moratom, line 1: Not sure if we'd still want this? We can keep it in though, might come in handy.

tests/src/ondevice_tests/encoded_frame_test.cpp
moratom, line 1: The example doesn't compile/run anymore, so we can just add a TODO to port to v3.

tests/src/ondevice_tests/pipeline_test.cpp
moratom, line 1: Better to use some other node for that purpuse, no need to use XLinks

tests/src/stability_stress_test.cpp
moratom, line 1: Better to add a TODO that it needs porting to v3

moratom (approved):
I would add TODOs to tests that aren't ported rather than adding the internal namespace to them, but not critical.

lnotspotl:
Will wait for the CI to finish and then I'll perform the merge

PR #1325, opened 2025-05-19
Add C++ examples

lnotspotl:
This PR adds cpp examples, essentially copies of the python examples. I've changed the `CMakeLists.txt` file for the cpp example folders to mirror the python examples folder tree structure.

Current state of completion: https://app.clickup.com/t/86c2t5rru

There still are a couple of examples missing, though the majority of the important simple examples are there. I did not want the task to be a block and thus we're ready to merge. The rest of the examples will be added asap.


moratom (approved):
Thanks!

:ship:

lnotspotl:
:shipit:

PR #1329, opened 2025-05-21
fix include directory for tests

phodina:
Use include options in cmake rather than the "../"


lnotspotl:
Can use this for some of the other tests that include headers from `src`?

SzabolcsGergely (approved):

PR #1335, opened 2025-05-27
Fix example build on Windows and add examples to the testing suite

lnotspotl:
## Purpose

This PR fixes building C++ examples on Windows. Moreover, this PR adds the majority of the newly introduced C++ examples into our testing suite.


lnotspotl:
As part of this PR, I've also added a new `cmake` function useful for relaxing the requirements on the output of examples when being tested. More specifically, when an example is relaxed, any `[warning]` logs will be considered as expected and consequently the relaxed test will pass, even in the presence of `[warning]` logs.

PR #1336, opened 2025-05-27
ci: Add job for Nix build on x86_64 linux

phodina:
Just run the CI for Nixpkgs build of `depthai-core` on Linux `x86_64` system

- [ ] Add support for Darwin (`x86_64` and `aarch64`)
- [ ] Add support for Linux `aarch64`

- [ ] Add Cachix push at the end of the build so the derivations are cached and easily reused

- [ ] Run tests against real HW on the HIL testbeds

### DIFF
--- a/.github/workflows/python-main.yml
+++ b/.github/workflows/python-main.yml
@@ -175,6 +175,40 @@ jobs:
         run: |
           cmake --build build --target pytest --config Release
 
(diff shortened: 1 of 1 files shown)


phodina:
https://github.com/luxonis/depthai-core/pull/1337 use name to trigger the workflow without major changes to CI

PR #1337, opened 2025-05-27
ci: Add job for Nix build on x86_64 linux

phodina:
Just run the CI for Nixpkgs build of `depthai-core` on Linux `x86_64` system

- [ ] Add support for Darwin (`x86_64` and `aarch64`)
- [ ] Add support for Linux `aarch64`

- [ ] Add Cachix push at the end of the build so the derivations are cached and easily reused

- [ ] Run tests against real HW on the HIL testbeds

### DIFF
--- a/.github/workflows/python-main.yml
+++ b/.github/workflows/python-main.yml
@@ -175,6 +175,40 @@ jobs:
         run: |
           cmake --build build --target pytest --config Release
 
+# Build DepthAI using Nix
+jobs:
+  nix-build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+        with:
+          submodules: true
+
+      - name: Install Nix
+        uses: cachix/install-nix-action@v24
+
+
+      - name: Build using custom nixpkgs with depthai-core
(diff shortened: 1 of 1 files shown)


.github/workflows/main.workflow.yml
```diff
@@ -95,6 +95,42 @@ jobs:
...
+        run: |
+          nix-build --expr 'let pkgs = import (fetchTarball "https://github.com/phodina/nixpkgs/archive/refs/heads/depthaiv3_alpha15_draft.zip") {}; in pkgs.depthai-core'
+
+      - name: Run tests inside nix-shell
+        run: |
+          nix-shell -E '
+            with import (fetchTarball "https://github.com/phodina/nixpkgs/archive/refs/heads/depthaiv3_alpha15_draft.zip") {};
+            mkShell {
+              buildInputs = [ depthai-core ];
+              shellHook = "pytest bindings/python/tests";
```
moratom, line 121: @phodina does this always build the latest and only pick up the dependencies from your `alpha15_draft`?

We've updated the `vcpkg` baseline, so these might not build anymore as a few of the dependencies made braking changes.

PR #1358, opened 2025-06-11
Replace dai::Path with std::filesystem::path

lnotspotl:
Since the C++ minimum version required by depthai-core is C++17, we have decided to replace the old `dai::Path` with `std::filesystem::path`, a new path object introduced in the C++17 standard. 

On top of that, this PR tries to represent all filesystem paths used internally as C++ paths, not strings, to make depthai-core a little more cross-platform and handle paths with exotic characters in them, including š, č and ř.


src/common/ModelType.cpp
```diff
@@ -1,27 +1,33 @@
...
-ModelType readModelType(const std::string& modelPath) {
+ModelType readModelType(const std::filesystem::path& modelPath) {
     auto endsWith = [](const std::string& path, const std::string& ending) {
         if(ending.size() > path.size()) return false;
         return std::equal(ending.rbegin(), ending.rend(), path.rbegin());
     };
 
+    // Safely turn path into a string
+    std::string path = fmt::format("{}", modelPath);
+
```
moratom: How does this differ to
`modelPath.string();` ?

src/device/DeviceBootloader.cpp
```diff
@@ -22,6 +22,7 @@
 // libraries
 #include "XLink/XLink.h"
 #include "spdlog/fmt/chrono.h"
+#include "spdlog/fmt/std.h"
```
moratom: Used?

src/modelzoo/Zoo.cpp
```diff
@@ -22,10 +29,12 @@
...
 namespace dai {
 
+namespace fs = std::filesystem;
+
 static std::string MODEL_ZOO_HEALTH_ENDPOINT = "https://easyml.cloud.luxonis.com/models/api/v1/health/";
 static std::string MODEL_ZOO_DOWNLOAD_ENDPOINT = "https://easyml.cloud.luxonis.com/models/api/v1/models/download";
-static std::string MODEL_ZOO_DEFAULT_CACHE_PATH = ".depthai_cached_models";  // hidden cache folder
-static std::string MODEL_ZOO_DEFAULT_MODELS_PATH = "depthai_models";         // folder
+static fs::path MODEL_ZOO_DEFAULT_CACHE_PATH = ".depthai_cached_models";  // hidden cache folder
+static fs::path MODEL_ZOO_DEFAULT_MODELS_PATH = "depthai_models";         // folder
```
moratom:
```suggestion
const static std::string MODEL_ZOO_HEALTH_ENDPOINT = "https://easyml.cloud.luxonis.com/models/api/v1/health/";
const static std::string MODEL_ZOO_DOWNLOAD_ENDPOINT = "https://easyml.cloud.luxonis.com/models/api/v1/models/download";
const static fs::path MODEL_ZOO_DEFAULT_CACHE_PATH = ".depthai_cached_models";  // hidden cache folder
const static fs::path MODEL_ZOO_DEFAULT_MODELS_PATH = "depthai_models";         // folder
```

src/modelzoo/Zoo.cpp
```diff
@@ -386,20 +393,20 @@ void ZooManager::downloadModel(const nlohmann::json& responseJson, std::unique_p
...
-    std::vector<std::string> folderFiles = getFilesInFolder(cacheFolder);
+    std::vector<fs::path> folderFiles = getFilesInFolder(cacheFolder);
 
     // Make sure there are files in the folder
-    if(folderFiles.empty()) throw std::runtime_error("No files found in cache folder " + cacheFolder);
+    if(folderFiles.empty()) throw std::runtime_error(fmt::format("No files found in cache folder {}", cacheFolder));
 
     // Return absolute path to the first file found
-    return std::filesystem::absolute(folderFiles[0]).string();
+    return fs::path(std::filesystem::absolute(folderFiles[0]));
```
moratom:
```suggestion
    return std::filesystem::absolute(folderFiles[0]);
```

src/modelzoo/Zoo.cpp
```diff
@@ -718,43 +718,53 @@ bool downloadModelsFromZoo(const std::string& path, const std::string& cacheDire
...
+                yamlPath,
+                ymlPath,
+                name));
         }
         return path;
     }
 
     // We treat the name either as a relative path or an absolute path
-    return name;
+    return fs::path(name);
```
moratom:
```suggestion
    return name;
```

src/nn_archive/NNArchiveBlob.cpp
moratom, line 1: The file not used anymore, we can remove it in this PR IMO

src/opencv/HolisticRecordReplay.cpp
moratom, line 1: Tested that it still works as expected?

src/utility/ArchiveUtil.cpp
```diff
@@ -67,9 +67,13 @@ ArchiveUtil::ArchiveUtil(const std::vector<uint8_t>& data, NNArchiveEntry::Compr
...
 
-ArchiveUtil::ArchiveUtil(const dai::Path& filepath, NNArchiveEntry::Compression format) {
+ArchiveUtil::ArchiveUtil(const std::filesystem::path& filepath, NNArchiveEntry::Compression format) {
     init(format);
-    const auto res = archive_read_open_filename(aPtr, filepath.string().c_str(), 10240);
+#if defined(_WIN32) && defined(_MSC_VER)
+    const auto res = archive_read_open_filename_w(aPtr, filepath.c_str(), 10240);
+#else
+    const auto res = archive_read_open_filename(aPtr, filepath.c_str(), 10240);
+#endif
```
moratom, line 76: The old way did not work?

src/common/ModelType.cpp
```diff
@@ -1,27 +1,33 @@
...
-    if(endsWith(modelPath, ".dlc")) return ModelType::DLC;
+    if(endsWith(path, ".dlc")) return ModelType::DLC;
 
     // == NNArchive ==
-    if(endsWith(modelPath, ".tar")) return ModelType::NNARCHIVE;
-    if(endsWith(modelPath, ".tar.gz")) return ModelType::NNARCHIVE;
-    if(endsWith(modelPath, ".tar.xz")) return ModelType::NNARCHIVE;
+    if(endsWith(path, ".tar")) return ModelType::NNARCHIVE;
+    if(endsWith(path, ".tar.gz")) return ModelType::NNARCHIVE;
+    if(endsWith(path, ".tar.xz")) return ModelType::NNARCHIVE;
```
moratom: Can we use .extension() here instead perhaps?

src/nn_archive/NNArchive.cpp
```diff
@@ -183,25 +183,28 @@ std::vector<dai::Platform> NNArchive::getSupportedPlatforms() const {
...
         if(model.device == OpenVINO::Device::VPUX) {
             return {Platform::RVC3};
         }
         if(model.device == OpenVINO::Device::VPU) {
             return {Platform::RVC2};
         }
+
         // Should never get here
         return {};
     }
```
moratom, line 214: Can we use .extension() here instead perhaps?

src/utility/Compression.cpp
```diff
@@ -77,7 +77,11 @@ void tarFiles(const std::string& path, const std::vector<std::string>& files, co
 
     a = archive_write_new();
     archive_write_set_format_pax_restricted(a);
+#ifdef _WIN32
+    archive_write_open_filename_w(a, path.c_str());
+#else
     archive_write_open_filename(a, path.c_str());
+#endif
```
moratom: The previous implementation didn't work correctly on Windows?

moratom (commented):
Thanks! More LOC removed than added is always nice to see. 

Left some nitpicks.

src/common/ModelType.cpp
lnotspotl: That is something I tried the first time around when I was changing the path datatype from `std::string` to `std::filesystem::path`. The downside of the `extension` method is that it only returns the very last "extension", i.e. for `/hello/world/test.tar.gz` it only returns `".gz"` and not `"tar.gz"`


I am open to switch to using `.extension()`, it would simplify things a bit, though we would loose a bit of control, which tbh I am not sure whether we actually need. :D

src/nn_archive/NNArchive.cpp
lnotspotl: See above

src/utility/ArchiveUtil.cpp
lnotspotl: It did the majority of the times. Before introducing the above change, we would first turn a path into a string and then generate a C-equivalent string pointer. However, the problem would arise when working with exotic paths such as


```
C:/user/mača/depthai_models/model.tar.gz
```

Turning this into a string generates (I made it up but the point is clear)

```
C:/user/ma⇅⇅a/depthai_models/model.tar.gz
```

`std::filesystem::path::c_str` [returns](https://en.cppreference.com/w/cpp/filesystem/path/native.html) `const char *` on Linux/MacOS and `const wchar_t *` on Windows which solves the above issue, albeit at the cost of a little more complicated pre-processor logic.

src/utility/Compression.cpp
lnotspotl: Same as https://github.com/luxonis/depthai-core/pull/1358#discussion_r2150382333

src/modelzoo/Zoo.cpp
lnotspotl: I guess the letter capitalization for these makes it a bit confusing and seems as though they were constants, however, all the way down in `Zoo.cpp` we have setters for these, in case somebody wants to change the defaults. 

I think we should change the variable names to camel case

src/common/ModelType.cpp
lnotspotl: It does not. removed 4d49a271b3138a3d2931e4eeb47fd7fd885e8cec

src/device/DeviceBootloader.cpp
lnotspotl: Nope, removed 4d49a271b3138a3d2931e4eeb47fd7fd885e8cec

src/modelzoo/Zoo.cpp
moratom: Oh, in that case let's switch yeah.

PR #1359, opened 2025-06-12
Add docker tests

danilo-pejovic:
Adding docker tests for hil to reduce usage


tests/src/ondevice_tests/manual_focus_test.cpp
```diff
@@ -5,6 +5,22 @@
 TEST_CASE("Test manual focus") {
     dai::Pipeline pipeline;
     auto fps = 30.0f;
+    auto features = device->getConnectedCameraFeatures();
```
moratom: This doesn't compile, device is first used here.

tests/Dockerfile
```diff
@@ -0,0 +1,109 @@
...
+        echo "Using vanilla flavor (no toolchain)"; \
+    fi && \
+    cmake -S . -B build \
+        -D CMAKE_BUILD_TYPE=Release \
+        -D DEPTHAI_VCPKG_INTERNAL_ONLY=OFF \
+        -D DEPTHAI_BUILD_EXAMPLES=ON \
+        $(if [ "$PULL_REQUEST" = "false" ]; then echo "-D DEPTHAI_TEST_EXAMPLES=ON \
+        -D DEPTHAI_BUILD_PYTHON=ON \
+        -D DEPTHAI_PYTHON_TEST_EXAMPLES=ON \
+        -D DEPTHAI_PYTHON_ENABLE_EXAMPLES=ON"; fi)
```
moratom: I think we also need DEPTHAI_BUILD_TESTS here right?

moratom (commented):
Looks good, I just left some comments to address.

moratom (commented):
I tried to run the CI outside of a PR and Windows examples are not run, other than that LGTM to merge.

PR #1363, opened 2025-06-23
Camera undistort for RVC4

asahtik:
## Purpose

Implements undistortion on RVC4 (Camera / ImageManip)

## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

### DIFF
--- a/include/depthai/utility/ImageManipImpl.hpp
+++ b/include/depthai/utility/ImageManipImpl.hpp
@@ -3226,7 +3226,8 @@ void printSpecs(spdlog::async_logger& logger, FrameSpecs specs);
 template <template <typename T> typename ImageManipBuffer, typename ImageManipData>
 void WarpH<ImageManipBuffer, ImageManipData>::apply(const std::shared_ptr<ImageManipData> src, std::shared_ptr<ImageManipData> dst) {
     auto undistortDst = this->isIdentityWarp() || this->undistortOneShot ? dst : auxFrame;
-    auto undistortSpecs = this->isIdentityWarp() || this->undistortOneShot ? this->dstSpecs : getDstFrameSpecs(this->srcSpecs.width, this->srcSpecs.height, this->type);
+    auto undistortSpecs =
+        this->isIdentityWarp() || this->undistortOneShot ? this->dstSpecs : getDstFrameSpecs(this->srcSpecs.width, this->srcSpecs.height, this->type);
     auto warpSrc = this->enableUndistort ? auxFrame : src;
     auto warpSrcSpecs = this->enableUndistort ? undistortSpecs : this->srcSpecs;
     // Apply transformation multiple times depending on the image format
--- a/tests/src/ondevice_tests/xlink_tests.cpp
+++ b/tests/src/ondevice_tests/xlink_tests.cpp
@@ -1,10 +1,9 @@
 #include <catch2/catch_all.hpp>
 #include <catch2/catch_test_macros.hpp>
+#include <chrono>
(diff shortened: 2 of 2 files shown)


tests/src/ondevice_tests/pipeline/node/image_manip_test.cpp
```diff
@@ -199,3 +199,181 @@ TEST_CASE("Multiple image manips") {
...
+    runManipTests(dai::ImgFrame::Type::NV12, true, {-7.56764030456543, 18.97133445739746, 0.0006435539107769728, -5.642612813971937e-05, 6.156050682067871, -7.587080001831055, 19.094820022583008, 5.732314109802246, 0.0, 0.0, 0.0, 0.0, -0.0009434317471459508, 0.002672438742592931});
+}
+
+TEST_CASE("ImageManip GRAY8 undistort") {
+    runManipTests(dai::ImgFrame::Type::GRAY8, true, {-7.56764030456543, 18.97133445739746, 0.0006435539107769728, -5.642612813971937e-05, 6.156050682067871, -7.587080001831055, 19.094820022583008, 5.732314109802246, 0.0, 0.0, 0.0, 0.0, -0.0009434317471459508, 0.002672438742592931});
+}
+
+TEST_CASE("ImageManip RGB888i undistort") {
+    runManipTests(dai::ImgFrame::Type::RGB888i, true, {-7.56764030456543, 18.97133445739746, 0.0006435539107769728, -5.642612813971937e-05, 6.156050682067871, -7.587080001831055, 19.094820022583008, 5.732314109802246, 0.0, 0.0, 0.0, 0.0, -0.0009434317471459508, 0.002672438742592931});
+}
```
moratom: This doesn't actually test if undistortion is working? I think we should add a distorted and an undistorted image and then compare byte by byte.

src/utility/ImageManipImpl.cpp
```diff
@@ -915,3 +916,29 @@ dai::RotatedRect dai::impl::getRotatedRectFromPoints(const std::vector<std::arra
...
+        throw std::runtime_error(fmt::format("UndistortImpl: Output size does not match the expected size (got {}x{}, expected {}x{} or {}x{})",
+                                             dst.size().width,
+                                             dst.size().height,
+                                             width,
+                                             height,
+                                             width / 2,
+                                             height / 2));
+    }
+}
+#endif
```
moratom: I think we should get the camera matrix for initUndistortRectifyMap from the ImageManip itself (and alpha should not be used) to keep the behavior predictable.

So right now however ImageManip node is configured the ImgTransformations and internally as well we should be able to tell what are the target image intrinsics and then use those.

moratom (commented):
Thanks!

Left some comments/questions.
We should also move the camera_undistort.py example from the RVC2 folder before merging.

include/depthai/utility/ImageManipImpl.hpp
```diff
@@ -3218,18 +3421,30 @@ void WarpH<ImageManipBuffer, ImageManipData>::apply(const std::shared_ptr<ImageM
...
+                          warpSrcSpecs.height,
+                          warpSrcSpecs.p1Stride,
+                          this->dstSpecs.width,
+                          this->dstSpecs.height,
+                          this->dstSpecs.p1Stride,
+                          1,
+                          1,
+                          this->matrix,
+                          {this->backgroundColor[0]});
+            }
```
moratom: Does this work?

moratom (approved):
LGTM, thanks!

PR #1371, opened 2025-06-30
Unique input output names

Serafadam:
## Purpose

Creating inputs/outputs without explicitly giving them name can result in crashes 
## Specification

Adding a method to append unique ID to input/output and input/output queues so that the names won't clash

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable


src/pipeline/Node.cpp
```diff
@@ -162,7 +171,8 @@ std::shared_ptr<dai::MessageQueue> Node::Output::createOutputQueue(unsigned int
     if(pipelinePtr.isBuilt()) {
         throw std::runtime_error("Cannot create queue after pipeline is built");
     }
-    auto queue = std::make_shared<MessageQueue>(maxSize, blocking);
+    auto name = getParent().createUniqueOutputName("output_queue");
+    auto queue = std::make_shared<MessageQueue>(name, maxSize, blocking);
     link(queue);
```
moratom: Do we need to name the queue here?

src/pipeline/Node.cpp
```diff
@@ -178,7 +188,8 @@ std::shared_ptr<InputQueue> Node::Input::createInputQueue(unsigned int maxSize,
 
     // Construct a new InputQueue interface
     // Cannot use make_shared as the InputQueue's constructor is private - only send method is exposed
-    auto inputQueuePtr = std::shared_ptr<InputQueue>(new InputQueue(maxSize, blocking));
+    auto name = getParent().createUniqueInputName("input_queue");
+    auto inputQueuePtr = std::shared_ptr<InputQueue>(new InputQueue(name, maxSize, blocking));
```
moratom: Why is the name needed?

tests/src/ondevice_tests/input_output_naming_test.cpp
moratom, line 1: I would also test the usecase for which was the reason we added this (adding multiple inputs in a host node in the usual fashion)

moratom (commented):
Thanks!

We should make style pass and test that host nodes can have more than one unnammed inputs/outputs (a test that fails before this but passes after).

src/pipeline/Node.cpp
Serafadam: Shouldn't we also have unique names for queues as well?

moratom: I don't think it's needed, the name there is generally not used

PR #1388, opened 2025-07-16
Zoo: check for write permissions before updating a model

lnotspotl:
## Purpose
Without this check, the zoo manager tries to update a model, create a lock and in case of a protected path (such as `/usr/lib`) this leads to an insufficient permission error.

With this check in place, we do not try to update the model and simply return the already cached one or in the edge case of there being no cached model, we error out.

### DIFF
--- a/src/modelzoo/Zoo.cpp
+++ b/src/modelzoo/Zoo.cpp
@@ -72,12 +72,42 @@ class ZooManager {
             throw std::runtime_error("Cache directory is not set");
         }
 
-        // Lock the cache directory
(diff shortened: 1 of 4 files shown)


src/modelzoo/Zoo.cpp
```diff
@@ -171,7 +201,6 @@ class ZooManager {
      */
     static bool connectionToZooAvailable();
 
-   private:
```
moratom, line 174: Intentionally removed?

moratom (approved):
Thanks!

PR #1390, opened 2025-07-18
Depthai USB Support

TheMutta:
## Purpose
This branch adds the ability to Depthai to communicate with Gate via USB through XLink

## Specification
All the API calls in the DepthaiGate class use USB communication if TCP/IP is not available
None / not applicable

All the dependencies are already integrated in XLink

## Deployment Plan
The changes don't affect the normal functioning via TCP/IP and don't break existing features, they can be rolled out easily.

## Testing & Validation
The branch was tested by using all the API calls, which all completed successfully.

### DIFF
--- a/cmake/depthaiDependencies.cmake
+++ b/cmake/depthaiDependencies.cmake
@@ -176,7 +176,7 @@ else()
     FetchContent_Declare(
         XLink
         GIT_REPOSITORY https://github.com/luxonis/XLink.git
-        GIT_TAG        8423b90465d9f91d8c3fd108bb8c18a7b27d8a3e
+        GIT_TAG        fdc7ca6699431bab598e366057274c9e215967c0
     )
 
     FetchContent_MakeAvailable(
--- a/cmake/depthaiDependencies.cmake
+++ b/cmake/depthaiDependencies.cmake
@@ -176,7 +176,7 @@ else()
     FetchContent_Declare(
         XLink
         GIT_REPOSITORY https://github.com/luxonis/XLink.git
-        GIT_TAG        fdc7ca6699431bab598e366057274c9e215967c0
+        GIT_TAG        e519b8698d81722b05b389551e1038006198ef77
     )
 
     FetchContent_MakeAvailable(
--- a/cmake/depthaiDependencies.cmake
+++ b/cmake/depthaiDependencies.cmake
@@ -176,7 +176,7 @@ else()
     FetchContent_Declare(
         XLink
         GIT_REPOSITORY https://github.com/luxonis/XLink.git
-        GIT_TAG        1f1ade046e25424958937a23cedecb165fe0fafa
+        GIT_TAG        0e2e7d11f0f128c1363b1ac208fb09a363afb9d8
     )
 
     FetchContent_MakeAvailable(
--- a/cmake/depthaiDependencies.cmake
+++ b/cmake/depthaiDependencies.cmake
@@ -176,7 +176,7 @@ else()
     FetchContent_Declare(
         XLink
         GIT_REPOSITORY https://github.com/luxonis/XLink.git
-        GIT_TAG        0e2e7d11f0f128c1363b1ac208fb09a363afb9d8
+        GIT_TAG        8423b90465d9f91d8c3fd108bb8c18a7b27d8a3e
     )
 
     FetchContent_MakeAvailable(
--- a/cmake/depthaiDependencies.cmake
+++ b/cmake/depthaiDependencies.cmake
@@ -176,7 +176,7 @@ else()
     FetchContent_Declare(
         XLink
         GIT_REPOSITORY https://github.com/luxonis/XLink.git
-        GIT_TAG        1958f31ab14dff84b0edba35db524ebecd00b1f8
+        GIT_TAG        1f1ade046e25424958937a23cedecb165fe0fafa
     )
 
     FetchContent_MakeAvailable(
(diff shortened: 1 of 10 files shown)


TheMutta:
bugbot run

src/device/DeviceGate.cpp
```diff
@@ -56,12 +78,33 @@ DeviceGate::DeviceGate(const DeviceInfo& deviceInfo) : deviceInfo(deviceInfo) {
...
+	request.RequestNum = IS_OKAY;
+	request.RequestSize = 0;
+	XLinkPlatformGateWrite(&request, sizeof(USBRequest_t));
+
+	XLinkPlatformGateRead(&request, sizeof(request));
+	if(request.RequestNum == RESPONSE_ERROR) {
+		return false;
+	}
+
+	char *respBuffer = new char[request.RequestSize + 1];
```
themarpe: Lets use C++ here - eg a vector /w reserve or resize and pass its "data()" to XLink to not have to manually new & delete

cmake/depthaiDependencies.cmake
```diff
@@ -176,7 +176,7 @@ else()
     FetchContent_Declare(
         XLink
         GIT_REPOSITORY https://github.com/luxonis/XLink.git
-        GIT_TAG        87785828fabdb1718760bb0a044405d5bbfbb3a2
+        GIT_TAG        d46c049e30d2a56f7bbec2e25e7e2fa5c09ec6d0
```
moratom: Can we cross link the XLink PR in this PRs description? Thanks!

src/device/DeviceGate.cpp
```diff
@@ -44,7 +66,7 @@ DeviceGate::DeviceGate(const DeviceInfo& deviceInfo) : deviceInfo(deviceInfo) {
     if(deviceInfo.platform != X_LINK_RVC3 && deviceInfo.platform != X_LINK_RVC4) {
         throw std::invalid_argument("Gate only supports RVC3 and RVC4 platforms");
     }
-    this->platform = deviceInfo.platform;
+    this->platform =deviceInfo.platform;
```
moratom: whitespace change

src/device/DeviceGate.cpp
```diff
@@ -27,6 +28,27 @@ const std::string API_ROOT{"/api/v1"};
...
+	IS_OKAY,
+	GET_VERSION,
+	GET_STATE,
+	GET_FILE,
+};
+
+struct USBRequest_t {
+    uint16_t RequestNum;
+    uint32_t RequestSize;
+}__attribute__((packed));
```
moratom: is this cross platform?

moratom (commented):
Thanks @TheMutta!

I'm thinking if I'd be safer here to define an interface (pure virtual class) and then do two implementations, so we have compile-time safety to a degree that we implemented all current (and potential future) calls to gate.
What do you think?

src/device/DeviceGate.cpp
themarpe: My 2c, can just skip packed by making both uint32_t or swapping them around

TheMutta: It's definitely cross platform

shared/depthai-shared
moratom, line 1: Needs to be removed

src/device/DeviceGate.cpp
```diff
@@ -52,29 +73,91 @@ DeviceGate::DeviceGate(const DeviceInfo& deviceInfo) : deviceInfo(deviceInfo) {
...
+    return nlohmann::json::parse(res->body)["status"].get<bool>();
+}
+
+bool DeviceGate::USBImpl::isOkay() {
+    USBRequest_t request;
+    request.RequestNum = IS_OKAY;
+    request.RequestSize = 0;
+    XLinkGateWrite(&request, sizeof(USBRequest_t));
+
+    XLinkGateRead(&request, sizeof(request));
```
themarpe: Issue with Gate being in an "wrong state / timeout"
Pass a timeout explicitly here, instead of using a default.
Align the timeout, with Gate impl.

themarpe: Pin down "who" retries in this case when Gate isn't yet available

tests/run_tests.py
```diff
@@ -112,10 +112,16 @@ def compute_or_result(results):
             "labels": ["onhost"],
         },
         {
-            "name": "RVC4",
+            "name": "RVC4 - USB",
+            "env": {"DEPTHAI_PLATFORM": "rvc4", "DEPTHAI_GATE_USB_CONNECTION": "1"},
```
themarpe:
```suggestion
            "env": {"DEPTHAI_PLATFORM": "rvc4", "DEPTHAI_PROTOCOL": "usb"},
```

themarpe: CC: @moratom

tests/run_tests.py
```diff
@@ -112,10 +112,16 @@ def compute_or_result(results):
...
+            "labels": ["rvc4"],
+        },
+        {
+            "name": "RVC4 - POE",
             "env": {"DEPTHAI_PLATFORM": "rvc4"},
```
themarpe:
```suggestion
            "env": {"DEPTHAI_PLATFORM": "rvc4", "DEPTHAI_PROTOCOL": "tcpip"},
```

themarpe: CC: @moratom

PR #1398, opened 2025-07-28
[Draft] adding landmark and imu support to RTABMapSLAM

dragonblade316:
## Purpose

In this pr we add support for RTABMap landmarks and add imu support to the RTABMapSLAM node. These are mainly useful for localization. The markers can allow RTABMap to consider inputs from sources such as apriltags both for global localization and for loop closures. The IMU support we found makes the transform output a bit more stable.

## Specification

Two inputs have been added to the RTABMapSLAM node. One for Landmarks and one for imu input. In order to properly facilitate landmarks, we had to make a new landmark datatype. There are no current nodes in depthai-core that output it but we may make one in the future if we have the time to do so.

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

We have been using our fork to develop a localization software called [PurpleRanger](https://github.com/lasarobotics/PurpleRanger). We have seen mostly good results but the software is still a work in progress and we are still working on eliminating jitter from out output.

## Work in progress
This PR is not finished but it is getting close. We wanted to go ahead and get feedback on what we have written and address a couple problems we have had relating to inputs. 

In RTABMapSLAM.hpp we have had issues properly getting the input syntax used by features to work for us. As a duct tape solution we decided to use the syntax that is used for the required inputs but that does not allow you to turn on and off the landmark input. We think the issue is that the landmarks input was not working properly with the sync node. If you know how to easily implement the input correctly we would be grateful.

Thanks in advance.


moratom:
Closing for inactivity, feel free to reopen when applicable.

PR #1401, opened 2025-07-29
Run tests on oak4s

danilo-pejovic:
addign oak4_s

### DIFF
--- a/.github/workflows/test.workflow.yml
+++ b/.github/workflows/test.workflow.yml
@@ -24,6 +24,7 @@ jobs:
     with:
       flavor: "vanilla"
       luxonis_os_versions_to_test: "['debug-1.6.0', '1.14.1', '1.15.0']"
+      luxonis_os_versions_to_test_rgb: "['1.15.0']"
     secrets:
       CONTAINER_REGISTRY: ${{ secrets.CONTAINER_REGISTRY }}
 
--- a/.github/workflows/test_child.yml
+++ b/.github/workflows/test_child.yml
@@ -8,6 +8,11 @@ on:
       luxonis_os_versions_to_test:
         required: true
         type: string
+      luxonis_os_versions_to_test_rgb:
+        required: true
+        type: string
+        default: "[]"
+      
     secrets:
       CONTAINER_REGISTRY:
         required: true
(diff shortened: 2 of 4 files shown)


examples/cpp/Camera/CMakeLists.txt
```diff
@@ -5,25 +5,25 @@ cmake_minimum_required(VERSION 3.10)
 ## function: dai_set_example_test_labels(example_name ...)
 
 dai_add_example(camera_all camera_all.cpp ON OFF)
-dai_set_example_test_labels(camera_all ondevice rvc2_all rvc4 ci)
+dai_set_example_test_labels(camera_all ondevice rvc2_all rvc4 rvc4rgb ci)
```
moratom, line 8: maybe better to name the tag just "single_cam" as we might expand this for RVC2 too

examples/cpp/HostNodes/display.cpp
```diff
@@ -38,8 +38,8 @@ class HostDisplay : public dai::node::CustomNode<HostDisplay> {
 
 int main() {
     // Set up signal handlers
-    signal(SIGTERM, signalHandler);
-    signal(SIGINT, signalHandler);
+    // signal(SIGTERM, signalHandler);
```
moratom: nit: let's just remove this

examples/python/Benchmark/benchmark_simple.py
moratom, line 1: Is there a reproducible case to how this example fails? Just running it as is and sending a SIGINT?

tests/src/helpers/test_wrapper.cpp
```diff
@@ -91,11 +91,15 @@ int main(int argc, char* argv[]) {
...
         }
 
-        // Only call communicate() once and save the results
-        auto results = proc.communicate();
-
         // Now we can safely check the return code
         int retcode = proc.retcode();
+        // 2 signifies that proc was killed by a timeout
+        if (retcode == 2) {
+        retcode = 0;
```
moratom: nit: indentation

examples/cpp/Camera/CMakeLists.txt
danilo-pejovic: RVC2 tests have to be separate from RVC4 tests anyway - so equivalent for rvc2 would be rvc2rgb. Granted I can change the name, I didnt put much thought into it except to keep it shortish to preserve readability

examples/python/Benchmark/benchmark_simple.py
danilo-pejovic: Run it without my changes and sending sigint should fail it.

PR #1403, opened 2025-07-30
Add manual sensorType setting for the Camera node

lnotspotl:
## Purpose

This PR adds a new field to the `CameraProperties` datastructure called `sensorType`. This field is by default set to AUTO in which case the behavior of the firmware remains unchanged. However, a user can set this attribute to other possible values (e.g. COLOR or MONO), thereby overriding the firmware's automatically picked sensor type.

This is useful in scenarios where a single sensor supports multiple modes of functioning or the automatic way of choosing a mode doesn't suite the user's needs.

I also did a tiny bit of cleanup for the "to_string" helper functions for their easier use on the FW side for logging purposes.

Task on Clickup: https://app.clickup.com/t/86c4k9aez


lnotspotl:
bugbot run

include/depthai/pipeline/node/Camera.hpp
```diff
@@ -55,6 +55,23 @@ class Camera : public DeviceNodeCRTP<DeviceNode, Camera, CameraProperties>, publ
                                   std::optional<std::pair<uint32_t, uint32_t>> sensorResolution = std::nullopt,
                                   std::optional<float> sensorFps = std::nullopt);
 
+    /**
+     * Set the sensor type to use
+     * @param sensorType Sensor type to use
+     */
+    std::shared_ptr<Camera> sensorType(CameraSensorType sensorType) {
```
moratom: Maybe `setSensorType` to make it consistent with other nodes.

moratom (approved):
Thanks!

I would just tweak the setter with an explicit set for consistency, but LGTM otherwise.

PR #1415, opened 2025-08-07
Bring depthai_nodes into depthai-core

thehiddenwaffle:
## Purpose
https://github.com/luxonis/depthai-nodes/pull/230

## Specification
https://github.com/luxonis/depthai-nodes/pull/230

## Testing & Validation
Ensured that example works


moratom (commented):
Thanks @thehiddenwaffle for opening this!

Do you maybe have an example handy you could include?

thehiddenwaffle:
> Do you maybe have an example handy you could include?

My application is pretty nuanced but I can definitely create a quick example. Would it be better to include a more "useful" pipeline like cam->manip->NN->HostParser or something more concise and minimal like exampleImg->XLinkIn->NN->HostParser? Either way is fine with me.

thehiddenwaffle:
@moratom Ended up just making Cam->ParsedNN->Out to show that the tensor functions work. If the cam happens to be pointing at a human it might be "useful".

PR #1416, opened 2025-08-07
Added support for a pybind11 embedded module target

thehiddenwaffle:
## Purpose
https://discuss.luxonis.com/d/6203-are-depthai-core-contributions-welcome/9
Target built by CMake configuration, installing it as a separate(dependant on core) shared lib. This allows other C++ projects to include the additional lib and use python to build its pipeline. Incompatible with depthai_nodes currently.

## Specification
strictly additive, disabled by default

## Testing & Validation
Successfully using from a separate project locally.

### DIFF
--- a/include/depthai/utility/ImageManipImpl.hpp
+++ b/include/depthai/utility/ImageManipImpl.hpp
@@ -2855,7 +2855,7 @@ bool ImageManipOperations<ImageManipBuffer, ImageManipData, WarpBackend>::apply(
             switch(base.colormap) {  // TODO(asahtik): set correct stereo colormaps
                 case Colormap::TURBO:
                 case Colormap::STEREO_TURBO:
-                    // cvColormap = cv::COLORMAP_TURBO;
+                    cvColormap = cv::COLORMAP_TURBO;
                     break;
                 case Colormap::STEREO_JET:
                 case Colormap::JET:


thehiddenwaffle:
> I'm guessing something like a custom host node which is esentially an "on host Script node" would make sense here.

@moratom I've been thinking about it and I feel like maybe 2 examples are needed, one for each of the following cases 

1. My case where the C++ application does *not* need any python custom nodes(the case Python is merely a pipeline "setup" script). I would note that the interpreter can be destroyed after the pipeline is fully built as long as no Python defined host nodes are present(my reason for rewriting depthai_nodes in C++). I believe for most this will be the most appealing option for production and real time systems where speed is critical(so long as the speed of pipeline <ins>setup</ins> is not critical because pybind11 embedded interpreter is <ins>***SO SLOW***</ins>).

2. The case where C++ would like to utilize python to both build the pipeline as well as manage the running of the pipeline(and optionally utilize python derived `HostNode`s) then two things seem true to me: the interpreter *must* be kept alive, and C++ must be able to aquire the data(I think the best way to do this would be to give the Python script a [C++ callback](https://pybind11.readthedocs.io/en/stable/advanced/cast/functional.html) so that it can push frames back to the C++ application for further use)

include/depthai/utility/ImageManipImpl.hpp
```diff
@@ -2855,7 +2855,7 @@ bool ImageManipOperations<ImageManipBuffer, ImageManipData, WarpBackend>::apply(
             switch(base.colormap) {  // TODO(asahtik): set correct stereo colormaps
                 case Colormap::TURBO:
                 case Colormap::STEREO_TURBO:
-                    cvColormap = cv::COLORMAP_TURBO;
+                    // cvColormap = cv::COLORMAP_TURBO;
```
moratom: Intentional?

thehiddenwaffle: Definitely not intended for this PR my opencv version was out of date, will remove.

PR #1417, opened 2025-08-14
examples: Added calibration examples from V2.

zrezke:
## Purpose

Calibration examples/utils have been missing for a long time in depthai-core V3. Adding most of them back in.

I ommitted:
* `calibration_flash_v5.py`, because I don't know why it's useful.
* `calibreation_reader.py`, because it has some hardcoded models, and I don't want to spend time right now checking that this still works on all models.

## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

Tested locally by running `calibration_dump.py`, `calibration_flash.py` and `calibration_factory_reset.py` on an R8 OAK-T.
Tested `calibration_load.py` on an OAK4-D-Pro.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,3 +48,4 @@ _builds/
 .depthai_cached_models/
 install*
 build*
+depthai_calib_backup.json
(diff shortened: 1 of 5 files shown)


moratom (approved):
Thanks Filip!

PR #1418, opened 2025-08-14
Bundle related python wheels together

lnotspotl:
The goal of this task was to significantly reduce the accumulated size of the python wheels generated for each of the major operating systems, including MacOS, Windows and Linux. In case of Linux, the wheels are furthermore built for both the x64 and the arm64 architectures. Lastly, depthai has support for both MacOS 13 and MacOS 14. Overall, we thus have five different configurations. For each of these configurations, we build python bindings for a multiple of different python versions, ranging from python 3.7 all the way to 3.13. The following is a list of accumulated wheel sizes for each and every one of the 5 different configurations:

- Windows 10/11: (7 wheels, python 3.7 - 3.13), avg. size 37 MB
- Linux, x64: (7 wheels, python 3.7 - 3.13), avg. size 51 MB
- Linux, arm64: (7 wheels, python 3.7 - 3.13), avg. size 48 MB
- MacOS 13: (6 wheels, python 3.8 - 3.13), avg. size 44 MB
- MacOS 14: (6 wheels, python 3.8 - 3.13), avg. size 41 MB

This leaves us with a little under 1.5 GB worth of wheels.

The larger bulk of the of the overall size consists of copies of the `depthai-core` library, one copy in each wheel. As suggested by @themarpe, we could reduce the cumulative size by bundling wheels in the same (OS+arch) configuration together, leading to a significant reduction of the overall cumulative size at the cost of slightly larger wheel size. The overall size reduction stems from the re-use of the `depthai-core` library with each python version's binding library linking against a single shared `depthai-core` binary. This leaves us with 5 wheels

- Windows 10/11: 1 wheel, 85 MB
- Linux, x64: 1 wheel, 80 MB
- Linux, arm64: 1 wheel, 75 MB
- MacOS 13: 1 wheel, 60 MB
- MacOS 14: 1 wheel, 55 MB

The overall size shrunk from 1.5 GB to 355 GB, a 76 % reduction in size.

## Tools used
- delvewheel (on Windows) - building wheels on windows is a bit broken in that only libraries the `.pyd` wheel file links against directly are copied to the final wheel (thus if those libraries link against other libraries, those will not be included, resulting in a broken wheel) - delvewheel fixes that. Relinking on Windows is a bit more involved, I've tried many things but reusing delvewheel's internal functions proved to be the most straighforward.

- auditwheel (on Linux) - the most mature of wheel repairing utilities, works essentially out of the box, when bundling wheels together, one has to unmangle the names, introduce a new, common mangled name and relink all the shared libraries to reflect. To do the re-linking, one can use `patchelf`. Validation can be done via `readelf` or `ldd`.

- delocate (on MacOS) - does not do any name mangling so bundling wheels together is as easy as copying everything into a single zip file. MacOS 13 is the only platform that has two platform tags (macosx_13_0_x86_64 and macosx_13_0_universal2)

## Potential Further Improvements

Additional size reduction could be achieved by stripping all bundled shared libraries. This works when one tries to use the `strip` utility locally (I tried on my Linux machine), however, performing stripping (even with just the `--strip-unneeded` flag) in the runner (which uses the `manylinux_2_28` docker image) resulted in a runtime error with the error message saying `ELF load command address/offset not page-aligned`. What works is unzipping a wheel locally, stripping the shared libraries and re-creating the wheel. This yields another 10 % wheel size reduction, the wheel going from 80 MB in size to 72 MB. However, this optimization is not cross-platform (seemingly), at least not when using the set of primitive tools that I used.

Our CI builds `depthai-core` once for each (configuration, python version) tuple. I think we could build the shared `depthai-core` library in a pre-stage and then distribute it to downstream binding building jobs. I list it here for future reference, in case we want to perform further CI optimizations.

You can try installing combined `depthai` python wheels (on any of the supported platform) via

```bash
python3 -m pip install --extra-index-url https://artifacts.luxonis.com/artifactory/luxonis-python-snapshot-local/ depthai==2847814144ca5c9cbdd15dd0dc615f74b2a2ec6d
```


.github/workflows/python-main.yml
```diff
@@ -57,7 +57,7 @@ jobs:
           sudo apt update
           python -m pip install --upgrade pip
           sudo apt install libusb-1.0-0-dev pkg-config bison autoconf libtool libxi-dev libxtst-dev libxrandr-dev libx11-dev libxft-dev libxext-dev nasm flex libudev-dev automake libltdl-dev
-          python -m pip install -r bindings/python/docs/requirements_mkdoc.txt
+           python -m pip install -r bindings/python/docs/requirements_mkdoc.txt
```
moratom: whitespace change

bindings/python/CMakeLists.txt
```diff
@@ -287,17 +287,47 @@ if(DEPTHAI_RTABMAP_SUPPORT AND DEPTHAI_PCL_SUPPORT)
...
+            # pybind11
+            pybind11::pybind11
+            depthai_combined
+            hedley
+            pybind11_json
+            xtensor-python
+            spdlog::spdlog
+    )
+
+endif()
```
themarpe: Shouldn't we just build as BUILD_SHARED? (or atleast the depthai-core) instead of creating a new fixed shared depthai_combined library

.github/workflows/python-main.yml
```diff
@@ -239,22 +239,28 @@ jobs:
...
         run: cd bindings/python && bash ./ci/show_vcpkg_logs.sh
+      - name: Install delvewheel
+        run: python -m pip install delvewheel
+      - name: List wheelhouse
+        run: cd bindings/python && ls ./wheelhouse
+      - name: Auditing wheels
+        shell: bash
+        run: |
+          cd bindings/python
+          EXTRA_DLL_PATH=$(find ./build -type d -name "temp.win*" -exec ls -d {} \; | head -n 1)/Release/Release
```
moratom, line 254: Can we add a comment on what DDLs is this expected to find?

bindings/python/CMakeLists.txt
```diff
@@ -203,15 +203,17 @@ else()
     if(NOT bindings_directory)
         set(bindings_directory ${CMAKE_CURRENT_BINARY_DIR})
     endif()
+
```
moratom, line 206: whitespace change

bindings/python/CMakeLists.txt
```diff
@@ -203,15 +203,17 @@ else()
...
     message(STATUS "Mypy available, creating and checking stubs. Running with generate_stubs.py ${TARGET_NAME} ${bindings_directory}")
     if (NOT EXISTS "${CMAKE_CURRENT_LIST_DIR}/generate_stubs.py")
         message(FATAL_ERROR "Cannot add custom build command, ${CMAKE_CURRENT_LIST_DIR}/generate_stubs.py not found")
     endif ()
+
```
moratom: whitespace

bindings/python/CMakeLists.txt
```diff
@@ -289,17 +291,47 @@ if(DEPTHAI_RTABMAP_SUPPORT AND DEPTHAI_PCL_SUPPORT)
...
+    )
+
+else() # UNIX
+
+    # Create a shared library that combines all static libraries
+    # The libraries are combined here to ensure link-time optimization (and mostly importantly aggresive DCE) can happen
+    file(WRITE ${CMAKE_CURRENT_BINARY_DIR}/empty.cpp "")
+    add_library(depthai_combined SHARED ${CMAKE_CURRENT_BINARY_DIR}/empty.cpp)
+
+    target_link_options(depthai_combined PRIVATE -Wl,--whole-archive)
```
moratom: Maybe good to add a comment here why this whole-archive is needed.

I assume otherwise everything gets optimized out, since the empty.cpp doesn't require nothing from the libs.

bindings/python/CMakeLists.txt
```diff
@@ -289,17 +291,47 @@ if(DEPTHAI_RTABMAP_SUPPORT AND DEPTHAI_PCL_SUPPORT)
...
+    target_link_libraries(${TARGET_NAME}
+        PRIVATE
+            # pybind11
+            pybind11::pybind11
+            depthai_combined
+            hedley
+            pybind11_json
+            xtensor-python
+            spdlog::spdlog
+    )
```
moratom: Maybe better to remove this out of the if/else and then just set the DEPTHAI_LINK_TARGET to be depthai_combined, so the other libs are not duplicated

bindings/python/CMakeLists.txt
```diff
@@ -289,17 +291,47 @@ if(DEPTHAI_RTABMAP_SUPPORT AND DEPTHAI_PCL_SUPPORT)
...
-        ${DEPTHAI_LINK_TARGET}
-        hedley
-        pybind11_json
-        xtensor-python
-        spdlog::spdlog
-)
+
+
+
+if(WIN32 OR APPLE)
```
moratom: Do we just "take the hit" in case of Windows & Apple with the dependencies being there twice?

bindings/python/CMakeLists.txt
moratom: The reason for it is to avoid having opencv, eigen, etc twice in the .so[s], since both depthai::rtabmap and depthai::core depend on opencv, eigen, etc and we statically compile the dependanices to each .so since they are not "PUBLIC".

So taking the current state and just changing it to BUILD_SHARED=ON for all platforms results in a bit bigger wheels when compressed and quite a lot bigger wheels when extracted.

bindings/python/setup.py
```diff
@@ -129,6 +131,7 @@ def build_extension(self, ext):
             cmake_args += ['-DDEPTHAI_KOMPUTE_SUPPORT=ON']
         build_args += ['--target=depthai']
```
moratom, line 133: whitespace change

bindings/python/generate_stubs.py
```diff
@@ -4,23 +4,44 @@
...
+            src = os.path.join(DIRECTORY, item)
+            dst = os.path.join(PIP_TEMP_LIB_FOLDER, item)
+            print(f'Copying {src} to {dst}')
+            if os.path.isfile(src):
+                shutil.copy2(src, dst)
+            elif os.path.isdir(src):
+                shutil.copytree(src, dst)
+        stubgen_root = PIP_TEMP_LIB_FOLDER
+    else:
+        stubgen_root = DIRECTORY
```
moratom: Can we avoid this? This forces you to use `-DPIP_TEMP_LIB_FOLDER` when iterating with Python bindings (I guess avoiding it on Linux&MacOS is trivial, but ideally it wouldn't be needed on Windows either)

bindings/python/CMakeLists.txt
moratom: I think just extending the merged target "by hand" might be a fine way to proceed here as well potentially, similar to how we merge PCL & OpenCVs targets and we can avoid the combined target and the if/elses here.

bindings/python/CMakeLists.txt
lnotspotl: 86bd48c7f8ba2770bf524f757b06c186aad785e4

bindings/python/CMakeLists.txt
lnotspotl: 86bd48c7f8ba2770bf524f757b06c186aad785e4

.github/workflows/python-main.yml
lnotspotl: 86bd48c7f8ba2770bf524f757b06c186aad785e4

bindings/python/CMakeLists.txt
lnotspotl: I would not say we are taking a hit, having tested many different things and having asked different LLMs to generate possible solutions, there would always be some kind of a shortcoming. It seems the solution to merging two statically linked libraries into a single shared library is non-trivial on both macOS and Windows.

bindings/python/generate_stubs.py
lnotspotl: Putting the binaries and other files to a common folder is necessary on Windows for `stubgen` to properly find all the libraries it needs. I guess when one does local development, all libraries are in the same folder and no copying thus is necessary. I made the necessary to reflect that in c32e5d2208f55ee5d3f6d8533a78e1781f9f0c79

bindings/python/CMakeLists.txt
lnotspotl: 4e51643e89bb5eb32d0ce23f0c681d14ebf81c81

bindings/python/CMakeLists.txt
lnotspotl: @moratom Are you suggesting we dump the current solution of merging static libraries into a single combined library on Linux and employ the DEPTHAI_MERGED_TARGET flag instead, merging the current `rtabmap` and `basaltvio` libraries into `depthai-core` in case that flag is used?

That makes sense in my view and would be a little more cross-platform, alleviating the need to add platform-specific linker flags and what not.

bindings/python/CMakeLists.txt
themarpe: Then let me rephrase to not build ALL as shared, but just `depthai-core`
I don't think creating a special target for that is "worth it"

bindings/python/CMakeLists.txt
themarpe: https://github.com/luxonis/depthai-core/pull/1418#discussion_r2284463239 I hope this applies to it, where depthai-core is shared, rest is static and then depthai-core is just "the target that merges everything anyway"

themarpe: AKA

https://cmake.org/cmake/help/latest/command/add_library.html
```
add_library(<name> [<type>] [EXCLUDE_FROM_ALL] <sources>...)
Add a library target called <name> to be built from the source files listed in the command invocation.

The optional <type> specifies the type of library to be created:

STATIC
A [Static Library](https://cmake.org/cmake/help/latest/manual/cmake-buildsystem.7.html#static-libraries): an archive of object files for use when linking other targets.

SHARED
A [Shared Library](https://cmake.org/cmake/help/latest/manual/cmake-buildsystem.7.html#shared-libraries): a dynamic library that may be linked by other targets and loaded at runtime.

MODULE
A [Module Library](https://cmake.org/cmake/help/latest/manual/cmake-buildsystem.7.html#module-libraries): a plugin that may not be linked by other targets, but may be dynamically loaded at runtime using dlopen-like functionality.

If no <type> is given the default is STATIC or SHARED based on the value of the [BUILD_SHARED_LIBS](https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html#variable:BUILD_SHARED_LIBS) variable.
```

If you pass a TYPE in this case as `SHARED` as a variable (which by default is "empty") to the depthai-core library, should address this

moratom: >@moratom Are you suggesting we dump the current solution of merging static libraries into a single combined library on Linux and employ the DEPTHAI_MERGED_TARGET flag instead, merging the current rtabmap and basaltvio libraries into depthai-core in case that flag is used?

> That makes sense in my view and would be a little more cross-platform, alleviating the need to add platform-specific linker flags and what not.

Yeah, that's exactly what I had in mind.

@themarpe the challange with just making the depthai target**s** shared libraries is multiple copies of the statically compiled dependanices that are inside each one of them.
```
depthai[bindings] -> depthai-core.so (statically compiled in opencv, eigen, ...)
                  -> depthai-rtabmap.so (statically compiled in opencv, eigen, ...)
                  -> depthai-basalt.so (statically compiled in opencv, eigen, ...)
```
This is fine and AFAIK works, but the size is noticably bigger compared to making one large target, so the statically compiled in libraries are statically compiled only once instead of three times.

Complexity mostly introduced by mixing shared&static libraries.

themarpe: Ah I see - what if depthai-core is "root" (aka it includes rtabmap & basalt as static as well)?
That way we have 1 so and everything else it includes is static

& we make that potential first class citizen also for regular (C++/native) usage

moratom: >Ah I see - what if depthai-core is "root" (aka it includes rtabmap & basalt as static as well)?
That way we have 1 so and everything else it includes is static

Yeah, I think that's the way to go, we already have this made with `depthai-pcl` & `depthai-opencv`.

lnotspotl: Let's do that, it only makes sense to combine all the sources into a single array and then generate a shared library once. It is both cross-platform and does not use any exotic, platform-specific linker flags.

bindings/python/CMakeLists.txt
lnotspotl: No more of this, we're merging all the parts into a single `depthai-core` shared library.

bindings/python/CMakeLists.txt
```diff
@@ -404,4 +404,4 @@ endif()
 ########################
 if (DEPTHAI_PYTHON_ENABLE_EXAMPLES)
     add_subdirectory(../../examples/python ${CMAKE_BINARY_DIR}/examples/python)
-endif()
+endif()
```
moratom, line 407: nitpick: whitespace change :)

.github/workflows/python-main.yml
moratom: Let's check if this is still needed with the DEPTHAI_MERGED_TARGET

bindings/python/CMakeLists.txt
lnotspotl: We no longer use the `--whole-archive` flag.

.github/workflows/python-main.yml
lnotspotl: This is still necessary. As discussed yesterday, setting the `BUILD_SHARED_LIBS` option to `ON` causes that there are more than two "levels" of shared library dependencies. For instance, `wheel.so` depends on `depthai-core.so` which depends on `libusb.so`. There is a [utility](https://github.com/luxonis/depthai-core/blob/lnotspotl/combined_python_wheels_opt/bindings/python/CMakeLists.txt#L179-L195) that only copies the first layer of dependencies on Windows. 

Note that this is not an issue on Linux or MacOS as on these two operating systems the libraries have their runtime paths set correctly.
Windows, on the other hand, does not have the notion of a runtime path (RPATH), thus libraries have to be co-located.

PR #1435, opened 2025-09-02
Re-expose camera pool setters and getters, cap the number of pool frames (either by number or by size in bytes, however always at least 2)

lnotspotl:
Add missing setters and getters, cap the number of pool frames

> There is a corresponding FW codebase PR

### DIFF
--- a/cmake/Depthai/DepthaiDeviceRVC4Config.cmake
+++ b/cmake/Depthai/DepthaiDeviceRVC4Config.cmake
@@ -3,4 +3,4 @@
 set(DEPTHAI_DEVICE_RVC4_MATURITY "snapshot")
 
 # "version if applicable"
-set(DEPTHAI_DEVICE_RVC4_VERSION "0.0.1+4c03dcd1a75e6eaeec79d7638c70fefe4bea3a80")
+set(DEPTHAI_DEVICE_RVC4_VERSION "0.0.1+d9ffe3ca006f85ae4d976468c1a342828739ae18")
--- a/cmake/Depthai/DepthaiDeviceSideConfig.cmake
+++ b/cmake/Depthai/DepthaiDeviceSideConfig.cmake
@@ -2,7 +2,7 @@
 set(DEPTHAI_DEVICE_SIDE_MATURITY "snapshot")
 
 # "full commit hash of device side binary"
-set(DEPTHAI_DEVICE_SIDE_COMMIT "9a69a3cbcc7e5b96d07cbe3bba38dde6bd93df97")
+set(DEPTHAI_DEVICE_SIDE_COMMIT "dc466b4462739d1edbaa00fdccbe63805bff8078")
 
 # "version if applicable"
 set(DEPTHAI_DEVICE_SIDE_VERSION "")


bindings/python/src/pipeline/node/CameraBindings.cpp
```diff
@@ -40,6 +40,22 @@ void bind_camera(pybind11::module& m, void* pCallstack) {
...
+        .def("setStillNumFramesPool", &Camera::setStillNumFramesPool, "num"_a, DOC(dai, node, Camera, setStillNumFramesPool))
+        .def("setNumFramesPools", &Camera::setNumFramesPools, "raw"_a, "isp"_a, "preview"_a, "video"_a, "still"_a, DOC(dai, node, Camera, setNumFramesPools))
+        .def("setMaxSizePools", &Camera::setMaxSizePools, "raw"_a, "isp"_a, DOC(dai, node, Camera, setMaxSizePools))
+        .def("getRawNumFramesPool", &Camera::getRawNumFramesPool, DOC(dai, node, Camera, getRawNumFramesPool))
+        .def("getMaxSizePoolRaw", &Camera::getMaxSizePoolRaw, DOC(dai, node, Camera, getMaxSizePoolRaw))
+        .def("getIspNumFramesPool", &Camera::getIspNumFramesPool, DOC(dai, node, Camera, getIspNumFramesPool))
+        .def("getMaxSizePoolIsp", &Camera::getMaxSizePoolIsp, DOC(dai, node, Camera, getMaxSizePoolIsp))
+        .def("getPreviewNumFramesPool", &Camera::getPreviewNumFramesPool, DOC(dai, node, Camera, getPreviewNumFramesPool))
+        .def("getVideoNumFramesPool", &Camera::getVideoNumFramesPool, DOC(dai, node, Camera, getVideoNumFramesPool))
+        .def("getStillNumFramesPool", &Camera::getStillNumFramesPool, DOC(dai, node, Camera, getStillNumFramesPool))
```
moratom: I think most of these are unsused now (legacy fileds in the properties), so we should rather remove them from properties instead.

src/pipeline/node/Camera.cpp
```diff
@@ -243,6 +253,11 @@ Node::Output* Camera::requestOutput(std::pair<uint32_t, uint32_t> size,
     cap.type = type;
     cap.resizeMode = resizeMode;
     cap.enableUndistortion = enableUndistortion;
+
+    // Update internal image manip pool sizes
+    properties.numFramesPoolImageManip = numFramesPoolImageManip.has_value() ? numFramesPoolImageManip.value() : properties.numFramesPoolImageManip;
+    properties.maxSizePoolImageManip = maxSizePoolImageManip.has_value() ? maxSizePoolImageManip.value() : properties.maxSizePoolImageManip;
```
lnotspotl: This seemingly has no effect, why??

I've tried printf debugging and the value does get set correctly, however, when I log the same value on device, it remains unchanged.

moratom: Missing fileds in DEPTHAI_SERIALIZE

moratom:
> @moratom pool sizes should probably not be tied to outputs as a user could then requests two or more outputs with different pool size settings, ultimately leading to confusion

+1 - I think we should make this a setting on the Camera node `setOutputsNumFramesPool` or something like that.
We can add the setting to the `requestOutput` at a later time if we'll expose this more granularly.

moratom (approved):
Thanks!

One last nit, I would change `setOutputsNumFramesPool` (skip ImageManip)
The why is mostly that this an internal "detail" that there is an ImageManip and which is not transparent, so this might bring more confusion than good.

PR #1445, opened 2025-09-08
Update/stereo examples

jakaskerl:
## Purpose
Change the requested type for L & R cameras. Was NV12 which implicitly tries to convert mono (RAW8) frames and impacts performance.

## Testing & Validation
Tested on python, didn't test on cpp.


moratom:
Thansk @jakaskerl ! I fixed the compilation of the C++ examples, but otherwise looks good.
This was legacy from back when RVC4 didn't support grayscale outputs and everything was NV12.

PR #1446, opened 2025-09-10
Added testing of wheel download after upload

MaticTonin:
Reason for that PR request is to test if the uploaded wheels to artifactory are correctly installed after each upload for all platforms.

Idea is to catch error like this one:
https://github.com/luxonis/depthai-core/actions/runs/17500499704/job/49713758148

Needed change would be to introduce the artifactory page IP as a secret (required, otherwise GitHub labeled it as untrusty page.

### DIFF
--- a/.github/workflows/python-main.yml
+++ b/.github/workflows/python-main.yml
@@ -361,7 +361,46 @@ jobs:
           ARTIFACTORY_URL: ${{ secrets.ARTIFACTORY_URL }}
           ARTIFACTORY_USER: ${{ secrets.ARTIFACTORY_USER }}
           ARTIFACTORY_PASS: ${{ secrets.ARTIFACTORY_PASS }}
+      - name: Append build hash if not a tagged commit
+        if: startsWith(github.ref, 'refs/tags/v') != true
+        run: echo "BUILD_COMMIT_HASH=${{github.sha}}" >> $GITHUB_ENV
+      - name: Install from Artifactory and smoke-test (macOS)
+        if: startsWith(github.ref, 'refs/tags/v') != true
+        env:
+          ARTIFACTORY_URL: ${{ secrets.ARTIFACTORY_URL }}
+        run: |
+          set -euo pipefail
 
+          # Resolve the exact dev version (includes commit hash)
+          ver=$(python -c "import os,sys,pathlib; sys.path.insert(0, str(pathlib.Path('bindings/python').resolve())); import find_version as v; print(v.get_package_dev_version(os.environ['BUILD_COMMIT_HASH']))")
+          echo "Installing depthai==$ver using $(python -V)"
+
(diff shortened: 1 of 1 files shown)


lnotspotl:
I really like this, @MaticTonin, and once https://github.com/luxonis/depthai-core/pull/1469 is merged in, this should start working.

.github/workflows/python-main.yml
```diff
@@ -398,7 +398,46 @@ jobs:
...
+          set -euo pipefail
+
+          # Resolve the exact dev version (includes commit hash)
+          ver=$(python -c "import os,sys,pathlib; sys.path.insert(0, str(pathlib.Path('bindings/python').resolve())); import find_version as v; print(v.get_package_dev_version(os.environ['BUILD_COMMIT_HASH']))")
+          echo "Installing depthai==$ver using $(python -V)"
+
+          # Install from Artifactory snapshot
+          python -m pip install -U pip
+          python -m pip uninstall -y depthai || true
+          python -m pip install -U --prefer-binary \
```
lnotspotl: `--force-reinstall` would also ensure that if there already is a depthai installation, it gets overwritten

`--prefer-binary`, I think, is not necessary here but we can keep it in

.github/workflows/python-main.yml
```diff
@@ -398,7 +398,46 @@ jobs:
...
+
+          expected = os.environ["ver"]  # read the env var you passed in
+          installed = getattr(dai, "__version__", "<unknown>")
+          if installed != expected:
+              print(f"Version mismatch: installed {installed} vs expected {expected}", file=sys.stderr)
+              sys.exit(1)
+
+          print("depthai:", installed)
+          print("python:", platform.python_version(), "ABI:", sys.implementation.cache_tag)
+          PY
```
lnotspotl: This code snippet is repeated three times, so maybe we could move it to a dedicated script file? Just a suggestion, no push-back.

.github/workflows/python-main.yml
MaticTonin: WIll do force reinstall remove the currently required install

.github/workflows/python-main.yml
MaticTonin: I think it is a good idea! SInce yea, duplication is harder to debug than script!

MaticTonin: Added script in [b1e1969](https://github.com/luxonis/depthai-core/commit/b1e19696cf95d2b5cb438f2fd5c4338451765ad3)

.github/workflows/python-main.yml
MaticTonin: Changed in [b1e1969](https://github.com/luxonis/depthai-core/commit/b1e19696cf95d2b5cb438f2fd5c4338451765ad3)

PR #1452, opened 2025-09-11
Add a -Wweak-tables clang compiler flag, resolve all warnings

lnotspotl:
## Purpose
Due to the fact that python wheels are linking against a single shared library depthai-core and we're making extensive use of dynamic casting in our code base, it was necessary to remove all weak vtables from our codebase (some were ignored) to ensure proper casting. Specifically, on MacOS, dynamic_casts would rather confusingly return `nullptr` on objects with their parent class being that class being cast to.

A compiler flag was added to identify any weak tables generated by the compiler and to throw an error if there are any.

> This is a clang issue only and happens exclusively on MacOS.

Note: *There is a corresponding PR in the FW repository.*

### DIFF
--- a/cmake/Depthai/DepthaiDeviceSideConfig.cmake
+++ b/cmake/Depthai/DepthaiDeviceSideConfig.cmake
@@ -2,7 +2,7 @@
 set(DEPTHAI_DEVICE_SIDE_MATURITY "snapshot")
 
 # "full commit hash of device side binary"
-set(DEPTHAI_DEVICE_SIDE_COMMIT "7896225482f51543ae2f213478ebae61cd877a6c")
+set(DEPTHAI_DEVICE_SIDE_COMMIT "9fc3ffc2089dcb51977da0208407f0dd88a33e04")
 
 # "version if applicable"
 set(DEPTHAI_DEVICE_SIDE_VERSION "")


cmake/ports/libusb/portfile.cmake
moratom, line 1: Needed?

lnotspotl: libusb does not compile without the introduced changes

CMakeLists.txt
```diff
@@ -3,7 +3,7 @@ cmake_minimum_required(VERSION 3.20)
 include(cmake/depthaiOptions.cmake)
 
 if(WIN32)
-    add_compile_options(/MP)
+    add_compile_options(/MP /Ob2)
```
moratom: Let's just add a comment on why we added this

include/depthai/xlink/XLinkStream.hpp
```diff
@@ -123,6 +110,11 @@ class XLinkStream {
     std::string getStreamName() const;
 };
 
+#ifdef __clang__
+    #pragma clang diagnostic push
+    #pragma clang diagnostic ignored "-Wweak-vtables"
+#endif
```
moratom: How come we ignore those here?

include/depthai/pipeline/datatype/NNData.hpp
```diff
@@ -42,7 +42,9 @@
 #endif
 #if defined(__clang__)
     #if __has_warning("-Wswitch-enum")
-        #pragma clang diagnostic pop
+        #ifdef __clang__
+            #pragma clang diagnostic pop
+        #endif
```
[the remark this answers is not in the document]
moratom: How come this is needed?

CMakeLists.txt
lnotspotl: Agreed, added a little comment with a link to the PR in the tensorflow codebase that motivates this compiler flag addition.

include/depthai/xlink/XLinkStream.hpp
lnotspotl: We should not. I added explicit definition of the three destructors for the structs within this block ensuring that there are no weak virtual tables. [38ce5dc](https://github.com/luxonis/depthai-core/pull/1452/commits/38ce5dc689d412c38fc24c4b39565b8bc280dc38)

moratom (approved):

PR #1461, opened 2025-09-17
Feature/point cloud data bindings update

aljazdu:
## Purpose
Currently it is not possible to make changes to dai.PointCloudData from Python

## Specification
Update of python PointCloudDataBindings, by adding setPoints() and setPointsRGB() corresponding to the already implemented getPoints() and getPointsRGB bindings. Add a new example showcasing the modification of a captured point cloud with the use of new bindings in a custom processing node

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation
The new bindings were tested by accessing the existing point cloud data, modifying it and setting the modified values of point cloud data, which were observed for correctness. Additional testing was done with the use of a newly created examples, to test if the desired behavior of a custom pointcloud processing node was achieved

### DIFF
--- a/bindings/python/tests/CMakeLists.txt
+++ b/bindings/python/tests/CMakeLists.txt
@@ -13,6 +13,7 @@ set(PYBIND11_TEST_FILES
     "utf8_support_test.py"
     # "dai_path_conversion_test.py" requires device
     "nndata_tensor_test.py"
+    "pcldata_set_points_test.py"
     "messsage_queue_test.py"
     "imgframe_test.py"
     "inherited_messages_test.py"
--- a/bindings/python/tests/pcldata_set_points_test.py
+++ b/bindings/python/tests/pcldata_set_points_test.py
@@ -1,0 +1,33 @@
+import depthai as dai
+import numpy as np
+
+def test_pcldata_set_points():
+    pointCloudData = dai.PointCloudData()
+
+    # Points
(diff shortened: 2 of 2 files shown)


examples/python/CMakeLists.txt
```diff
@@ -239,6 +239,9 @@ dai_set_example_test_labels(visualizer_rgbd_autocreate ondevice rvc2 usb rvc4 ci
 add_python_example(rgbd_o3d RGBD/rgbd_o3d.py)
 dai_set_example_test_labels(rgbd_o3d ondevice rvc2_all rvc4 ci nowindows)
 
+add_python_example(rgbd_pcl_processing RGBD/rgbd_pcl_processing.py)
+dai_set_example_test_labels(rgbd_pcl_processing ondevice rvc2 usb rvc4 ci nowindows)
```
moratom: Any reason to go with rvc2_usb instead of rvc2_all here?

moratom (commented):
LGTM - let's just add a test (see https://gitlab.luxonis.com/luxonis/depthai/depthai-device-kb/-/merge_requests for an example) before merging.

Thanks!

examples/python/CMakeLists.txt
aljazdu: Not particularly, will change it to rvc2_all

moratom (approved):
Thanks!

PR #1467, opened 2025-09-21
zoo helper json format enhancements

lnotspotl:
This PR fixes the compilation workflow for the zoo_helper binary and streamlines that process by disabling all irrelevant features (OpenCV support, Protobuf support,  etc.). Moreover, several unused includes were removed from the `Logging.hpp` header to reduce the number of zoo_helper's dependencies. Finally, a new requested feature was added to `Zoo.cpp` that reports errors in the form of a single-line json in case `downloadModelsFromZoo`'s `progressFormat` argument is set to `json`


moratom:
Thanks!

PR #1469, opened 2025-09-23
add numpy as depthai's dependency

lnotspotl:
## Purpose
Add numpy as a dependency such that it gets installed (if not already installed) alongside the depthai wheel itself.

> Without the `--no-deps` option, the `pip wheel` command would all the dependencies' wheels and places them in the given output file, alongside the depthai wheel itself.  See https://pip.pypa.io/en/stable/cli/pip_wheel/#cmdoption-no-deps

image

### DIFF
--- a/bindings/python/setup.py
+++ b/bindings/python/setup.py
@@ -280,7 +280,7 @@ setup(
     ],
     python_requires='>=3.7',
     install_requires=[
-        "numpy>=1.26.0",
+        "numpy>=2.0.0",
     ],
     entry_points={
         "console_scripts": [


bindings/python/setup.py
```diff
@@ -279,6 +279,9 @@ def build_extension(self, ext):
         "Topic :: Software Development",
     ],
     python_requires='>=3.7',
+    install_requires=[
+        "numpy>=1.26.0",
```
moratom: Any reason for the 1.26 version set as minimum?

moratom (approved):
Thanks!

Apart from the question why we went with 1.26, LGTM

bindings/python/setup.py
lnotspotl: Changed to 2.0.0

PR #1470, opened 2025-09-24
Pipeline debugging

asahtik:
## Purpose

Implements pipeline events which should simplify debugging stuck pipelines and help profile the pipeline.

## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

### DIFF
--- a/README.md
+++ b/README.md
@@ -226,6 +226,7 @@ The following environment variables can be set to alter default behavior of the 
 | DEPTHAI_RECORD | Enables holistic record to the specified directory. |
 | DEPTHAI_REPLAY | Replays holistic replay from the specified file or directory. |
 | DEPTHAI_PROFILING | Enables runtime profiling of data transfer between the host and connected devices. Set to 1 to enable. Requires DEPTHAI_LEVEL=debug or lower to print. |
+| DEPTHAI_PIPELINE_DEBUGGING | Enables pipeline debugging with state dumps. DEPTHAI_LEVEL=trace is required to print the state dumps. |
 
 ## Running tests
 


bindings/python/CMakeLists.txt
```diff
@@ -98,7 +98,7 @@ set(SOURCE_LIST
     src/pipeline/node/UVCBindings.cpp
     src/pipeline/node/ToFBindings.cpp
     src/pipeline/node/PointCloudBindings.cpp
-    src/pipeline/node/SyncBindings.cpp
```
asahtik, line 101: SyncBindings were duplicated (line 96)

include/depthai/pipeline/ThreadedNode.hpp
```diff
@@ -43,7 +51,7 @@ class ThreadedNode : public Node {
     virtual void run() = 0;
 
     // check if still running
-    bool isRunning() const;
+    bool isRunning();
```
asahtik, line 54: isRunning sends the loop event so const needed to be booted

include/depthai/pipeline/NodeObjInfo.hpp
```diff
@@ -15,6 +15,7 @@ struct NodeObjInfo {
 
     std::string name;
     std::string alias;
+    std::string device;
```
moratom: Will this be the deviceId?

src/pipeline/node/host/Display.cpp
```diff
@@ -37,8 +37,12 @@ Display::Display(std::string name) : name(std::move(name)) {}
 
 void Display::run() {
     auto fpsCounter = FPSCounter();
-    while(isRunning()) {
-        std::shared_ptr<dai::ImgFrame> imgFrame = input.get<dai::ImgFrame>();
+    while(mainLoop()) {
+        std::shared_ptr<dai::ImgFrame> imgFrame = nullptr;
+        {
+            auto blockEvent = this->inputBlockEvent();
```
moratom, line 43: What does this do?

include/depthai/pipeline/NodeObjInfo.hpp
moratom: I think we should add `bool deviceNode` as well, to be explicit.

bindings/python/src/pipeline/datatype/PipelineEventBindings.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+    // Message
+    pipelineEvent.def(py::init<>())
+        .def("__repr__", &PipelineEvent::str)
+        .def_readwrite("nodeId", &PipelineEvent::nodeId, DOC(dai, PipelineEvent, nodeId))
+        .def_readwrite("status", &PipelineEvent::status, DOC(dai, PipelineEvent, status))
+        .def_readwrite("queueSize", &PipelineEvent::queueSize, DOC(dai, PipelineEvent, queueSize))
+        .def_readwrite("interval", &PipelineEvent::interval, DOC(dai, PipelineEvent, interval))
+        .def_readwrite("type", &PipelineEvent::type, DOC(dai, PipelineEvent, type))
+        .def_readwrite("source", &PipelineEvent::source, DOC(dai, PipelineEvent, source))
```
moratom, line 58: Doesn't need bindings, should work "automagically"

bindings/python/src/pipeline/datatype/PipelineStateBindings.cpp
```diff
@@ -0,0 +1,115 @@
...
+    // Message
+    pipelineState.def(py::init<>())
+        .def("__repr__", &PipelineState::str)
+        .def_readwrite("nodeStates", &PipelineState::nodeStates, DOC(dai, PipelineState, nodeStates))
+        .def("getTimestamp", &PipelineState::Buffer::getTimestamp, DOC(dai, Buffer, getTimestamp))
+        .def("getTimestampDevice", &PipelineState::Buffer::getTimestampDevice, DOC(dai, Buffer, getTimestampDevice))
+        .def("getSequenceNum", &PipelineState::Buffer::getSequenceNum, DOC(dai, Buffer, getSequenceNum))
+        .def("setTimestamp", &PipelineState::setTimestamp, DOC(dai, Buffer, setTimestamp))
+        .def("setTimestampDevice", &PipelineState::setTimestampDevice, DOC(dai, Buffer, setTimestampDevice))
+        .def("setSequenceNum", &PipelineState::setSequenceNum, DOC(dai, Buffer, setSequenceNum));
```
moratom, line 114: Same as for PipelineEvent bindings

include/depthai/pipeline/Pipeline.hpp
```diff
@@ -114,6 +123,12 @@ class PipelineImpl : public std::enable_shared_from_this<PipelineImpl> {
     std::unordered_map<std::string, std::filesystem::path> recordReplayFilenames;
     bool removeRecordReplayFiles = true;
     std::string defaultDeviceId;
+    bool pipelineOnHost = true;
```
moratom: What does this signify? If the pipeline has any host nodes?

include/depthai/pipeline/Pipeline.hpp
```diff
@@ -477,6 +497,10 @@ class Pipeline {
     void build() {
         impl()->build();
     }
+    void buildDevice() {
+        impl()->pipelineOnHost = false;
+        impl()->build();
+    }
```
moratom, line 508: Not used?

examples/python/Misc/PipelineDebugging/get_pipeline_state.py
```diff
@@ -0,0 +1,79 @@
...
+                            print("-----")
+                            print(nodeState.outputsSendTiming)
+                            print("-----")
+                        print(f"## inputStates: {'empty' if not nodeState.inputStates else ''}")
+                        for inputName, inputState in nodeState.inputStates.items():
+                            if inputState.isValid():
+                                print(f"### {inputName}:\n-----{inputState}\n-----")
+                            else:
+                                print(f"### {inputName}: invalid")
+                        print(f"## outputStates: {'empty' if not nodeState.outputStates else ''}")
```
moratom: Can we print json version of the object instead? Likely easier to digest

src/pipeline/Pipeline.cpp
```diff
@@ -536,101 +581,165 @@ bool PipelineImpl::isBuilt() const {
...
+        auto allNodes = getAllNodes();
+        if(std::find_if(allNodes.begin(), allNodes.end(), [](const std::shared_ptr<Node>& n) { return strcmp(n->getName(), "PipelineEventAggregation") == 0; })
+           == allNodes.end()) {
+            for(auto& node : allNodes) node->pipelineEventDispatcher->sendEvents = false;
+        }
+    }
+
+    isBuild = true;
+
+    // Go through the build stages sequentially
```
moratom, line 625: I think we should split this in a separate functions as the `build()` is already hard to follow.

src/pipeline/node/host/Display.cpp
asahtik: `mainLoop` sends a main loop event and returns `isRunning()`, blockEvent sends the start event in the constructor and the end event in the destructor.

include/depthai/pipeline/Pipeline.hpp
asahtik: It's used on RVC4

include/depthai/pipeline/Pipeline.hpp
asahtik: Pipeline debugging and record & replay sections should be skipped when the pipeline is being built on device. This flag specifies if it's building on host. I added a comment and changed the name to make it more clear.

examples/python/Misc/PipelineDebugging/get_pipeline_state.py
asahtik: I think this makes it more clear what data is in the pipeline state and how the user can interact with it. Also just printing the serialized json doesn't handle invalid timings so there'd be a lot more junk.

bindings/python/src/pipeline/node/PipelineEventAggregationBindings.cpp
```diff
@@ -0,0 +1,37 @@
...
+    // pipelineEventAggregationProperties.def_readwrite("aggregationWindowSize", &PipelineEventAggregationProperties::aggregationWindowSize)
+    //     .def_readwrite("eventBatchSize", &PipelineEventAggregationProperties::eventBatchSize);
+    //
+    // // Node
+    // pipelineEventAggregation.def_readonly("out", &PipelineEventAggregation::out, DOC(dai, node, PipelineEventAggregation, out))
+    //     .def_readonly("inputs", &PipelineEventAggregation::inputs, DOC(dai, node, PipelineEventAggregation, inputs))
+    //     .def("setRunOnHost", &PipelineEventAggregation::setRunOnHost, py::arg("runOnHost"), DOC(dai, node, PipelineEventAggregation, setRunOnHost))
+    //     .def("runOnHost", &PipelineEventAggregation::runOnHost, DOC(dai, node, PipelineEventAggregation, runOnHost));
+    // daiNodeModule.attr("PipelineEventAggregation").attr("Properties") = pipelineEventAggregationProperties;
+}
```
moratom: no-op?

src/pipeline/Pipeline.cpp
```diff
@@ -536,101 +618,115 @@ bool PipelineImpl::isBuilt() const {
 void PipelineImpl::build() {
     // TODO(themarpe) - add mutex and set running up ahead
```
moratom: Let's actually add a mutex now

src/utility/PipelineEventDispatcher.cpp
```diff
@@ -0,0 +1,255 @@
...
+    else
+        event.setTimestamp(ts.value());
+    event.tsDevice = event.ts;
+    event.nodeId = nodeId;
+    event.interval = PipelineEvent::Interval::END;
+
+    if(out) try {
+            out->send(std::make_shared<dai::PipelineEvent>(event));
+        } catch(const MessageQueue::QueueException&) {
+        }
```
moratom, line 170: Is it ever expected for out not to be null? We should assert if not.

src/remote_connection/RemoteConnectionImpl.cpp
```diff
@@ -405,34 +405,98 @@ void RemoteConnectionImpl::exposeKeyPressedService() {
...
+                auto state = pipeline.getPipelineState().nodes().detailed();
+                stateStr = state.toJson().dump();
+            } catch(const std::runtime_error& e) {
+                stateStr = R"({"error": "Pipeline debugging disabled. Cannot get pipeline state."})";
+            }
+            auto response = foxglove::ServiceResponse();
+            response.data = std::vector<uint8_t>(stateStr.begin(), stateStr.end());
+            return response;
+        };
+    }
```
moratom: What happens if pipeline goes out of scope before RemoteConnection does?

For example in viewer, they hold a single remote connection and switch out the pipelines.

src/utility/Environment.hpp
```diff
@@ -97,14 +97,15 @@ T getEnvAs(const std::string& var, T defaultValue, spdlog::logger& logger, bool
...
+                if(value == "1" || value == "true" || value == "t" || value == "on") {
                     returnValue = true;
-                } else if(value == "0" || value == "false" || value == "FALSE" || value == "False") {
+                } else if(value == "0" || value == "false" || value == "f" || value == "off") {
                     returnValue = false;
                 } else {
                     std::ostringstream message;
                     message << "Failed to convert environment variable " << var << " from '" << value << "' to type " << typeid(T).name();
-                    message << ". Possible values are '1', 'true', 'TRUE', 'True', '0', 'false', 'FALSE', 'False'";
+                    message << ". Possible values are '1', 'true', 'TRUE', 'True', 'T', 'ON', '0', 'false', 'FALSE', 'False', 'F', 'OFF'";
```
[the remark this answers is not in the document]
moratom: Valid point IMO

include/depthai/pipeline/datatype/PipelineEvent.hpp
```diff
@@ -0,0 +1,43 @@
...
+        INPUT_BLOCK = 4,
+        OUTPUT_BLOCK = 5,
+    };
+    enum class Interval : std::int32_t { NONE = 0, START = 1, END = 2 };
+
+    PipelineEvent() = default;
+    virtual ~PipelineEvent();
+
+    int64_t nodeId = -1;
+    int32_t status = 0;
```
moratom: What does this status mean? Should it be an enum?

include/depthai/pipeline/datatype/PipelineState.hpp
```diff
@@ -0,0 +1,120 @@
...
+    struct Timing {
+        float fps = 0.0f;
+        TimingStats durationStats;
+
+        bool isValid() const {
+            return durationStats.minMicros <= durationStats.maxMicros;
+        }
+
+        DEPTHAI_SERIALIZE(Timing, fps, durationStats);
+    };
```
moratom: A bit strange naming perhaps?
`Timing = FPS + durationStats which are of type TimingStats

moratom: Perhaps we can go with just TimingStats containing everything?

include/depthai/pipeline/datatype/PipelineEvent.hpp
```diff
@@ -0,0 +1,43 @@
...
+    std::optional<uint32_t> queueSize;
+    Interval interval = Interval::NONE;
+    Type type = Type::CUSTOM;
+    std::string source;
```
moratom: What are possible options here? Should it be an enum?

src/utility/PipelineImplHelper.cpp
```diff
@@ -0,0 +1,223 @@
...
+
+    // Initialize event dispatchers
+    for(const auto& node : pipeline->getAllNodes()) {
+        auto threadedNode = std::dynamic_pointer_cast<ThreadedNode>(node);
+        if(threadedNode) {
+            threadedNode->initPipelineEventDispatcher(threadedNode->id);
+        }
+    }
+
+    // Disable event sending if no PipelineEventAggregation node is present
```
moratom: Could the dispatchers get the node ID "live" from the node instead perhaps? To not have duplicate data.

include/depthai/pipeline/ThreadedNode.hpp
```diff
@@ -45,6 +54,8 @@ class ThreadedNode : public Node {
     // check if still running
     bool isRunning() const;
 
+    bool mainLoop();
```
moratom, line 54: Let's add a docstring of what this function does

include/depthai/pipeline/node/internal/PipelineEventAggregation.hpp
```diff
@@ -0,0 +1,61 @@
...
+    void setRunOnHost(bool runOnHost);
+
+    /**
+     * Check if the node is set to run on host
+     */
+    bool runOnHost() const override;
+
+    void run() override;
+
+    PipelineEventAggregation& setTraceOutput(bool enable);
```
moratom, line 56: Maybe exposing frequency would make sense here?

include/depthai/pipeline/node/internal/PipelineStateMerge.hpp
```diff
@@ -0,0 +1,46 @@
...
+    Output outRequest{*this, {"outRequest", DEFAULT_GROUP, {{{DatatypeEnum::PipelineEventAggregationConfig, false}}}}};
+
+    /**
+     * Output message of type
+     */
+    Output out{*this, {"out", DEFAULT_GROUP, {{{DatatypeEnum::PipelineState, false}}}}};
+
+    std::shared_ptr<PipelineStateMerge> build(bool hasDeviceNodes, bool hasHostNodes);
+
+    PipelineStateMerge& setAllowConfiguration(bool allow);
```
moratom, line 43: Docstring on what this does exactly? (AKA why would we NOT allow it?)

include/depthai/utility/CircularBuffer.hpp
moratom, line 1: We should add some unit tests & documentation for this (is it thread safe, etc)

include/depthai/utility/LockingQueue.hpp
moratom, line 1: Let's expand the current tests with the new functionality

include/depthai/utility/PipelineEventDispatcher.hpp
```diff
@@ -0,0 +1,70 @@
...
+#include <unordered_map>
+
+#include "PipelineEventDispatcherInterface.hpp"
+#include "depthai/pipeline/Node.hpp"
+#include "depthai/pipeline/datatype/PipelineEvent.hpp"
+
+namespace dai {
+namespace utility {
+
+class PipelineEventDispatcher : public PipelineEventDispatcherInterface {
```
moratom, line 16: How come we have the interface separated here?

src/pipeline/datatype/PipelineState.cpp
```diff
@@ -0,0 +1,12 @@
...
+namespace dai {
+
+PipelineState::~PipelineState() = default;
+
+nlohmann::json PipelineState::toJson() const {
+    nlohmann::json j;
+    j["nodeStates"] = nodeStates;
+    return j;
+}
+}  // namespace dai
```
moratom, line 12: Do we need this explicitly?

src/pipeline/MessageQueue.cpp
```diff
@@ -122,7 +141,20 @@ bool MessageQueue::send(const std::shared_ptr<ADatatype>& msg, std::chrono::mill
...
+    return queue.tryWaitAndPush(msg, timeout, [&](LockingQueueState state, size_t size) {
+        if(pipelineEventDispatcher) {
+            switch(state) {
+                case LockingQueueState::BLOCKED:
+                    pipelineEventDispatcher->pingInputEvent(name, -1, size);
+                    break;
+                case LockingQueueState::CANCELLED:
+                    pipelineEventDispatcher->pingInputEvent(name, -2, size);
+                    break;
+                case LockingQueueState::OK:
```
moratom: No event to send here?

src/pipeline/Pipeline.cpp
```diff
@@ -767,13 +758,21 @@ void PipelineImpl::build() {
...
+                    xlinkBridges.push_back({xLinkBridge.xLinkOut->id, xLinkBridge.xLinkInHost->id});
                 }
                 auto xLinkBridge = bridgesOut[queueConnection.output];
                 queueConnection.output->unlink(queueConnection.queue);  // Unlink the original connection
                 xLinkBridge.xLinkInHost->out.link(queueConnection.queue);
             }
         }
     }
+
+    utility::PipelineImplHelper(this).setupPipelineDebuggingPost(bridgesOut, bridgesIn);
```
moratom, line 777: Why do we need this to be a class, can't it be a normal function?
Same for record&replay.

src/pipeline/node/internal/PipelineEventAggregation.cpp
```diff
@@ -0,0 +1,629 @@
...
+            }
+            auto [outState, updated] = makeOutputState(handler, currentConfig, sequenceNum++, sendEvents);
+            if(gotConfig
+               || (currentConfig.has_value() && currentConfig->repeatIntervalSeconds.has_value() && updated
+                   && (now - lastSentTime) >= std::chrono::seconds(currentConfig->repeatIntervalSeconds.value()))) {
+                lastSentTime = now;
+                out.send(outState);
+            }
+        }
+        std::this_thread::sleep_for(std::chrono::milliseconds(10));
```
moratom: Can we make this event driven instead?

src/pipeline/node/internal/PipelineEventAggregation.cpp
moratom, line 1: I think we should refactor this to be split into more functions (max 3-4 layers of indentation) & add comments to make this easier to follow

src/utility/PipelineImplHelper.cpp
```diff
@@ -0,0 +1,223 @@
...
+#endif
+                } catch(std::runtime_error& e) {
+                    Logging::getInstance().logger.warn("Could not set up record / replay: {}", e.what());
+                }
+            } else if(pipeline->enableHolisticRecordReplay || !recordPath.empty() || !replayPath.empty()) {
+                throw std::runtime_error("Holistic record/replay is only supported on RVC2 devices for now.");
+            }
+        }
+    }
+}
```
moratom, line 108: Let's add a TODO to refactor this in the future, as it has 10 levels of indentation at the deepest bit

tests/src/ondevice_tests/pipeline_debugging_rvc2_test.cpp
```diff
@@ -0,0 +1,87 @@
...
+            REQUIRE(inputState.timing.isValid());
+        }
+        for(const auto& [outputName, outputState] : nodeState.outputStates) {
+            REQUIRE(outputState.timing.isValid());
+        }
+        for(const auto& [otherName, otherTiming] : nodeState.otherTimings) {
+            REQUIRE(otherTiming.isValid());
+        }
+    }
+}
```
moratom, line 77: Can we be a bit stricter than just having them valid (some range of the expected FPS, latency, etc)

src/pipeline/Pipeline.cpp
asahtik: Done in `eaed8adfcbac8a6e7373a737b89a73e37d466c1a`

src/pipeline/Pipeline.cpp
asahtik: For protected PipelineImpl member access. It's possible to add a function as a friend but you need to declare the function in the hpp file which includes the types of arguments that are currently not defined in public headers (XLinkBridges).

moratom: Ideally we avoid this as it's quite hacky (same congnitive complexity as if it was a member function).
Maybe we can do something like:
`setupXLinkBridgesPipelineDebugging(bridges, pipelineAgregator)` 
plus if something else is needed, but not pass in the whole `Pipeline`.

include/depthai/utility/PipelineEventDispatcher.hpp
asahtik: The event dispatcher needs to be constructed in the node so it can be accessed where needed. The implementation pulls in some headers that cause a circular dependency so the interface definition needs to be separated out.

src/pipeline/datatype/PipelineState.cpp
asahtik: This outputs a cleaner json (without useless metadata) for viewer API and trace logging.

src/pipeline/Pipeline.cpp
asahtik: That'd result in functions with 5-6 arguments. IMO this is cleaner

include/depthai/utility/PipelineEventDispatcher.hpp
```diff
@@ -0,0 +1,68 @@
...
+
+    int64_t& nodeId;
+    std::unordered_map<std::string, EventStatus> events;
+    Node::Output* out = nullptr;
+
+    void checkNodeId();
+
+    uint32_t sequenceNum = 0;
+
+    std::mutex mutex;
```
aljazkonec1, line 28: I might be missing it, but the mutex lib is not included

src/utility/PipelineEventDispatcher.cpp
```diff
@@ -0,0 +1,264 @@
...
+    if(type == PipelineEvent::Type::OUTPUT && source == "pipelineEventOutput") return true;
+    return false;
+}
+
+void PipelineEventDispatcher::checkNodeId() {
+    if(nodeId == -1) {
+        throw std::runtime_error("Node ID not set on PipelineEventDispatcher");
+    }
+}
+void PipelineEventDispatcher::startEvent(PipelineEvent::Type type, const std::string& source, std::optional<uint32_t> queueSize) {
```
aljazkonec1, line 45: Should we guardrail against calling startEvent multiple times with the same `(type, source)`? Because currently, if you call it twice, it will overwrite `event.ongoing` to `true` every time. This would produce missleading event durations and other bugs.

include/depthai/utility/PipelineEventDispatcher.hpp
```diff
@@ -0,0 +1,68 @@
...
+        bool ongoing;
+    };
+
+    int64_t& nodeId;
+    std::unordered_map<std::string, EventStatus> events;
+    Node::Output* out = nullptr;
+
+    void checkNodeId();
+
+    uint32_t sequenceNum = 0;
```
aljazkonec1: I might have missed it, but sequenceNum doesnt appear to be used anywhere in this header or the PipelineEventDispatcher.cpp

src/remote_connection/RemoteConnectionImpl.cpp
moratom: Is this handled now?

asahtik: Yes

asahtik: It uses MessageQueue shared pointers directly now

include/depthai/pipeline/node/internal/PipelineStateMerge.hpp
moratom: Do we have a separate PipelineStateMerge node made when pipeline debugging is turned on through the environment variable?

include/depthai/pipeline/MessageQueue.hpp
```diff
@@ -15,6 +17,8 @@ namespace dai {
  * Thread safe queue to send messages between nodes
  */
 class MessageQueue : public std::enable_shared_from_this<MessageQueue> {
+    friend std::unordered_map<std::string, std::shared_ptr<ADatatype>> getAny(std::unordered_map<std::string, MessageQueue&> queues);
```
moratom: Can we make it a static member function instead?

src/pipeline/MessageQueue.cpp
```diff
@@ -143,6 +201,86 @@ void MessageQueue::callCallbacks(std::shared_ptr<ADatatype> message) {
...
 
+std::unordered_map<std::string, std::shared_ptr<ADatatype>> getAny(std::unordered_map<std::string, MessageQueue&> queues) {
+    std::vector<std::pair<MessageQueue&, MessageQueue::CallbackId>> callbackIds;
+    std::vector<std::pair<MessageQueue&, MessageQueue::CallbackId>> condVarIds;
+    callbackIds.reserve(queues.size());
+    condVarIds.reserve(queues.size());
+    std::unordered_map<std::string, std::shared_ptr<ADatatype>> inputs;
+
+    std::mutex inputsWaitMutex;
+    auto inputsWaitCv = std::make_shared<std::condition_variable>();
```
moratom, line 224: Needs to be a shared_ptr? Can't it be owned by this function and live on the stack?

bindings/python/src/utility/PipelineEventDispatcherBindings.cpp
```diff
@@ -0,0 +1,29 @@
...
+
+    using namespace dai::utility;
+    auto pipelineEventDispatcher = py::class_<PipelineEventDispatcher>(m, "PipelineEventDispatcher");
+
+    pipelineEventDispatcher
+        .def("startCustomEvent", &PipelineEventDispatcher::startCustomEvent, py::arg("source"), DOC(dai, utility, PipelineEventDispatcher, startCustomEvent))
+        .def("endCustomEvent", &PipelineEventDispatcher::endCustomEvent, py::arg("source"), DOC(dai, utility, PipelineEventDispatcher, endCustomEvent))
+        .def("pingCustomEvent", &PipelineEventDispatcher::pingCustomEvent, py::arg("source"), DOC(dai, utility, PipelineEventDispatcher, pingCustomEvent))
+        .def("inputBlockEvent", &PipelineEventDispatcher::inputBlockEvent, DOC(dai, utility, PipelineEventDispatcher, inputBlockEvent))
+        .def("outputBlockEvent", &PipelineEventDispatcher::outputBlockEvent, DOC(dai, utility, PipelineEventDispatcher, outputBlockEvent))
```
aljazkonec1: The return type of `PipelineEventDispatcher::BlockPipelineEvent` is `BlockPipelineEvent` class, but that class is not bound to python so this will fail.

bindings/python/src/utility/PipelineEventDispatcherBindings.cpp
```diff
@@ -0,0 +1,29 @@
...
+        .def("customBlockEvent", &PipelineEventDispatcher::customBlockEvent, py::arg("source"), DOC(dai, utility, PipelineEventDispatcher, customBlockEvent));
```
aljazkonec1: Same comment as above.

bindings/python/src/utility/PipelineEventDispatcherBindings.cpp
```diff
@@ -0,0 +1,29 @@
...
+    callstack->pop();
+    cb(m, pCallstack);
+    ///////////////////////////////////////////////////////////////////////
+    ///////////////////////////////////////////////////////////////////////
+    ///////////////////////////////////////////////////////////////////////
+
+    using namespace dai::utility;
+    auto pipelineEventDispatcher = py::class_<PipelineEventDispatcher>(m, "PipelineEventDispatcher");
+
+    pipelineEventDispatcher
```
aljazkonec1: Need to define`py::init<>()` and any properties that the class has

include/depthai/pipeline/datatype/PipelineEvent.hpp
```diff
@@ -0,0 +1,48 @@
+#pragma once
+
+#include <vector>
+
```
aljazkonec1, line 4: Missing standar header for `std::string`

include/depthai/pipeline/datatype/PipelineEvent.hpp
```diff
@@ -0,0 +1,48 @@
...
+    PipelineEvent() = default;
+    virtual ~PipelineEvent();
+
+    int64_t nodeId = -1;
+    Status status = Status::SUCCESS;  // Used for input status reporting (if it is blocked)
+    std::optional<uint32_t> queueSize;
+    Interval interval = Interval::NONE;
+    Type type = Type::CUSTOM;
+    std::string source;  // e.g., input/output/custom event name - used in conjunction with type to identify the event
+
```
aljazkonec1, line 39: Missing override function:  `DatatypeEnum getDatatype() const override`. Without it, the PipelineEvent dataype will be enumerated as generic Buffer type

bindings/python/src/pipeline/datatype/PipelineEventBindings.cpp
```diff
@@ -0,0 +1,63 @@
...
+    cb(m, pCallstack);
+    // Actual bindings
+    ///////////////////////////////////////////////////////////////////////
+    ///////////////////////////////////////////////////////////////////////
+    ///////////////////////////////////////////////////////////////////////
+
+    pipelineEventType.value("CUSTOM", PipelineEvent::Type::CUSTOM)
+        .value("LOOP", PipelineEvent::Type::LOOP)
+        .value("INPUT", PipelineEvent::Type::INPUT)
+        .value("OUTPUT", PipelineEvent::Type::OUTPUT);
```
aljazkonec1: Any reason INPUT_BLOCK and OUTPUT_BLOCK are not binded?

src/pipeline/MessageQueue.cpp
```diff
@@ -143,6 +201,86 @@ void MessageQueue::callCallbacks(std::shared_ptr<ADatatype> message) {
...
+    }
+
+    // Check if any messages already present
+    bool hasAnyMessages = false;
+    for(auto& kv : queues) {
+        if(kv.second.has()) {
+            hasAnyMessages = true;
+            break;
+        }
+    }
```
moratom: This can go upfront and we skip setting up the callbacks, etc in such case.

src/pipeline/node/internal/PipelineEventAggregation.cpp
```diff
@@ -0,0 +1,671 @@
...
+                                                                 bool sendEvents) {
+    auto outState = std::make_shared<PipelineState>();
+    bool updated = handler.getState(outState, sendEvents);
+    outState->sequenceNum = sequenceNum;
+    outState->configSequenceNum = currentConfig.has_value() ? currentConfig->sequenceNum : 0;
+    outState->setTimestamp(std::chrono::steady_clock::now());
+    outState->tsDevice = outState->ts;
+
+    // Filter state based on current configuration
+    if(!currentConfig->nodes.empty()) {
```
aljazkonec1: Need to check that currentConfig has value

src/pipeline/MessageQueue.cpp
```diff
@@ -143,6 +201,86 @@ void MessageQueue::callCallbacks(std::shared_ptr<ADatatype> message) {
...
+
+    // Remove callbacks
+    for(auto& [input, callbackId] : callbackIds) {
+        input.removeCallback(callbackId);
+    }
+    // Remove condition variables
+    for(auto& [input, condVarId] : condVarIds) {
+        input.removeCondVar(condVarId);
+    }
+
```
moratom, line 265: This look flaky in case any of the above throws, we'll "leak".

src/pipeline/ThreadedNode.cpp
```diff
@@ -18,6 +21,7 @@ ThreadedNode::ThreadedNode() {
         level = Logging::parseLevel(envLevel);
     }
     pimpl->logger->set_level(level);
+    pipelineEventDispatcher = std::make_unique<utility::PipelineEventDispatcher>(this->id, &pipelineEventOutput);
```
moratom, line 25: Should we only construct it when pipeline debugging is turned on? For performance.

tests/src/ondevice_tests/img_transformation_test.cpp
```diff
@@ -63,7 +63,7 @@ inline std::array<std::array<float, 3>, 3> matmul(std::array<std::array<float, 3
 TEST_CASE("ImgTransformation in ImgFrame") {
     dai::Pipeline pipeline;
     auto cam = pipeline.create<dai::node::Camera>()->build();
-    auto camOut = cam->requestOutput({1300, 200}, dai::ImgFrame::Type::NV12);
+    auto camOut = cam->requestOutput({600, 400}, dai::ImgFrame::Type::NV12);
```
moratom, line 66: Why was this needed?

include/depthai/pipeline/datatype/PipelineState.hpp
```diff
@@ -0,0 +1,120 @@
+#pragma once
+
+#include <vector>
+
+#include "depthai/pipeline/datatype/Buffer.hpp"
+#include "depthai/pipeline/datatype/PipelineEvent.hpp"
+
```
aljazkonec1, line 9: Missing standard header include for `std::unordered_map` and `std::string`

include/depthai/pipeline/MessageQueue.hpp
```diff
@@ -157,6 +178,8 @@ class MessageQueue : public std::enable_shared_from_this<MessageQueue> {
      */
     CallbackId addCallback(const std::function<void()>& callback);
 
+    CallbackId addCondVar(std::shared_ptr<std::condition_variable> cv);
```
aljazkonec1, line 199: Docstring

include/depthai/pipeline/MessageQueue.hpp
```diff
@@ -165,6 +188,8 @@ class MessageQueue : public std::enable_shared_from_this<MessageQueue> {
      */
     bool removeCallback(CallbackId callbackId);
 
+    bool removeCondVar(CallbackId condVarId);
```
aljazkonec1, line 215: Docstring

src/pipeline/node/internal/XLinkOutHost.cpp
```diff
@@ -53,9 +53,13 @@ void XLinkOutHost::run() {
...
         };
-        while(isRunning()) {
+        while(mainLoop()) {
             try {
-                auto outgoing = in.get();
+                std::shared_ptr<ADatatype> outgoing;
+                {
+                    auto blockEvent = this->inputBlockEvent();
+                    outgoing = in.get();
+                }
```
aljazkonec1: This is more just for my own understanding, but why is this in its own scope?

PipelineDebugging.md
```diff
@@ -0,0 +1,211 @@
...
+    State state = State::IDLE;
+    std::vector<DurationEvent> events;
+    std::unordered_map<std::string, OutputQueueState> outputStates;
+    std::unordered_map<std::string, InputQueueState> inputStates;
+    Timing inputsGetTiming;
+    Timing outputsSendTiming;
+    Timing mainLoopTiming;
+    std::unordered_map<std::string, Timing> otherTimings;
+};
+```
```
moratom, line 139: Needs to be updated before merge as the names have changed slightly, best to go over the whole document)

src/pipeline/ThreadedNode.cpp
```diff
@@ -86,4 +90,19 @@ bool ThreadedNode::isRunning() const {
     return running;
 }
 
+bool ThreadedNode::mainLoop() {
+    this->pipelineEventDispatcher->pingMainLoopEvent();
+    return isRunning();
+}
+
+utility::PipelineEventDispatcherInterface::BlockPipelineEvent ThreadedNode::blockEvent(PipelineEvent::Type type, const std::string& source) {
```
aljazkonec1: I would add docstrings.

src/utility/PipelineImplHelper.cpp
```diff
@@ -0,0 +1,221 @@
+#include "PipelineImplHelper.hpp"
+
+#include "pipeline/node/internal/PipelineEventAggregation.hpp"
+#include "pipeline/node/internal/PipelineStateMerge.hpp"
+#include "pipeline/node/internal/XLinkIn.hpp"
+#include "pipeline/node/internal/XLinkInHost.hpp"
+#include "pipeline/node/internal/XLinkOut.hpp"
+#include "pipeline/node/internal/XLinkOutHost.hpp"
```
aljazkonec1, line 8: Seem to be unused.

PipelineDebugging.md
```diff
@@ -0,0 +1,211 @@
+# Pipeline Debugging
```
aljazkonec1, line 1: I would add a section on how to use pipeline_graph and a link to the repo.

examples/cpp/Misc/PipelineDebugging/get_pipeline_state.cpp
```diff
@@ -0,0 +1,107 @@
+#include <iostream>
```
aljazkonec1: I would rewrite this example to make it more clear what it does. Having it only work with stereo cameras is a bit annoying since pipeline debugging applies to all cameras not just stereo depth cameras.

If you run this example with pipeline_graph the actual graph is not interesting, there is no blockages or issues. I would refactor the example to be something along the lines of:
1) Have a camera output at 30 FPS
2) Create a (Threaded)HostNode that sends data at 15 FPS
3) Create a (Threaded)HostNode which would have as input the camera and the first host node, the node logic would be to sleep for X ms, then send the camera stream to an outputQueue

When you run the debug, you could see the FPS of each node, the processing times, there would be a blocking queue, etc etc. And we can add commented `.setBlocking(False)` commands in the code where a user could turn blocking off and see how the pipeline behaves differently.

cc @moratom for opinion

examples/cpp/Misc/PipelineDebugging/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+project(pipeline_debugging_examples)
+cmake_minimum_required(VERSION 3.10)
+
+## function: dai_add_example(example_name example_src enable_test use_pcl)
+## function: dai_set_example_test_labels(example_name ...)
+
+dai_add_example(get_pipeline_state get_pipeline_state.cpp ON OFF)
+dai_add_example(node_pipeline_events node_pipeline_events.cpp ON OFF)
```
aljazkonec1, line 8: I would add an example to make it more clear what pipeline debugging does. I would do something along the lines of:

1) Have a camera output at 30 FPS
2) Create a (Threaded)HostNode that sends data at 15 FPS
3) Create a (Threaded)HostNode which would have as input the camera and the first host node, the node logic would be to sleep for X ms, then send the camera stream to an outputQueue

When you run the pipeline_graph, you could see the FPS of each node, the processing times, there would be a blocking queue, etc etc. And we can add commented `.setBlocking(False)` commands in the code where a user could turn blocking off and see how the pipeline behaves differently.

src/pipeline/MessageQueue.cpp
asahtik: It's safer because it's not guaranteed that this instance of the cv will outlive the callback (for example if the thread calling getAny gets killed before the MessageQueue closes)

examples/cpp/Misc/PipelineDebugging/get_pipeline_state.cpp
```diff
@@ -0,0 +1,107 @@
...
+#include <opencv2/opencv.hpp>
+
+#include "depthai/depthai.hpp"
+
+int main() {
```
aljazkonec1, line 6: Having the example only work with stereo cameras is a bit annoying since pipeline debugging applies to all cameras not just stereo depth cameras.

src/pipeline/MessageQueue.cpp
asahtik: This check is needed after setting callbacks to make sure nothing is skipped. I can also add this check at the beginning to skip adding callbacks, it'll make the code slightly more complex though

src/pipeline/ThreadedNode.cpp
asahtik: PipelineEventDispatcher needs to be created at node construction time so that inputs and outputs have access to it when they are created but pipeline debugging can be enabled any time between pipeline initialization and pipeline build. If pipeline debugging is disabled there is not much overhead because dispatcher functions return immediately.

tests/src/ondevice_tests/img_transformation_test.cpp
asahtik: This test failed on OAK-D W

src/pipeline/node/DetectionParser.cpp
```diff
@@ -413,9 +413,13 @@ void DetectionParser::run() {
...
     using namespace std::chrono;
-    while(isRunning()) {
+    while(mainLoop()) {
         auto tAbsoluteBeginning = steady_clock::now();
-        std::shared_ptr<dai::NNData> sharedInputData = input.get<dai::NNData>();
+        std::shared_ptr<dai::NNData> sharedInputData;
+        {
+            auto blockEvent = this->inputBlockEvent();
+            sharedInputData = input.get<dai::NNData>();
+        }
```
aljazkonec1, line 422: I have some concerns about the usability of this approach with scopes. Most issues with pipelines (latency, blockage, ...) happen in custom nodes that customers develop. With this approach, if a user updates to v3.3.0+, it would be good to have full support of the event sending without the need to update the while loop of all nodes.

 One more issue is if user uses python, would the customer need to manually send the event data as there would be no scope? Or would they need to wrap code in a `with` statement?

 One possible solution could be to send the event data per `.get()` and `.send()` call? It would also allow us to visualize get and send times per output / input of each node.

aljazkonec1 (changes requested):
Thanks! I went over the code and left some comments. I ran pipeline_graph with the your branch and its much much more usable.  

I have a general question:  When I was running debugging on a larger pipeline it got confusing to know which node is which from their name. Would it be plausible to have a name parameter for nodes so that a user could rename a node to make it more clear what it does or would this be too much work?

src/utility/PipelineEventDispatcher.cpp
asahtik: I had a check for that but I removed it because it's not necessary that a started event finishes. So if you start a new event when the previous one is ongoing, that one is cancelled and a new one is started.

src/pipeline/node/DetectionParser.cpp
asahtik: Sending events per gets and sends is already implemented. These block events are needed to determine how much time per loop a node is spending on one of the tasks: getting inputs, processing, sending outputs.

For nodes with multiple inputs or outputs this is impossible to do without explicitly defining input and output blocks.

If the customer doesn't implement these blocks, they will still have stats and states per I/O, but won't have node states and G,S,P (in pipeline graph) timings.

src/pipeline/Pipeline.cpp
```diff
@@ -808,6 +817,23 @@ void PipelineImpl::start() {
         const auto weak = std::weak_ptr<PipelineImpl>(shared);
         defaultDevice->pipelinePtr = weak;
     }
+
+    // Setup pipeline state trace logging if enabled
+    if(buildingOnHost && utility::getEnvAs<bool>("DEPTHAI_PIPELINE_DEBUGGING", false)) {
```
moratom, line 822: Needs to be added to README.md

moratom (commented):
I would switch to a simpler FPS counting mechanism, other than that LGTM.

PR #1473, opened 2025-09-26
V3 dcl remove data start calib

MaticTonin:
## Purpose
- adding factories to simplify API command
- adding new e2e tests based on the real data
- allow different resolution between left and right frame


MaticTonin:
The PR was seperated into more commits:

https://github.com/luxonis/depthai-core/pull/1477
Then
https://github.com/luxonis/depthai-core/pull/1478
Then
https://github.com/luxonis/depthai-core/pull/1479

PR #1475, opened 2025-09-30
Add DEPTHAI_INSTALL_PREFIX to depthaiConfig.cmake, install libusb DLL for every depthai example

lnotspotl:
The purpose of this PR is to introduce a new depthai-specific cmake constant called `DEPTHAI_INSTALL_PREFIX` to facilitate dependency installation for projects linking against depthai, especially those on Windows. This is paramount for users who build depthai with `BUILD_SHARED_LIBS` turned ON as the folder containing generated `.dll` libraries either has to be added to PATH or libraries contained within have to be copied over to the very location of a dependent executable.

The depthai-core-example makes use of this new variable, a PR for that project is opened [here](https://github.com/luxonis/depthai-core-example/pull/9).


CMakeLists.txt
```diff
@@ -1297,4 +1274,8 @@ if(DEPTHAI_INSTALL)
         DESTINATION "${CMAKE_INSTALL_LIBDIR}/cmake/${PROJECT_NAME}"
     )
 
+    # Install vcpkg libraries that we link against dynamically
+    if(DEPTHAI_ENABLE_LIBUSB)
+        install(IMPORTED_RUNTIME_ARTIFACTS usb-1.0)
```
Serafadam: I think we also need to do that for dynamic calibration library? I also think this makes `add_runtime_dependencies` obsolete (at least that was the case when building deb packages for ROS)

lnotspotl: You're making a really good point here. The issue was that when I tried installing the project, `libusb-1.0.so` would not be present as a runtime dependency but the dynamic_calibration library would. Turns out there was a bug in the `add_runtime_dependencies` macro that would override the `required_dll_files` variable every time it was invoked, thereby causing that only the last added runtime dependency would be added.

I guess this macro is still needed for libraries that we link against with PRIVATE visibility.

moratom (commented):
Test if the OpenCV is installed by choco and DepthAI core as a shared library, will it work? Can we make it work?

lnotspotl:
Choco installs OpenCV without modifying the PATH env variable, so we do need to copy its dlls over for all our examples. With our current cmake setup, this happens automatically and examples work even when depthai-core is compiled as a shared library.

image

PR #1476, opened 2025-09-30
Fix X_LINK_LOCAL_SHDMEM protocol mode only working once and then requiring rm /tmp/xlink.sock or reboot.

zrezke:
## Purpose

* Fix X_LINK_LOCAL_SHDMEM protocol mode only working once and then requiring rm /tmp/xlink.sock or reboot.
* [FW] Enable depthai device usage from device side during X_LINK_GATE_SETUP state.

## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

Tested creating dai.Device locally multiple times in a row.
Tested as described in [XLink PR](https://github.com/luxonis/XLink/pull/98) as well

### DIFF
--- a/cmake/depthaiDependencies.cmake
+++ b/cmake/depthaiDependencies.cmake
@@ -177,7 +177,7 @@ else()
     FetchContent_Declare(
         XLink
         GIT_REPOSITORY https://github.com/luxonis/XLink.git
-        GIT_TAG        3743bd8c3738dc6875169b32d49f9043ec3a4b56
+        GIT_TAG        fcc46245f61d99c4dcd8c33756e0fe19ede8f230
     )
 
     FetchContent_MakeAvailable(


src/device/DeviceGate.cpp
```diff
@@ -52,8 +52,12 @@ DeviceGate::DeviceGate(const DeviceInfo& deviceInfo) : deviceInfo(deviceInfo) {
     } else {
         throw std::runtime_error("Unknown platform");  // Should never happen
     }
+    auto gateHost = deviceInfo.name;
+    if (deviceInfo.protocol == X_LINK_LOCAL_SHDMEM) {
+        gateHost = "127.0.0.1";
+    }
     // Discover and connect
-    pimpl->cli = std::make_unique<httplib::Client>(deviceInfo.name, DEFAULT_PORT);
+    pimpl->cli = std::make_unique<httplib::Client>(gateHost, DEFAULT_PORT);
```
moratom, line 60: How come this is needed?

zrezke: We still have to connect to gate to boot the fw - but the `deviceInfo.name == /tmp/xlink.sock`, so have to overwrite that. Perhaps X_LINK_LOCAL_SHDMEM should be removed from discovery altogether.

themarpe:
@zrezke with this in place, does shared memory actually work as expected? @moratom whats the main reason to closing this?

PR #1477, opened 2025-09-30
V3 dcl adding factories; Change in API

MaticTonin:
Reason for adding factories is to clear API with hard to remember Commands.
Commands can still be used the same way as previously this is just a simplification function.


moratom (approved):
LGTM

PR #1483, opened 2025-10-02
Enable exposure time limit on RVC4

jakgra:
## Purpose
Enable exposure time limit on RVC4 that wasn't supported before this PR

## Specification
None / not applicable

Needs OS 1.19 for exposure time limit to work. Other OS versions supported for all other functionalities.

## Deployment Plan
Needs OS 1.19 for the functionality to work. Otherwise it is a no op.

## Testing & Validation
Verified it is working with updated cam_test.py


moratom (approved):
Thanks!

PR #1484, opened 2025-10-03
Pointcloud example failing Windows Build

MaticTonin:
One of the RGBD examples was failing Windows build, due to designated initializers being added (notation for Cmake 20). 

See error here:
https://github.com/luxonis/depthai-core/pull/1461/checks

This was solved in this PR as seen here:
https://github.com/luxonis/depthai-core/actions/runs/18210124854


aljazdu:
LGTM, added the testable label to run the checks

PR #1492, opened 2025-10-10
Feature/snaps events v2 integration

aljazdu:
## Purpose
Update snaps and events to V2

## Specification

None / not applicable
Changes in the uploading of the events and snaps to the Luxonis Hub, the protobuf definition used in connection to events and snaps, examples in C++ & Python

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation
This was tested using the stg and prod versions of the Hub, by uploading different snaps and events

### DIFF
--- a/README.md
+++ b/README.md
@@ -217,7 +217,7 @@ The following environment variables can be set to alter default behavior of the 
 | DEPTHAI_CRASHDUMP_TIMEOUT | Specifies the duration in milliseconds to wait for device reboot when obtaining a crash dump. Crash dump retrieval disabled if 0. |
 | DEPTHAI_ENABLE_ANALYTICS_COLLECTION | Enables automatic analytics collection (pipeline schemas) used to improve the library |
 | DEPTHAI_DISABLE_CRASHDUMP_COLLECTION | Disables automatic crash dump collection used to improve the library |
-| DEPTHAI_HUB_EVENTS_BASE_URL | URL for the Luxonis Hub |
+| DEPTHAI_HUB_EVENTS_BASE_URL | URL for events of the Luxonis Hub |
 | DEPTHAI_HUB_API_KEY | API key for the Luxonis Hub |
 | DEPTHAI_ZOO_INTERNET_CHECK | (Default) 1 - perform internet check, if available, download the newest model version 0 - skip internet check and use cached model |
 | DEPTHAI_ZOO_INTERNET_CHECK_TIMEOUT | (Default) 1000 - timeout in milliseconds for the internet check |
--- a/examples/cpp/Events/events.cpp
+++ b/examples/cpp/Events/events.cpp
@@ -15,9 +15,8 @@ cv::Rect frameNorm(const cv::Mat& frame, const dai::Point2f& topLeft, const dai:
 int main() {
     dai::Pipeline pipeline(true);
 
-    // Enter your hub team's api-key
+    // Set your Hub team's api-key using the environment variable DEPTHAI_HUB_API_KEY. Or use the EventsManager setToken() method.
     auto eventsManager = std::make_shared<dai::utility::EventsManager>();
-    eventsManager->setToken("");
 
     auto camRgb = pipeline.create<dai::node::Camera>()->build();
     auto detectionNetwork = pipeline.create<dai::node::DetectionNetwork>();
@@ -34,7 +33,7 @@ int main() {
     pipeline.start();
 
     while(pipeline.isRunning()) {
-        if(cv::waitKey(1) == 'q') {
+        if(cv::waitKey(1) != -1) {
             break;
         }
 
--- a/examples/python/Events/events.py
+++ b/examples/python/Events/events.py
@@ -8,9 +8,8 @@ import time
 
 # Create pipeline
 with dai.Pipeline() as pipeline:
-    # Enter your hub team's api-key
+    # Set your Hub team's api-key using the environment variable DEPTHAI_HUB_API_KEY. Or use the EventsManager setToken() method.
     eventMan = dai.EventsManager()
-    eventMan.setToken("")
 
     cameraNode = pipeline.create(dai.node.Camera).build()
     detectionNetwork = pipeline.create(dai.node.DetectionNetwork).build(cameraNode, dai.NNModelDescription("yolov6-nano"))
@@ -31,7 +30,7 @@ with dai.Pipeline() as pipeline:
 
 
     while pipeline.isRunning():
-        if cv2.waitKey(1) == ord("q"):
+        if cv2.waitKey(1) != -1:
             pipeline.stop()
             break
 
--- a/examples/cpp/Events/events_file_group.cpp
+++ b/examples/cpp/Events/events_file_group.cpp
@@ -15,9 +15,8 @@ cv::Rect frameNorm(const cv::Mat& frame, const dai::Point2f& topLeft, const dai:
 int main() {
     dai::Pipeline pipeline(true);
 
-    // Enter your hub team's api-key
+    // Set your Hub team's api-key using the environment variable DEPTHAI_HUB_API_KEY. Or use the EventsManager setToken() method.
     auto eventsManager = std::make_shared<dai::utility::EventsManager>();
-    eventsManager->setToken("");
 
     auto camRgb = pipeline.create<dai::node::Camera>()->build();
     auto detectionNetwork = pipeline.create<dai::node::DetectionNetwork>();
@@ -35,7 +34,7 @@ int main() {
 
     int counter = 0;
     while(pipeline.isRunning()) {
-        if(cv::waitKey(1) == 'q') {
+        if(cv::waitKey(1) != -1) {
             break;
         }
 
--- a/examples/python/Events/events_file_group.py
+++ b/examples/python/Events/events_file_group.py
@@ -7,9 +7,8 @@ import numpy as np
 
 # Create pipeline
 with dai.Pipeline() as pipeline:
-    # Enter your hub team's api-key
+    # Set your Hub team's api-key using the environment variable DEPTHAI_HUB_API_KEY. Or use the EventsManager setToken() method.
     eventMan = dai.EventsManager()
-    eventMan.setToken("")
 
     cameraNode = pipeline.create(dai.node.Camera).build()
     detectionNetwork = pipeline.create(dai.node.DetectionNetwork).build(cameraNode, dai.NNModelDescription("yolov6-nano"))
@@ -31,7 +30,7 @@ with dai.Pipeline() as pipeline:
 
     counter = 0
     while pipeline.isRunning():
-        if cv2.waitKey(1) == ord("q"):
+        if cv2.waitKey(1) != -1:
             pipeline.stop()
             break
 


include/depthai/utility/EventsManager.hpp
```diff
@@ -147,33 +165,93 @@ class EventsManager {
...
+
+    UploadRetryPolicy uploadRetryPolicy;
+
+    static constexpr int eventValidationNameLength = 56;
+    static constexpr int eventValidationMaxTags = 20;
+    static constexpr int eventValidationTagLength = 56;
+    static constexpr int eventValidationMaxExtras = 25;
+    static constexpr int eventValidationExtraKeyLength = 40;
+    static constexpr int eventValidationExtraValueLength = 100;
+    static constexpr int eventValidationMaxAssociateFiles = 20;
```
moratom: Nit: we use UPPER_CASE for compile-time evaluated variables

include/depthai/utility/EventsManager.hpp
```diff
@@ -120,12 +143,7 @@ class EventsManager {
...
     void setVerifySsl(bool verifySsl);
 
-    /**
-     * Check if the device is connected to Hub. Performs a simple GET request to the URL/health endpoint
-     * @return bool
-     */
-    bool checkConnection();
-
+    // TO DO: Should be private?
     /**
```
moratom: How come we removed this?

include/depthai/utility/EventsManager.hpp
```diff
@@ -11,102 +12,124 @@
...
 
    private:
-    std::string fileName;
+    /**
+     * Calculate SHA256 checksum for the given data
+     * @param data Data for checksum calculation
+     * @return checksum string
+     */
+    std::string calculateSHA256Checksum(const std::string& data);
+
```
moratom: Doesn't need to be a member function?

include/depthai/utility/EventsManager.hpp
```diff
@@ -11,102 +12,124 @@
...
     std::string mimeType;
+    std::string fileName;
     std::string data;
-    EventDataType type;
+    uint64_t size;
+    std::string checksum;
+    proto::event::PrepareFileUploadClass classification;
+    friend class EventsManager;
```
moratom, line 47: Why does the EventsManager need to be a friend here?

include/depthai/utility/EventsManager.hpp
```diff
@@ -11,102 +12,124 @@
...
+};
+
+class FileGroup {
+   public:
+    void clearFiles();
+    void addFile(std::string fileName, std::string data, std::string mimeType);
+    void addFile(std::string fileName, std::string filePath);
```
moratom:
```suggestion
    void addFile(std::string fileName, std::filesystem::path filePath);
```

include/depthai/utility/EventsManager.hpp
```diff
@@ -11,102 +12,124 @@
...
+     * @param imgFrame ImgFrame to send
+     * @param imgDetections ImgDetections to sent
+     * @param tags List of tags to send
+     * @param extras Extra data to send
+     * @param deviceSerialNo Device serial number
+     * @return bool
      */
-    void setSourceAppIdentifier(const std::string& sourceAppIdentifier);
+    bool sendSnap(const std::string& name,
+                  const std::string& fileName,
```
moratom: Do we need both name and fileName be required arguments?
Where are the two strings used afterwards?

include/depthai/utility/EventsManager.hpp
```diff
@@ -11,102 +12,124 @@
...
+                  const std::shared_ptr<ImgFrame> imgFrame,
+                  const std::shared_ptr<ImgDetections> imgDetections,
```
moratom: I would have this optional with the default being `std::nullopt` to allow the usage of the functions with frames only

examples/python/Events/events.py
```diff
@@ -8,37 +8,84 @@
 
 # Create pipeline
 with dai.Pipeline() as pipeline:
-    # Define sources and outputs
-    camRgb = pipeline.create(dai.node.Camera).build()
-    # Properties
+    # Enter you hub team's api-key
+    eventMan = dai.EventsManager()
+    eventMan.setToken("")
```
moratom: Needed?

examples/python/Events/events.py
```diff
@@ -8,37 +8,84 @@
...
+        borderDetectionsList = []
+        for detection in inDet.detections:
+            if detection.confidence > 0.5 and detection.confidence < 0.6:
+                borderDetectionsList.append(detection)
+        
+        # Are there any border detections
+        if len(borderDetectionsList) > 0:
+            borderDetections = dai.ImgDetections()
+            borderDetections.detections = borderDetectionsList
+            fileName = f"ImageDetection_{counter}"
```
moratom: I think we should ideally avoid putting a hard requirement for the customers to name each snap here

examples/python/Events/events.py
```diff
@@ -8,37 +8,84 @@
...
+
+            fileGroup.clearFiles();
+            fileGroup.addImageDetectionsPair(fileName, inRgb, borderDetections)
+            eventMan.sendSnap("ImageDetection", fileGroup, ["EventsExample", "Python"], {"key_0" : "value_0", "key_1" : "value_1", "key_2" : "value_2"}, "")
```
moratom: Can we use the "leaner" API in the examples (without the filegroup)?

Or better yet, have one example use the file group and the other use the leaner API

examples/python/Events/events.py
```diff
@@ -8,37 +8,84 @@
 
 # Create pipeline
 with dai.Pipeline() as pipeline:
-    # Define sources and outputs
-    camRgb = pipeline.create(dai.node.Camera).build()
-    # Properties
+    # Enter you hub team's api-key
+    eventMan = dai.EventsManager()
+    eventMan.setToken("")
+    eventMan.setLogResponse(False)
```
moratom: Is this the default already?

examples/cpp/Events/events.cpp
moratom, line 1: (most comments from the Python example apply)

src/utility/EventsManager.cpp
```diff
@@ -171,165 +541,188 @@ void EventsManager::sendEventBuffer() {
...
+    }
+    for(int i = 0; i < inputEvent.tags_size(); ++i) {
+        const auto& tag = inputEvent.tags(i);
+        if(tag.empty()) {
+            logger::error("Invalid event tags: tag[{}] empty string", i);
+            return false;
         }
-        if(send == false) {
-            logger::error("Only image files are allowed for snaps");
+        if(tag.length() > 56) {
```
moratom: better to make 56 a constexpr

src/utility/EventsManager.cpp
```diff
@@ -62,80 +145,374 @@ EventData::EventData(const std::shared_ptr<EncodedFrame>& encodedFrame, std::str
...
     ss.write((const char*)encodedFrame->getData().data(), encodedFrame->getData().size());
     data = ss.str();
-    mimeType = "image/jpeg";
+    size = data.size();
+    checksum = calculateSHA256Checksum(data);
 }
 
-EventData::EventData(const std::shared_ptr<NNData>& nnData, std::string fileName)
-    : fileName(std::move(fileName)), mimeType("application/octet-stream"), type(EventDataType::NN_DATA) {
+FileData::FileData(const std::shared_ptr<NNData>& nnData, std::string fileName)
```
moratom: I think we should remove this until we have a proper way to use this (can be commented out)

src/utility/EventsManager.cpp
```diff
@@ -62,80 +145,374 @@ EventData::EventData(const std::shared_ptr<EncodedFrame>& encodedFrame, std::str
...
+    }
+    if(!snapAnnotation.SerializeToString(&data)) {
+        logger::error("Failed to serialize SnapAnnotations proto object to string");
+        return;
+    }
+    size = data.size();
+    checksum = calculateSHA256Checksum(data);
+}
+
+bool FileData::toFile(const std::string& inputPath) {
```
moratom: std::filesystem::path on the input (if the function is needed still)

src/utility/EventsManager.cpp
```diff
@@ -17,43 +19,124 @@ namespace dai {
...
+
+FileData::FileData(std::string data, std::string fileName, std::string mimeType)
+    : mimeType(std::move(mimeType)),
+      fileName(std::move(fileName)),
+      data(std::move(data)),
+      size(data.size()),
+      checksum(calculateSHA256Checksum(data)),
+      classification(proto::event::PrepareFileUploadClass::UNKNOWN_FILE) {}
+
+FileData::FileData(std::string filePath, std::string fileName) : fileName(std::move(fileName)) {
```
moratom: std::filesystem::path

examples/python/Events/events.py
aljazdu: As suggested by the Hub team, the event/snap name is a must. I will however rename it to be more clear, e.g. "LowConfidenceDetection".
As for the filename, we can make it optional, so the customer does not need to name each file.

include/depthai/utility/EventsManager.hpp
aljazdu: This has been fixed

include/depthai/utility/EventsManager.hpp
aljazdu: Fixed so it's no longer a member function

include/depthai/utility/EventsManager.hpp
aljazdu: std::filesystem::path is now used

src/utility/EventsManager.cpp
aljazdu: This has been fixed

include/depthai/utility/EventsManager.hpp
aljazdu: In the current implementation it's accessing FileData's private members directly

examples/python/Events/events.py
aljazdu: Removed this, as it was the default already

src/utility/EventsManager.cpp
aljazdu: Commented out in all occurrences. Can be used at a later point as suggested.

examples/python/Events/events.py
aljazdu: I left it here to "alert" the user that the token must be set - specifically in the peripheral mode. I can remove it, if this doesn't make sense

examples/python/Events/events.py
aljazdu: Added another simplified example with sendSnap() triggered by key press. This example uses the simplified sendSnap() overload, so no FileGroups are created before sending the snap

include/depthai/utility/EventsManager.hpp
aljazdu: This was removed because on the Hub's side the health endpoint (which was used to check the connection) had been removed.

include/depthai/utility/EventsManager.hpp
aljazdu: ImgDetections are now optional. sendSnap() can be used with only ImgFrame

include/depthai/utility/EventsManager.hpp
aljazdu: filleName is now optional and is not needed to be chosen

examples/python/Events/events.py
aljazdu: When adding files such as ImgFrame, ImgDetections, EncodedImg, or adding pairs of Image & Detection, the fileName is now optional. If no fileName is chosen, a default value is selected.

src/utility/EventsManager.cpp
aljazdu: The function is still used, changed it so it uses std::filesystem::path

README.md
```diff
@@ -217,6 +217,7 @@ The following environment variables can be set to alter default behavior of the
 | DEPTHAI_CRASHDUMP_TIMEOUT | Specifies the duration in milliseconds to wait for device reboot when obtaining a crash dump. Crash dump retrieval disabled if 0. |
 | DEPTHAI_ENABLE_ANALYTICS_COLLECTION | Enables automatic analytics collection (pipeline schemas) used to improve the library |
 | DEPTHAI_DISABLE_CRASHDUMP_COLLECTION | Disables automatic crash dump collection used to improve the library |
+| DEPTHAI_HUB_EVENTS_BASE_URL | URL for the Luxonis Hub |
```
moratom: Is this the base URL for Hub applied also for models or just events?
If it's just events maybe "URL for events of the Luxonis Hub

examples/python/Events/events.py
```diff
@@ -8,37 +8,67 @@
...
+    # nn data, being the bounding box locations, are in <0..1> range - they need to be normalized with frame width/height
+    def frameNorm(frame, bbox):
+        normVals = np.full(len(bbox), frame.shape[0])
+        normVals[::2] = frame.shape[1]
+        return (np.clip(np.array(bbox), 0, 1) * normVals).astype(int)
 
 
-    eventSent = False
     while pipeline.isRunning():
+        if cv2.waitKey(1) == ord("q"):
```
moratom: (applies to all examples)
Better to do `key = cv2.waitKey(1)` and then check for all keys, otherwise this is not responsive.

examples/python/Events/events.py
moratom: I think we should comment out and include in the comment to either use this or the ENV variable.
(comment out mostly so example can be used as is if the ENV variable is used)

Applies to all four examples.

moratom (approved):
Thanks, left a few final nitpicks, but otherwise LGTM.

README.md
aljazdu: I believe it is only for the events when looking at the default url value.
Will change it to the suggested

src/utility/EventsManager.cpp
aljazdu: This is outdated and has been resolved already.

PR #1497, opened 2025-10-16
RemoteConnection - message priority 

lnotspotl:
This PR introduces priorities when sending messages over to a client via RemoteConnection. The priority is currently assigned using a fixed mapping, in the future, this could potentially be extended to be user-specifiable. 

Relevant `ws-protocol` PR: https://github.com/luxonis/ws-protocol/pull/1


src/remote_connection/RemoteConnectionImpl.cpp
```diff
@@ -109,17 +109,24 @@ bool RemoteConnectionImpl::initWebsocketServer(const std::string& address, uint1
                 logger::debug(msgStr);
                 break;
             case foxglove::WebSocketLogLevel::Info:
-                // TODO(Morato) - revisit and potentially pass the levels correctly
-                // Right now, the happy path produces errors when the client disconnects
+                logger::info(msgStr);
+                break;
             case foxglove::WebSocketLogLevel::Warn:
```
moratom, line 114: Was this resolved? I recall quite the happy path produced error logs with this?

lnotspotl: @moratom I think those warning messages informing the user that the send buffer has reached its size limit are very useful, however, seeing those error and critical logs at the bottom might make the user think that there is some issue on the depthai side which is not the case. I've reverted the relevant changes. This information is still available is the user sets the log level to info.

image

src/remote_connection/RemoteConnectionImpl.cpp
```diff
@@ -119,7 +119,10 @@ bool RemoteConnectionImpl::initWebsocketServer(const std::string& address, uint1
         }
     };
     foxglove::ServerOptions serverOptions;
-    serverOptions.sendBufferLimitBytes = 100 * 1024 * 1024;  // 100 MB
+    constexpr size_t MAX_SEND_BUFFER_LIMIT_BYTES = 100 * 1024 * 1024;  // 100 MB
+    serverOptions.sendBufferPriorityLimitBytes = {{0, static_cast<size_t>(0.96 * MAX_SEND_BUFFER_LIMIT_BYTES)}, {1, MAX_SEND_BUFFER_LIMIT_BYTES}};
```
moratom: Let's remove this since it's not used and then makes it confusing.

PR #1502, opened 2025-10-22
Update DetectionParser to use new ImgDetections message and parse all YOLO families

aljazkonec1:
## Purpose

This PR ports depthai-nodes [YoloExtendedParser](https://github.com/luxonis/depthai-nodes/blob/main/depthai_nodes/node/parsers/yolo.py) into depthai library. DetectionParser can now process all YOLO families with the additional Segmentation and Keypoints output heads and return them as an updated `dai::ImgDetections` message.

## Specification

In depthai-core this port changes the following:
- Update to `dai::ImgDetections` to include new optional `Keypoints` and `SegmentationMask` attributes 
- Update to `dai::ImgDetections` to include `dai::RotatedRect` attribute and new getters and setters. When using the getters and setters, values between the legacy `xmin` `ymin` `xmax` `ymax` attributes and `RotatedRect` are synced. 
- New **Keypoint** common data structure: It is defined as a collection of `Point3f coordinates`, `float confidence` and `int label`.
- New **KeypointsList** common data structure: It stores a vector of `dai::Keypoint` objects, a vector of possible edge idices between Keypoints and provides utility functions for faster and easier conversions between `dai::Keypoints` types and other common types like `dai::Point3f` / `dai::Point2f`.
- New **Keypoints** message type: serialization of `dai::KeypointsList`.
- Pybindings for all new additions.
- Unit test for `DetectionParser`: Runs 8 models and compares results to stored ground truth bounding boxes, keypoints and segmentation masks
- Unit test for `dai::ImgDetections`

Potential risks include already built pipelines that rely on `dai::ImgDetections`.

## Testing & Validation

Tested on RVC4 device across all public YOLO models on our zoo.


moratom:
Thanks @aljazkonec1 ! 

Let's add an example for the new parser capabilities on RVC4 too.

bindings/js/bindings.cpp
```diff
@@ -77,19 +77,46 @@ EMSCRIPTEN_BINDINGS(depthai_js) {
         .value("TrackedFeatures", dai::DatatypeEnum::TrackedFeatures)
         .value("BenchmarkReport", dai::DatatypeEnum::BenchmarkReport)
         .value("MessageGroup", dai::DatatypeEnum::MessageGroup)
-        .value("ImageFiltersConfig", dai::DatatypeEnum::ImageFiltersConfig);
+        .value("ImageFiltersConfig", dai::DatatypeEnum::ImageFiltersConfig)
```
lnotspotl: As far as I know, we don't have an official support for javascript bindings and I am not sure we ever will? @moratom 

Maybe this is redundant if we decide to remove the `js` folder completely in the future.

bindings/python/src/pipeline/datatype/ImgDetectionsBindings.cpp
```diff
@@ -35,13 +37,50 @@ void bind_imgdetections(pybind11::module& m, void* pCallstack) {
...
+        .def("setKeypoints", py::overload_cast<const std::vector<Point3f>>(&ImgDetection::setKeypoints), py::arg("keypoints"))
+        .def("setKeypoints", py::overload_cast<const std::vector<Point2f>>(&ImgDetection::setKeypoints), py::arg("keypoints"))
+        .def("getKeypoints", &ImgDetection::getKeypoints, DOC(dai, ImgDetection, getKeypoints))
+        .def("setEdges", &ImgDetection::setEdges, py::arg("edges"))
+        .def("getEdges", &ImgDetection::getEdges, DOC(dai, ImgDetection, getEdges))
+        .def("centerX", &dai::ImgDetection::getCenterX)
+        .def("centerY", &dai::ImgDetection::getCenterY)
+        .def("width", &dai::ImgDetection::getWidth)
+        .def("height", &dai::ImgDetection::getHeight)
+        .def("angle", &dai::ImgDetection::getAngle);
```
lnotspotl: I guess this is a matter of preference but for consistency, should we add the "get" prefix to differentiate between getters and setters?

bindings/python/src/pipeline/datatype/KeypointsBindings.cpp
```diff
@@ -0,0 +1,64 @@
...
+#include "depthai/common/Point3f.hpp"
+#include "depthai/pipeline/datatype/Keypoints.hpp"
+
+// pybind
+#include <pybind11/chrono.h>
+#include <pybind11/detail/common.h>
+#include <pybind11/numpy.h>
+#include <pybind11/pybind11.h>
+
+// #include "spdlog/spdlog.h"
```
lnotspotl:
```suggestion
```

lnotspotl: redundant?

bindings/python/src/pipeline/datatype/KeypointsBindings.cpp
```diff
@@ -0,0 +1,64 @@
...
+        .def("setKeypoints", py::overload_cast<const std::vector<Point3f>&>(&Keypoints::setKeypoints), py::arg("keypoints"))
+        .def("setKeypoints", py::overload_cast<const std::vector<Point2f>&>(&Keypoints::setKeypoints), py::arg("keypoints"))
+        .def("setKeypoints",
+             py::overload_cast<const std::vector<Keypoint>&, const std::vector<std::array<uint32_t, 2>>&>(&Keypoints::setKeypoints),
+             py::arg("keypoints"),
+             py::arg("edges"))
+        .def("setEdges", &dai::Keypoints::setEdges, py::arg("edges"))
+        .def("getKeypoints", &Keypoints::getKeypoints, DOC(dai, Keypoints, getKeypoints))
+        .def(
+            "getEdges", [](Keypoints& msg) { return msg.getEdges(); }, DOC(dai, Keypoints, getEdges))
```
lnotspotl: Is a lambda function necessary here?

include/depthai/common/Keypoint.hpp
```diff
@@ -0,0 +1,35 @@
...
+// std
+#include <cstdint>
+
+// project
+#include <stdexcept>
+#include <string>
+
+#include "depthai/common/Point2f.hpp"
+#include "depthai/common/Point3f.hpp"
+#include "depthai/utility/Serialization.hpp"
```
lnotspotl:
```suggestion
// std
#include <cstdint>
#include <stdexcept>

// project
#include "depthai/common/Point2f.hpp"
#include "depthai/common/Point3f.hpp"
#include "depthai/utility/Serialization.hpp"
```

the `<string>` header seems not to be used here

include/depthai/common/Keypoint.hpp
```diff
@@ -0,0 +1,35 @@
...
+
+namespace dai {
+
+struct Keypoint {
+    Point3f coordinates{};
+    float confidence = 0.f;
+    uint32_t label = 0;
+
+    Keypoint() = default;
+    explicit Keypoint(Point3f coordinates, float conf = 0.f, uint32_t label = 0) : coordinates(coordinates), confidence(conf), label(label) {
```
lnotspotl: Is the `explicit` keyword necessary here? And if so, should we also add it to the other constructor bellow with `Point2f` as its first argument?

include/depthai/common/KeypointsList.hpp
```diff
@@ -0,0 +1,149 @@
...
+     * @param keypoints list of Keypoint objects and edges to set.
+     * @note This will clear any existing keypoints and edges.
+     */
+    void setKeypoints(const std::vector<Keypoint> keypoints, const std::vector<Edge> edges) {
+        this->edges.clear();
+        this->keypoints.clear();
+        this->keypoints = std::move(keypoints);
+        this->edges = std::move(edges);
+        validateEdges();
+    }
```
lnotspotl: Are the `clear` method calls necessary here? Also, since `keypoints` and `edges` are marked as `consts`, will this elicit the expected behavior, i.e. a move will happen instead of a copy?

include/depthai/common/KeypointsList.hpp
```diff
@@ -0,0 +1,149 @@
...
+     * @return Vector of Point3f coordinates.
+     */
+    std::vector<Point3f> getCoordinates3f() const {
+        std::vector<Point3f> coordinates;
+
+        for(const auto& kp : keypoints) {
+            coordinates.emplace_back(kp.coordinates);
+        }
+        return coordinates;
+    }
```
lnotspotl: Capacity could be specified here.

include/depthai/common/KeypointsList.hpp
```diff
@@ -0,0 +1,149 @@
...
+     * @return Vector of Point2f coordinates.
+     */
+    std::vector<Point2f> getCoordinates2f() const {
+        std::vector<Point2f> coordinates;
+
+        for(const auto& kp : keypoints) {
+            coordinates.emplace_back(Point2f(kp.coordinates.x, kp.coordinates.y));
+        }
+        return coordinates;
+    }
```
lnotspotl: Capacity could be specified here to prevent the vector's internal memory re-allocations

include/depthai/pipeline/datatype/ImgDetections.hpp
```diff
@@ -1,30 +1,155 @@
...
+
+#ifdef DEPTHAI_HAVE_OPENCV_SUPPORT
+    #include <opencv2/core/mat.hpp>
+    #include <opencv2/opencv.hpp>
+#endif
+
 namespace dai {
 
 struct ImgDetection {
+   public:
```
lnotspotl: probably not necessary, struct's attributes are public by default.

include/depthai/pipeline/datatype/ImgDetections.hpp
```diff
@@ -38,6 +163,94 @@ class ImgDetections : public Buffer, public ProtoSerializable {
...
+     * Retrieves data by instance index
+     *
+     */
+    cv::Mat getCvSegmentationMaskByIndex(uint8_t index, cv::MatAllocator* allocator = nullptr);
+
+    /**
+     * Retrieves data by semantic class
+     *
+     */
+    cv::Mat getCvSegmentationMaskByClass(uint8_t semantic_class, cv::MatAllocator* allocator = nullptr);
```
lnotspotl: Could we add a note on why we accept a pointer to an opencv allocator? I guess this is for efficiency reasons and in our case we use it when we create python bindings to remove any unnecessary copy operations?

include/depthai/pipeline/datatype/Keypoints.hpp
```diff
@@ -0,0 +1,97 @@
...
+#include "depthai/common/optional.hpp"
+#include "depthai/pipeline/datatype/Buffer.hpp"
+#include "depthai/utility/ProtoSerializable.hpp"
+
+namespace dai {
+
+/**
+ * Keypoints message. Carries a normalized keypoint 3D position, confidence score and label index.
+ */
+class Keypoints : public Buffer, public ProtoSerializable {
```
lnotspotl: `Keypoints` inherits from `ProtoSerializable` but there is not correspondings proto message definition. Is this intended? Calling `serializeProto` or `serializeSchema` results in a runtime error so there is no reason to have those methods anyways?

include/depthai/pipeline/datatype/Keypoints.hpp
```diff
@@ -0,0 +1,97 @@
...
+    std::vector<std::uint8_t> serializeProto(bool = false) const override;
+
+    /**
+     * Serialize schema to proto buffer
+     *
+     * @returns serialized schema
+     */
+    ProtoSerializable::SchemaPair serializeSchema() const override;
+
+#endif
```
lnotspotl: I guess when `DEPTHAI_ENABLE_PROTOBUF` is not defined (someone compiles depthai without protobuf support), there won't be any out of line virtual function definitions and `clang` will generate weak vtables - could we move one of the other virtual methods, e.g. the destructor, to the `Keypoints.cpp` file such that this does not happen? `depthai-core` won't compile without this change.

Relevant PR: https://github.com/luxonis/depthai-core/pull/1452

src/pipeline/datatype/Keypoints.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+#ifdef DEPTHAI_ENABLE_PROTOBUF
+ProtoSerializable::SchemaPair Keypoints::serializeSchema() const {
+    return utility::serializeSchema(utility::getProtoMessage(this));
+}
+
+std::vector<std::uint8_t> Keypoints::serializeProto(bool) const {
+    return utility::serializeProto(utility::getProtoMessage(this));
+}
+#endif
```
lnotspotl: `getProtoMessage` is not yet defined for `Keypoints`
   - technically it is, but it throws an error :smile:

src/pipeline/node/DetectionParser.cpp
```diff
@@ -209,6 +271,23 @@ void DetectionParser::setIouThreshold(float thresh) {
     properties.parser.iouThreshold = thresh;
 }
 
+void DetectionParser::setSubtype(const std::string& subtype) {  // TODO add bindings
```
lnotspotl: bindings still missing?

bindings/python/src/pipeline/datatype/ImgDetectionsBindings.cpp
```diff
@@ -82,8 +121,28 @@ void bind_imgdetections(pybind11::module& m, void* pCallstack) {
...
+        .def("setMask", &ImgDetections::setMask, py::arg("mask"), py::arg("width"), py::arg("height"), DOC(dai, ImgDetections, setMask))
+        .def("getMaskData", &ImgDetections::getMaskData, DOC(dai, ImgDetections, getMaskData))
+        .def("getSegmentationMaskAsImgFrame", &ImgDetections::getSegmentationMaskAsImgFrame, DOC(dai, ImgDetections, getSegmentationMaskAsImgFrame))
+#ifdef DEPTHAI_HAVE_OPENCV_SUPPORT
+        .def(
+            "getSegmentationMask", [](ImgDetections& self) { return self.getSegmentationMask(false); }, DOC(dai, ImgDetections, getSegmentationMask))
+        .def("setSegmentationMask", &ImgDetections::setSegmentationMask, DOC(dai, ImgDetections, setSegmentationMask))
+        .def(
+            "getCvSegmentationMask",
+            [](ImgDetections& self) { return self.getCvSegmentationMask(&g_numpyAllocator); },
```
lnotspotl, line 151: Just out of curiosity, what does the numpy allocator do and why doesn't OpenCV's default allocator suffice?

bindings/python/src/pipeline/datatype/KeypointsBindings.cpp
```diff
@@ -0,0 +1,64 @@
...
+#include <pybind11/detail/common.h>
+#include <pybind11/numpy.h>
+#include <pybind11/pybind11.h>
+
+// #include "spdlog/spdlog.h"
+
+void bind_keypoints(pybind11::module& m, void* pCallstack) {
+    using namespace dai;
+
+    // py::class_<Keypoints, Py<Keypoints>, Buffer, std::shared_ptr<Keypoints>> keypoints(m, "Keypoints", DOC(dai, Keypoints));
```
lnotspotl:
```suggestion
```

include/depthai/common/KeypointsList.hpp
```diff
@@ -0,0 +1,149 @@
+#pragma once
+
+// std
+#include <array>
+#include <cstdint>
+#include <stdexcept>
+#include <string>
```
lnotspotl:
```suggestion
```
unused

include/depthai/common/KeypointsList.hpp
```diff
@@ -0,0 +1,149 @@
...
+        validateEdges();
+    }
+
+    /**
+     * Set the indices of the edges.
+     * @param edges Vector of edge indices.
+     */
+
+    void setEdges(const std::vector<Edge> edges) {
+        this->edges = edges;
```
lnotspotl: maybe, just like in `setKeypoints`, we can do move here

include/depthai/pipeline/datatype/Keypoints.hpp
```diff
@@ -0,0 +1,97 @@
...
+
+    void setKeypoints(const std::vector<Keypoint>& keypoints);
+
+    void setKeypoints(const std::vector<Point3f>& keypoints);
+
+    void setKeypoints(const std::vector<Point2f>& keypoints);
+
+    void setKeypoints(const std::vector<Keypoint>& keypoints, const std::vector<std::array<uint32_t, 2>>& edges);
+
+    void setEdges(const std::vector<std::array<uint32_t, 2>>& edges);
```
lnotspotl: We could use `std::vector<Edge>> &edges` here. The same goes for the `getEdges` method.

`Edge` is specified as an alias in `KeypointsList.hpp`

src/opencv/ImgDetections.cpp
```diff
@@ -0,0 +1,98 @@
+#include "depthai/pipeline/datatype/ImgDetections.hpp"
+
+#include <cmath>
```
lnotspotl:
```suggestion
```
unused

bindings/python/src/pipeline/datatype/ImgDetectionsBindings.cpp
aljazkonec1: I wanted the syntax to be as similar as possible to accessing the attribute as possible. But yes for consistency, I do agree that 'get' would be more appropriate.

bindings/python/src/pipeline/datatype/ImgDetectionsBindings.cpp
aljazkonec1: The same is done in `dai::ImgFrame::getCvFrame` and the allocator is used to control memory characteristics (pooling, alignment, ...). Here its set to numpys' allocator because return type of this function inpython is ndarray (same as dai.ImgFrame).

include/depthai/common/Keypoint.hpp
aljazkonec1: After a think, it would be better to have it a converting constructor.

aljazkonec1: I made all three explicit to prevent accidental conversions

include/depthai/common/KeypointsList.hpp
aljazkonec1: Thanks, looks like I missed it!

include/depthai/pipeline/datatype/Keypoints.hpp
aljazkonec1: Added proto message.

include/depthai/pipeline/datatype/Keypoints.hpp
aljazkonec1: Thanks! I made the message before the PR was merged so I wasn't aware of the change.

src/pipeline/datatype/Keypoints.cpp
aljazkonec1: Missed it, thanks.

include/depthai/common/DetectionParserOptions.hpp
```diff
@@ -16,19 +20,47 @@ struct DetectionParserOptions {
     DetectionNetworkType nnFamily;
     std::string subtype;
     float confidenceThreshold;
+    int inputWidth;
+    int inputHeight;
```
moratom: Why is this needed? So far we got the input from the ImgTransformations

include/depthai/common/DetectionParserOptions.hpp
```diff
@@ -16,19 +20,47 @@ struct DetectionParserOptions {
...
     /// YOLO specific network properties
+    YoloDecodingFamily decodingFamily = YoloDecodingFamily::TLBR;  // top left bottom right anchor free
+    bool decodeKeypoints = false;
+    bool decodeSegmentation = false;
+
     int classes;
     std::optional<std::vector<std::string>> classNames;
     int coordinates;
+    std::optional<int> nKeypoints;
+    std::vector<int> strides = {8, 16, 32};
```
moratom, line 33: Where are these strides used, are the values default for some model?

include/depthai/common/DetectionParserOptions.hpp
```diff
@@ -16,19 +20,47 @@ struct DetectionParserOptions {
...
     std::vector<float> anchors;
     std::map<std::string, std::vector<int>> anchorMasks;
+    std::vector<std::string> outputNames;
```
moratom: Maybe outputNamesToUse would be better?

examples/cpp/DetectionNetwork/RVC4/detection_and_segmentation.cpp
```diff
@@ -0,0 +1,168 @@
...
+    cameraNode->build();
+
+    // Create and configure detection network node
+    auto detectionNetwork = pipeline.create<dai::node::DetectionNetwork>();
+
+    dai::NNModelDescription modelDescription;
+    modelDescription.model = "luxonis/yolov8-instance-segmentation-large:coco-640x480";
+    detectionNetwork->build(cameraNode, modelDescription);
+    auto labelMap = detectionNetwork->getClasses();
+
```
moratom: Shouldn't be needed anymore, labels are part of ImgDetections (likely applies to other examples as well)

src/pipeline/datatype/StreamMessageParser.cpp
```diff
@@ -258,7 +259,6 @@ std::shared_ptr<ADatatype> StreamMessageParser::parseMessage(streamPacketDesc_t*
             break;
         case DatatypeEnum::ObjectTrackerConfig:
             return parseDatatype<ObjectTrackerConfig>(metadataStart, serializedObjectSize, data, fd);
-            break;
```
moratom, line 261: ?

include/depthai/pipeline/datatypes.hpp
```diff
@@ -14,6 +14,7 @@
 #include "datatype/IMUData.hpp"
 #include "datatype/ImageManipConfig.hpp"
 #include "datatype/ImgDetections.hpp"
+#include "datatype/ImgDetectionsT.hpp"
```
moratom, line 17: Is this intended to be used by user? If not we don't need this include here.

bindings/python/src/pipeline/CommonBindings.cpp
```diff
@@ -92,6 +99,45 @@ void CommonBindings::bind(pybind11::module& m, void* pCallstack) {
...
     ///////////////////////////////////////////////////////////////////////
 
+    keypoint.def(py::init<>())
+        .def(py::init<Point3f, float, uint32_t>(), py::arg("coordinates"), py::arg("confidence") = 0.f, py::arg("label") = 0)
+        .def(py::init<Point2f, float, uint32_t>(), py::arg("coordinates"), py::arg("confidence") = 0.f, py::arg("label") = 0)
+        .def(py::init<float, float, float, float, uint32_t>(), py::arg("x"), py::arg("y"), py::arg("z"), py::arg("confidence") = 0.f, py::arg("label") = 0)
+        .def_readwrite("imageCoordinates", &Keypoint::imageCoordinates)
+        .def_readwrite("confidence", &Keypoint::confidence)
+        .def_readwrite("label", &Keypoint::label);
+
```
moratom, line 115: DOCs missing

bindings/python/src/pipeline/CommonBindings.cpp
```diff
@@ -92,6 +99,45 @@ void CommonBindings::bind(pybind11::module& m, void* pCallstack) {
...
+    keypointsList.def(py::init<>())
+        .def(py::init<std::vector<Keypoint>, std::vector<Edge>>(), py::arg("keypoints"), py::arg("edges"))
+        .def(py::init<std::vector<Keypoint>>(), py::arg("keypoints"))
+        .def(
```
moratom: DOCs missing

include/depthai/pipeline/node/DetectionParser.hpp
```diff
@@ -126,39 +131,150 @@ class DetectionParser : public DeviceNodeCRTP<DeviceNode, DetectionParser, Detec
...
+
+    /**
+     * Enable/disable segmentation mask decoding.
+     */
+    void setDecodeSegmentation(bool decode);
+
+    /**
+     * Set number of keypoints to decode. Automatically enables keypoints decoding.
+     */
+    void setNKeypoints(int nKeypoints);
```
moratom: setNumKeypoints is more consistent with other API

src/pipeline/node/DetectionParser.cpp
```diff
@@ -187,6 +261,7 @@ float DetectionParser::getConfidenceThreshold() const {
 
 void DetectionParser::setNumClasses(const int numClasses) {
     properties.parser.classes = numClasses;
+    properties.parser.classNames = std::nullopt;
```
moratom: Why?

include/depthai/pipeline/datatype/ImgDetectionsT.hpp
```diff
@@ -0,0 +1,172 @@
...
+    }
+    size_t size() const noexcept {
+        return detections.size();
+    }
+    value_type& operator[](size_t i) {
+        return detections[i];
+    }
+    const value_type& operator[](size_t i) const {
+        return detections[i];
+    }
```
moratom: Do we need this?
I think one can just use `ImgDetections.detections`

include/depthai/pipeline/datatype/ImgDetectionsT.hpp
```diff
@@ -0,0 +1,172 @@
...
+     * The size of the vector must be equal to width * height.
+     */
+    void setMask(const std::vector<std::uint8_t>& mask, size_t width, size_t height);
+
+    /*
+     * Returns a copy of the segmentation mask data as a vector of bytes. If mask data is not set, returns std::nullopt.
+     */
+    std::optional<std::vector<std::uint8_t>> getMaskData() const;
+
+    std::optional<dai::ImgFrame> getSegmentationMaskAsImgFrame() const;
```
moratom:
```suggestion
    std::optional<dai::ImgFrame> getSegmentationMask() const;
```
and then
getCvSegmentationMask() for the one that returns a cv::Mat to be consistent with ImgFrame

include/depthai/pipeline/datatype/ImgDetectionsT.hpp
```diff
@@ -0,0 +1,172 @@
...
+    /*
+     * Returns the height of the segmentation mask.
+     */
+    std::size_t getSegmentationMaskHeight() const;
+
+    /*
+     * Sets the segmentation mask from a vector of bytes, along with width and height.
+     * The size of the vector must be equal to width * height.
+     */
+    void setMask(const std::vector<std::uint8_t>& mask, size_t width, size_t height);
```
moratom:
```suggestion
    void setSegmentationMask(const std::vector<std::uint8_t>& mask, size_t width, size_t height);
```

moratom: (and unify return type for all setters, can be &ImgDetectionsT)

include/depthai/pipeline/datatype/ImgDetectionsT.hpp
```diff
@@ -0,0 +1,172 @@
...
+     * @param frame Input cv::Mat frame from which to copy the data
+     */
+    ImgDetectionsT& setSegmentationMask(cv::Mat mask);
+
+    /**
+     * Retrieves data as cv::Mat with specified width, height and type. If mask data is not set, returns std::nullopt.
+     *
+     * @param copy If false only a reference to data is made, otherwise a copy
+     */
+    std::optional<cv::Mat> getSegmentationMask(bool copy = false);
```
moratom: I would remove that one - is there a functional difference between getSegmentationMask and getCvSegmentationMask?

src/pipeline/datatype/ImgDetectionsT.cpp
```diff
@@ -0,0 +1,218 @@
...
+template <class DetectionT>
+ImgDetectionsT<DetectionT>& ImgDetectionsT<DetectionT>::setTensorSegmentationMask(typename ImgDetectionsT<DetectionT>::XArray2D mask) {
+    data->setSize(mask.size());
+    std::vector<uint8_t> dataVec(mask.begin(), mask.end());
+    setData(dataVec);
+    this->segmentationMaskWidth = mask.shape()[1];
+    this->segmentationMaskHeight = mask.shape()[0];
+    return *this;
+}
+
```
moratom: Let's add tests for these (or we could remove too)

src/pipeline/datatype/ImgDetectionsT.cpp
```diff
@@ -0,0 +1,218 @@
...
+        return std::nullopt;
+    }
+    cv::Mat output;
+    if(allocator != nullptr) {
+        output.allocator = allocator;
+    }
+    (*mask).copyTo(output);
+    return output;
+}
+
```
moratom, line 121: Can be merged into one IMO

moratom (commented):
Thanks!

Left a few nitpicks, requests for tests, but generally LGTM!

include/depthai/pipeline/datatypes.hpp
aljazkonec1: I would leave it to allow creating custom ImgDetection classes

include/depthai/common/DetectionParserOptions.hpp
aljazkonec1: They are used for scaling outputs from the default three heads of yolo models. The default number of heads is three, corresponding to these strides.

PR #1504, opened 2025-10-28
Fix: Improve calculation of spatial coordinates from depth

aljazkonec1:
## Purpose

Calculating spatial coordinates now uses intrinsic matrix instead of raw FOV angles in SpatialLocationCalculatorNode on both RVC4 and RVC2. Added syntetic test for both.
## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable


moratom (approved):
Thanks @aljazkonec1, LGTM.

PR #1509, opened 2025-10-29
Fix: Add configurable rpc timeouts

aljazkonec1:
## Purpose

This PR adds two overload functions that allow the configuration of RPC timeouts with the use of a scoped timeout guard. If no timeout is specified the previous default `RPC_READ_TIMEOUT` is used. Set timeout to 0 to allow infinate wait per RPC call.

## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

### DIFF
--- a/src/device/DeviceBase.cpp
+++ b/src/device/DeviceBase.cpp
@@ -52,9 +52,11 @@
 #include "utility/spdlog-fmt.hpp"
 
 namespace {
-thread_local std::optional<std::chrono::milliseconds> tlRpcTimeout;
 
 struct ScopedRpcTimeout {
+   public:
+    static thread_local std::optional<std::chrono::milliseconds> tlRpcTimeout;
+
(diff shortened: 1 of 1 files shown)


src/device/DeviceBase.cpp
```diff
@@ -51,6 +51,31 @@
...
+    ScopedRpcTimeout& operator=(const ScopedRpcTimeout&) = delete;
+
+    std::optional<std::chrono::milliseconds> prev;
+};
+
+thread_local std::optional<std::chrono::milliseconds> ScopedRpcTimeout::tlRpcTimeout;
+
+std::optional<std::chrono::milliseconds> currentRpcTimeout() {
+    return ScopedRpcTimeout::tlRpcTimeout;
+}
```
moratom, line 77: Why is this outside of the struct?

PR #1512, opened 2025-11-03
Add missing bindings to DetectionNetwork

aljazkonec1:
## Purpose

[User stated](https://discuss.luxonis.com/d/6350-old-blob-on-depthai-v3) missing bindings when moving from v2 to v3. This PR addresses those bindings.

## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

### DIFF
--- a/bindings/python/src/pipeline/node/DetectionNetworkBindings.cpp
+++ b/bindings/python/src/pipeline/node/DetectionNetworkBindings.cpp
@@ -87,12 +87,6 @@ void bind_detectionnetwork(pybind11::module& m, void* pCallstack) {
              DETECTION_NETWORK_PYARGS)
         // Copied from NN node
         .def("setBlobPath", &DetectionNetwork::setBlobPath, py::arg("path"), DOC(dai, node, DetectionNetwork, setBlobPath))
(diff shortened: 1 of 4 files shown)


bindings/python/src/pipeline/node/DetectionNetworkBindings.cpp
```diff
@@ -148,8 +154,18 @@ void bind_detectionnetwork(pybind11::module& m, void* pCallstack) {
             [](const DetectionNetwork& n) { return &n.neuralNetwork->passthrough; },
             py::return_value_policy::reference_internal,
             DOC(dai, node, NeuralNetwork, passthrough))
+        .def_property_readonly(
+            "detectionParser",
+            [](DetectionNetwork& n) { return &(*n.detectionParser); },
+            py::return_value_policy::reference_internal,
+            DOC(dai, node, DetectionNetwork, detectionParser))
```
moratom, line 155: I think we should only keep this bit and expose the same also in `SpatialDetectionNetwork`.

Otherwise looks good to me, thanks!

PR #1515, opened 2025-11-06
Feature/video encoder tests update

aljazdu:
## Purpose
Vastly expanded test functionality for video encoder node

## Specification

Changes pertain to video encoder node and to video record node

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

Testing was done using both RVC2 and RVC4 cameras


tests/src/ondevice_tests/encoded_frame_test.cpp
moratom, line 1: Does RVC4 run this in CI? Needs to be skipped.

aljazdu: Added a check for RVC4, skipping if it is used

PR #1521, opened 2025-11-13
Update DCL API in the DynamicCalibration node

JakubFara:
## Purpose
Update DCL api in the DynamicCalibration node

## Specification
- This will not change the API of the DynamicCalibration node itself

- need change of the DCL library to **f124a38** @ `v0.3.0rc2 origin/rel_0.3.0 `


moratom (approved):
Thanks, tested on MacOS 26 and works as expected (+ DynamicCalibration has the same target OS as DepthAI now, which is nice to see)

PR #1529, opened 2025-11-18
Fix/remote connection service test

aljazdu:
## Purpose

Added remote connection services tests

## Specification

Added remote connection services tests to bindings/tests. setup.py will now have install_requires for websockets as well

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

### DIFF
--- a/bindings/python/tests/remote_connection_service_test.py
+++ b/bindings/python/tests/remote_connection_service_test.py
@@ -24,7 +24,7 @@ async def call_service(ws, service_name, service_id):
     return None
 
 
-async def test_available_services():
+async def get_available_services():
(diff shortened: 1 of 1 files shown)


bindings/python/setup.py
```diff
@@ -311,6 +311,7 @@ def build_extension(self, ext):
     python_requires='>=3.7',
     install_requires=[
         "numpy<3.0.0",
+        "websockets>11.0.0"
     ],
     entry_points={
```
moratom, line 315: We should only add the dependancy for testing, the package itself does not depend on it.

aljazdu: I have moved this away from setup.py 
The pytest tests now pass successfully

PR #1530, opened 2025-11-18
Vio and VSLAM updates

Serafadam:
## Purpose

Hiding the dependencies for VIO and VSLAM so that external lining doesn't require them to be linked as well
## Specification

None / not applicable
The changes are ported from ROS. Linking type has been changed to private, PIMPL pattern has been used to hide headers for VIO and VSLAM, also methods that were used for RTABMap in TransformData and CalibrationHandler have been moved to other file. Additionally, new datatype has been introduces `MapData` which contains also minX and minY fields used to correctly position maps in ROS.

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable


bindings/python/src/pipeline/node/BasaltVIOBindings.cpp
```diff
@@ -36,73 +36,79 @@ void bind_basaltnode(pybind11::module& m, void* pCallstack) {
         .def_readonly("passthrough", &BasaltVIO::passthrough, DOC(dai, node, BasaltVIO, passthrough))
         .def("setImuUpdateRate", &BasaltVIO::setImuUpdateRate, py::arg("rate"), DOC(dai, node, BasaltVIO, setImuUpdateRate))
         .def("setConfigPath", &BasaltVIO::setConfigPath, py::arg("path"), DOC(dai, node, BasaltVIO, setConfigPath))
-        .def_readwrite("vioConfig", &BasaltVIO::vioConfig)
+        .def("setVioConfig", &BasaltVIO::setConfig, py::arg("config"), DOC(dai, node, BasaltVIO, setConfig))
```
lnotspotl: can we use the same name here, either setConfig or setVioConfig for both the C++ method and the pybind11 wrapper?

examples/cpp/RVC2/VSLAM/rerun_node.hpp
```diff
@@ -96,9 +97,10 @@ class RerunNode : public dai::NodeCRTP<dai::node::ThreadedHostNode, RerunNode> {
                 }
 #endif
                 if(mapData != nullptr) {
-                    auto image = mapData->getCvFrame();
+                    auto map = mapData->map;
```
lnotspotl: Doesn't this produce a copy? I guess it's a relatively cheap copy though, but I still think the intent here was to use a reference.
```suggestion
                    auto &map = mapData->map;
```

include/depthai/basalt/BasaltVIO.hpp
```diff
@@ -22,6 +17,89 @@ namespace node {
...
+        bool vio_sqrt_marg;
+
+        int vio_max_states;
+        int vio_max_kfs;
+        int vio_min_frames_after_kf;
+        float vio_new_kf_keypoints_thresh;
+        bool vio_debug;
+        bool vio_extended_logging;
+
+        //  double vio_outlier_threshold;
```
lnotspotl: are these needed?

include/depthai/device/CalibrationHandler.hpp
```diff
@@ -11,13 +11,6 @@
 #include "depthai/common/Point2f.hpp"
 #include "depthai/common/Size2f.hpp"
 
-#ifdef DEPTHAI_HAVE_RTABMAP_SUPPORT
-    #pragma push_macro("_res")
```
lnotspotl, line 16: I am glad to see this gone :)

include/depthai/pipeline/datatype/MapData.hpp
```diff
@@ -0,0 +1,34 @@
...
+   public:
+    /**
+     * Construct RGBD message.
+     */
+    MapData() = default;
+
+    virtual ~MapData() = default;
+
+    dai::ImgFrame map;
+    float minX;
```
lnotspotl: What's the point of having these floats? It does not seem they are used anywhere?

include/depthai/pipeline/datatype/MapData.hpp
```diff
@@ -0,0 +1,34 @@
...
+#include "depthai/common/ADatatypeSharedPtrSerialization.hpp"
+#include "depthai/pipeline/datatype/ADatatype.hpp"
+#include "depthai/pipeline/datatype/Buffer.hpp"
+#include "depthai/pipeline/datatype/ImgFrame.hpp"
+#include "depthai/utility/Serialization.hpp"
+
+namespace dai {
+
+/**
+ * RGBD message. Carries RGB and Depth frames.
```
lnotspotl: I don't fully understand this comment, what is MapData, how is it different from plain ImgFrame?

src/rtabmap/RTABMapConversions.hpp
```diff
@@ -0,0 +1,25 @@
...
+#include "depthai/pipeline/datatype/TransformData.hpp"
+
+namespace rtabmap {
+class Transform;
+class StereoCameraModel;
+}  // namespace rtabmap
+
+namespace dai {
+
+std::shared_ptr<TransformData> rtabmapToTransformData(rtabmap::Transform transformRTABMap);
```
lnotspotl, line 14: We still use rtabmap transforms here, is it guaranteed that it is declared? Which of the above included headers ensures that? 
(Looking at this from a perspective of a user who installs depthai and then wants to use it)

src/basalt/BasaltVIO.cpp
```diff
@@ -137,33 +156,103 @@ void BasaltVIO::stop() {
...
+void BasaltVIO::setGyroNoiseStd(const std::vector<double>& gyroNoiseStd) {
+    if(gyroNoiseStd.size() != 3) {
+        throw std::invalid_argument("Gyroscope noise vector must have 3 elements.");
+    }
+    this->gyroNoiseStd = gyroNoiseStd;
+}
+
+void BasaltVIO::setGyroBias(const std::vector<double>& gyroBias) {
+    if(gyroBias.size() != 12) {
+        throw std::invalid_argument("Gyroscope bias vector must have 9 elements.");
```
lnotspotl: The error's message does not match the check above (12 vs 9).

src/rtabmap/RTABMapSLAM.cpp
```diff
@@ -40,22 +148,28 @@ RTABMapSLAM::~RTABMapSLAM() {
             databasePath = "/tmp/rtabmap.db";
         }
         logger->info("Saving database at {}", databasePath);
-        rtabmap.close(true, databasePath);
+        pimplRtabmap->rtabmap.close(true, databasePath);
     }
 }
 void RTABMapSLAM::setParams(const rtabmap::ParametersMap& params) {
```
lnotspotl: the argument type should be std::map<std::string, std::string> &

CMakeLists.txt
```diff
@@ -233,15 +233,11 @@ set(THIRDPARTY_RTABMAP_LIBRARIES "rtabmap::utilite" CACHE STRING "Optional libra
 set(TARGET_RTABMAP_NAME ${PROJECT_NAME}-rtabmap)
 set(TARGET_RTABMAP_ALIAS rtabmap)
 if(DEPTHAI_RTABMAP_SUPPORT)
-    if(${RTABMap_FOUND} AND DEPTHAI_HAVE_PCL_SUPPORT)
+    if(${RTABMap_FOUND})
         set(DEPTHAI_HAVE_RTABMAP_SUPPORT ON)
         message(STATUC "RTABMAP found. RTABMAP Support enabled")
```
lnotspotl:
```suggestion
        message(STATUS "RTABMAP found. RTABMAP Support enabled")
```

src/basalt/BasaltVIO.cpp
```diff
@@ -137,33 +156,103 @@ void BasaltVIO::stop() {
...
+    }
+    this->accelBias = accelBias;
+}
+
+void BasaltVIO::setAccelNoiseStd(const std::vector<double>& accelNoiseStd) {
+    if(accelNoiseStd.size() != 3) {
+        throw std::invalid_argument("Accelerometer noise vector must have 3 elements.");
+    }
+    this->accelNoiseStd = accelNoiseStd;
+    ;
```
lnotspotl:
```suggestion
```

examples/cpp/RVC2/VSLAM/basalt_vio_rtabmap_slam.cpp
```diff
@@ -17,10 +17,10 @@ int main() {
     auto imu = pipeline.create<dai::node::IMU>();
     auto odom = pipeline.create<dai::node::BasaltVIO>();
     auto slam = pipeline.create<dai::node::RTABMapSLAM>();
-    auto params = rtabmap::ParametersMap();
-    params.insert(rtabmap::ParametersPair(rtabmap::Parameters::kRGBDCreateOccupancyGrid(), "true"));
-    params.insert(rtabmap::ParametersPair(rtabmap::Parameters::kGrid3D(), "true"));
-    params.insert(rtabmap::ParametersPair(rtabmap::Parameters::kRtabmapSaveWMState(), "true"));
+    std::map<std::string, std::string> params{};
+    params.insert(std::make_pair<std::string, std::string>("RGBD/CreateOccupancyGrid", "true"));
```
lnotspotl, line 21: What options can a user set, I think this should be documented (or at the very least, user should be pointed to rtabmap's documentation (maybe in the header, somewhere close to the declaration of setParams))

compile_commands.json
```diff
@@ -0,0 +1 @@
+/ws/src/depthai-core/build/compile_commands.json
```
lnotspotl: Commited by accident?

include/depthai/pipeline/datatype/MapData.hpp
Serafadam: These are used in ROS, when user wants to place the map in 3D space

src/rtabmap/RTABMapConversions.hpp
Serafadam: It's forward-declared, the include is in CPP file so that user doesn't have to worry about linking

include/depthai/pipeline/datatype/MapData.hpp
```diff
@@ -0,0 +1,31 @@
...
+   public:
+    MapData() = default;
+
+    virtual ~MapData() = default;
+
+    dai::ImgFrame map;
+    float minX;
+    float minY;
+
+    void serialize(std::vector<std::uint8_t>& metadata, DatatypeEnum& datatype) const override {
```
lnotspotl: We should also move serialize implementation to `MapData.cpp`, otherwise we're dealing with weak symbols and it won't compile on MacOS (not sure whether this is expected to work there). 

https://github.com/luxonis/depthai-core/pull/1452

Also, I'd explicitly initialize minX, minY (set it to zero)

src/rtabmap/RTABMapConversions.hpp
```diff
@@ -0,0 +1,24 @@
+#include <memory>
+
+#include "common/CameraBoardSocket.hpp"
+#include "depthai/device/CalibrationHandler.hpp"
+#include "depthai/pipeline/datatype/TransformData.hpp"
+
+namespace rtabmap {
+class Transform;
+class StereoCameraModel;
```
lnotspotl, line 9: Do we maybe want to move this entire header to src/rtabmap so that it signifies it is an internal header and when headers are installed via `cmake --build build --target install`, this header is not there at all

PR #1531, opened 2025-11-18
Add NeuralDepth node

moratom:
## Purpose
Add NeuralDepth node on RVC4.
4 models suppported (FPS tests were done end to end, not just model only):
* LARGE -> 768x480 @ 10.8 FPS
* MEDIUM -> 576x360 @ 25.5 FPS
* SMALL -> 480x300 @ 42.5 FPS
* NANO -> 384x240 @ 59.7 FPS

The node is built out of multiple nodes:
[Sync]->[Demux]->[Rectification]->[NeuralNetwork]->[PostProcessing]
allowing a high throughput and reusing a lot of code.

Requires Luxonis OS 1.20.5 or newer, no other new depenanices.

## Deployment Plan

None / not applicable

## Testing & Validation

Added a test to guard agains both performance and accuracy regressions.
Performance is measured with a live stream, the accuracy is checking the median depth and the fillrate remain in a good range.

Automated tests can be extended in the future with making sure confidence filter, edge filter etc are all working correctly.
For now this was tested manually.


include/depthai/pipeline/datatype/NeuralDepthConfig.hpp
```diff
@@ -0,0 +1,95 @@
...
+     * Construct NeuralDepthConfig message.
+     */
+    NeuralDepthConfig() = default;
+    virtual ~NeuralDepthConfig();
+
+    struct AlgorithmControl {
+        /**
+         * Measurement unit for depth data
+         */
+        enum class DepthUnit : int32_t { METER, CENTIMETER, MILLIMETER, INCH, FOOT, CUSTOM };
```
asahtik: Might be better to have just one definition of DepthUnit to prevent inconsistency with StereoDepth if we ever change anything

PR #1542, opened 2025-11-24
Calibration Handler test improvements

MaticTonin:
…etRotation, getStereoRectificationMatrices

## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

### DIFF
--- a/tests/src/onhost_tests/calibration_handler_test.cpp
+++ b/tests/src/onhost_tests/calibration_handler_test.cpp
@@ -69,6 +69,146 @@ static CalibrationHandler loadHandler() {
     return CalibrationHandler::fromJson(calibJson);
 }
 
+TEST_CASE("Invalid camera ID throws", "[getCameraIntrinsics]"){
+  auto handler = loadHandler();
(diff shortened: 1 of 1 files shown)


tests/src/onhost_tests/calibration_handler_test.cpp
```diff
@@ -66,37 +66,177 @@ static CalibrationHandler loadHandler() {
             {"specHfovDeg", 0.0},
             {"width", 1920}}}}}};
 
-    return CalibrationHandler::fromJson(calibJson);
+    return calibJson;
```
asahtik: From what I can tell `loadCalibration` is always used with `CalibrationHandler::fromJson`. How come this function was changed to return a json instead of the CH directly?

MaticTonin: Hi. Reason for it is that I didnt wanna tackle the issue of maybe having all transformation applied to the same calibrationHandler, but that seems like its not the case, will change if required

MaticTonin: Reverted in commit

PR #1543, opened 2025-11-24
Adding simple on host inverse transformations tests

MaticTonin:
## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

### DIFF
--- a/src/pipeline/datatype/ImgTransformations.cpp
+++ b/src/pipeline/datatype/ImgTransformations.cpp
@@ -66,17 +66,22 @@ inline bool mateq(const std::array<std::array<float, 3>, 3>& A, const std::array
     return true;
 }
 
-std::array<std::array<float, 3>, 3> getMatrixInverse(const std::array<std::array<float, 3>, 3>& matrix) {
+std::array<std::array<float, 3>, 3> getMatrixInverse(const std::array<std::array<float, 3>, 3>& matrix_float) {
+    // Step 1: Convert to double
+    std::array<std::array<double, 3>, 3> matrix;
+    for(int i = 0; i < 3; ++i)
+        for(int j = 0; j < 3; ++j) matrix[i][j] = static_cast<double>(matrix_float[i][j]);
+
     std::array<std::array<float, 3>, 3> inv;
-    float det = matrix[0][0] * (matrix[1][1] * matrix[2][2] - matrix[1][2] * matrix[2][1])
-                - matrix[0][1] * (matrix[1][0] * matrix[2][2] - matrix[1][2] * matrix[2][0])
-                + matrix[0][2] * (matrix[1][0] * matrix[2][1] - matrix[1][1] * matrix[2][0]);
+    double det = matrix[0][0] * (matrix[1][1] * matrix[2][2] - matrix[1][2] * matrix[2][1])
+                 - matrix[0][1] * (matrix[1][0] * matrix[2][2] - matrix[1][2] * matrix[2][0])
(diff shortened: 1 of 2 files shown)


tests/src/onhost_tests/image_transformations_test.cpp
```diff
@@ -230,3 +393,128 @@ TEST_CASE("interFrameRemap") {
...
 }
+// -----------------------------------------------------------------------------
+// ImgTransformation matrix inverse consistency (ImgFrame)
+// Purpose:
+//   Ensures that the forward matrix (M) and its stored inverse (Minv)
+//   multiply to identity. Same for intrinsics (K * Kinv).
+//   Validates correctness of ImgTransformation’s internal math.
+// -----------------------------------------------------------------------------
+TEST_CASE("ImgTransformation matrix inverse consistency (ImgFrame)") {
+    dai::Pipeline pipeline;
```
asahtik: This file is meant to run on host only. There is no guarantee that any device is connected to the runner. Tests that use devices should be moved to `ondevice_tests/img_transformation_test.cpp`

tests/src/onhost_tests/image_transformations_test.cpp
```diff
@@ -230,3 +393,128 @@ TEST_CASE("interFrameRemap") {
...
+    auto cam = pipeline.create<dai::node::Camera>()->build();
+    auto camOut = cam->requestOutput({1300, 200}, dai::ImgFrame::Type::NV12);
```
asahtik: Is there a specific reason for this resolution? This test fails on my RVC2 device (wide) because of it.

asahtik (changes requested):
Thanks. Always nice to have more tests, especially when it's not me adding them :smile:. There's a test in `onhost_tests` using a device which should be moved to `ondevice_tests`. The same test also fails on OAK-D W.

HIL should also be run before merging (add `testable` label to PR).

Other than that LGTM.

tests/src/onhost_tests/image_transformations_test.cpp
MaticTonin: No specific reason, tried to be as generic as possible, we can go to native resolution.

tests/src/onhost_tests/image_transformations_test.cpp
MaticTonin: Thanks for comment, will then move this test as on device!

tests/src/onhost_tests/image_transformations_test.cpp
MaticTonin: Added in https://github.com/luxonis/depthai-core/pull/1543/commits/b1b9d0fbf8cd38da79be4d34462051ababf096f6

tests/src/onhost_tests/image_transformations_test.cpp
MaticTonin: Added in https://github.com/luxonis/depthai-core/pull/1543/commits/b1b9d0fbf8cd38da79be4d34462051ababf096f6

tests/src/onhost_tests/image_transformations_test.cpp
MaticTonin: https://github.com/luxonis/depthai-core/pull/1543/commits/a98c7833c2c6369f0e07b2670565d1ef20bb79ce

tests/src/onhost_tests/image_transformations_test.cpp
MaticTonin: https://github.com/luxonis/depthai-core/pull/1543/commits/a98c7833c2c6369f0e07b2670565d1ef20bb79ce

PR #1563, opened 2025-12-02
Remote connection visualizer msg

aljazdu:
## Purpose
The message output of connecting to the visualizer always showed: "open http://localhost"
This was not the case when using devices in standalone mode.

## Specification
Changing the message to now use the device's ip. Also the examples now use http port of 8082 by default.
The video encoder example is also updated with the specified local host address in order to work!

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation
This was tested by running the devices in peripheral mode as well as in standalone using oakctl app.

### DIFF
--- a/src/remote_connection/RemoteConnectionImpl.cpp
+++ b/src/remote_connection/RemoteConnectionImpl.cpp
@@ -336,11 +336,13 @@ bool RemoteConnectionImpl::initHttpServer(const std::string& address, uint16_t p
     std::string externalAddress = address == "0.0.0.0" ? platform::getLocalIpAddress() : address;
     std::cout << "To connect to the OAK visualizer use the following:\n";
     std::cout << "If you are opening the visualizer locally, open http://localhost:" << port << " in your browser\n";
-    if (port == 8082) {
-        std::cout << "If you are using peripheral mode and viewing from another machine, open http://"  << externalAddress << ":" << port << " in your browser\n";
-        std::cout << "If you are using oakapp base image, open http://" << externalAddress << ":" << "9000" << " in your browser" << std::endl;
+    if(port == 8082) {
+        std::cout << "If you are using peripheral mode and viewing from another machine, open http://" << externalAddress << ":" << port
+                  << " in your browser\n";
+        std::cout << "If you are using oakapp base image, open https://" << externalAddress << ":" << "9000" << " in your browser" << std::endl;
     } else {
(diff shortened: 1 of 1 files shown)


examples/python/Visualizer/visualizer_encoded.py
```diff
@@ -5,11 +5,11 @@
...
 parser = ArgumentParser()
 parser.add_argument("--webSocketPort", type=int, default=8765)
-parser.add_argument("--httpPort", type=int, default=8080)
+parser.add_argument("--httpPort", type=int, default=8082)
 
 args = parser.parse_args()
 
-remoteConnector = dai.RemoteConnection(webSocketPort=args.webSocketPort, httpPort=args.httpPort)
+remoteConnector = dai.RemoteConnection(address="127.0.0.1", webSocketPort=args.webSocketPort, httpPort=args.httpPort)
 ENCODER_PROFILE = dai.VideoEncoderProperties.Profile.MJPEG
```
moratom, line 13: Needed?

src/remote_connection/RemoteConnectionImpl.cpp
```diff
@@ -35,6 +35,29 @@ static std::shared_ptr<Buffer> getVisualizableMessage(const std::shared_ptr<Buff
...
+        if(ip != "127.0.0.1" && strncmp(ifa->ifa_name, "lo", 2) != 0) {
+            result = ip;
+            break;
+        }
+    }
+
+    freeifaddrs(ifaddr);
+    return result;
+}
+
```
moratom: Is this cross-platform?
I think likely this won't work on Windows.
We can move to Platform.cpp for this and implement it for Windows as well.

examples/python/Visualizer/visualizer_encoded.py
aljazdu: Encoded stream is played only in the case of localhost: or 127.0.0.1
Without a specified address, the address of the PC is used, so the stream does not work. In the case of a standalone camera mode, i believe the encoded stream can only be seen using the nginx proxy, however this would have to be tested.

aljazdu: this has been reverted, as it's no longer needed with the updated output message

src/remote_connection/RemoteConnectionImpl.cpp
aljazdu: This is now fixed to be platform agnostic. With the updated changes (inside the platform.cpp) this was tested on the windows testbed to confirm correct behaviour.

PR #1568, opened 2025-12-02
Add missing build methods to the RGBD and SpatialDetectionNetwork nodes

lnotspotl:
## Purpose
Add missing build methods to the RGBD and SpatialDetectionNetwork nodes.

CU task: https://app.clickup.com/t/86c6kyk13


moratom (commented):
Thanks!

Tested and everything looks good apart from ToF - it seems to work at 5 FPS while the visualizer_rgbd_autocreate.py example works with 30 FPS.
We should identify the difference and fix it, so the ToF examples can run at 30 FPS.

PR #1570, opened 2025-12-03
Feat: Update spatial location calculator

aljazkonec1:
## Purpose

This PR updates the SpatialLocationCalculator node to accept `dai::ImgDetections` as input and return an updated `dai::SpatialImgDetections` message as output.

## Specification

The following changes have been made:
- New common object `dai::SpatialKeypoint`: Has same structure as `dai::Keypoint` with additional attribute `dai::Point3f spatialCoordinates`
- `dai::SpatialImgDetections` message has been updated to be in line with `dai::ImgDetections` to now support `dai::RotatedRect`, SpatialKeypoints, segmentation mask and spatial coordinates.
- Expanded `dai::SpatialLocationCalculatorConfig` to include global settings (and their getters / setters) which are used for calculating spatial coordinates of ImgDetections. The available global settings are:
   - `globalStepSize`
   - `globalLowerThreshold`
   - `globalUpperThreshold`
   - `calculateSpatialKeypoints`
   - `useSegmentation`
- `dai::SpatialLocationCalculator` now accepts a new input of `dai::ImgDetections` message and outputs a `dai::SpatialImgDetections` message on the `spatialOutput` Output link.
- SpatialLocationCalculator logic is moved SpatialUtils class on core and a new function to calculate SpatialDetections is added
- Update to proto messages
- Expanded `spatial_location_calculator_test.cpp`
- Added segmentation example in `spatial_segmentation`
- Add Pybinding for all the above mentioned classes. 

The spatial_location_calculator_test is depended on an update to `dai::ImgFrame::setCvFrame(mat, type)` which is dependent on an update to[ pybind_opencv_numpy](https://github.com/luxonis/pybind11_opencv_numpy/commit/30ac94f27a11ea73dbbeb164ec9ae41e720422d7).

SpatialLocationCalculator is backwards compatible so `SpatialDetectionNetwork` and `Tracker` nodes are not affected outside of a conversion change between `dai::SpatialImgDetection` <--> `dai::ImgDetection`. The two nodes should be updated in a later PR.

## Deployment Plan

None / not applicable

## Testing & Validation

New test and ran on examples.


tests/src/ondevice_tests/pipeline/node/spatial_location_calculator_test.cpp
```diff
@@ -175,7 +244,485 @@ TEST_CASE("SpatialLocationCalculator synthetic depth data test") {
...
+    static_cast<void>(unusedLegacy);
+    static_cast<void>(unusedPassthrough);
+    REQUIRE(spatialDetections.detections.size() == 1);
+
+    const auto& spatialDetection = spatialDetections.detections.front();
+    CHECK(spatialDetection.spatialCoordinates.z == Approx(1200.0F).margin(1.0F));
+
+    const auto keypoints = spatialDetection.getKeypoints();
+    const auto& kp = keypoints.at(0);
+    REQUIRE(kp.spatialCoordinates.z == Approx(0).margin(1.0F));
```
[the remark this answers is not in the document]
aljazkonec1: No :)

src/pipeline/utilities/SpatialLocationCalculator/SpatialUtils.cpp
```diff
@@ -0,0 +1,408 @@
...
+            logger->error("Segmentation mask expected but not available in ImgDetections.");
+            throw;
+        }
+        // add if opencv support here
+        cv::Mat seg = optSeg.value();
+        if(static_cast<int>(segmentationMaskHeight) != depthHeight || static_cast<int>(segmentationMaskWidth) != depthWidth) {
+            logger->debug("Segmentation mask size {}x{} does not match depth frame size {}x{}. Resizing to depth frame size.", segmentationMaskWidth, segmentationMaskHeight, depthWidth, depthHeight);
+            cv::resize(seg, seg, cv::Size(depthWidth, depthHeight), 0, 0, cv::INTER_NEAREST);
+        }
+        maskPtr = seg.data;
```
lnotspotl: `optSeg.value()` returns a reference to the matrix contained in `optSeq`, are we sure that it won't get deallocated randomly once the program gets out of this if block?

If so, maybe we can move the declaration of `optSeq` above the line of code with the if statement on it.

src/pipeline/datatype/SpatialImgDetections.cpp
```diff
@@ -1,11 +1,168 @@
...
+
+void SpatialImgDetection::setEdges(const std::vector<Edge> edges) {
+    if(!keypoints.has_value()) {
+        throw std::runtime_error("No keypoints set, cannot set edges.");
+    }
+    keypoints->setEdges(edges);
+}
+
+float SpatialImgDetection::getCenterX() const noexcept {
+    return getBoundingBox().center.x;
```
lnotspotl, line 148: This method, and a handful of other methods, that use `getBoundingBox` are labeled as noexcept. However, looking at the definition of it, it does have a `throw` inside. Is this valid?

src/pipeline/utilities/SpatialLocationCalculator/SpatialUtils.cpp
```diff
@@ -0,0 +1,408 @@
...
+                    sum += px;
+                    ++counter;
+                    if(px < min) min = px;
+                    if(px > max) max = px;
+                }
+            }
+        }
+        dai::SpatialImgDetection spatialDetection;
+        if(calculateSpatialKeypoints) {
+            logger->info("Calculating spatial keypoints for detection {}", i);
```
lnotspotl: This is inside of a for loop so I can imagine this beeing too verbose. Can we maybe change the log level here to be `debug` or even `trace` such that it does not overshadow other logs on the info log level?

src/pipeline/utilities/SpatialLocationCalculator/SpatialUtils.cpp
```diff
@@ -0,0 +1,408 @@
...
+                int kpx = static_cast<int>(kp.imageCoordinates.x * depthWidth);
+                int kpy = static_cast<int>(kp.imageCoordinates.y * depthHeight);
+
+                int kpxstart = std::max(0, kpx - keypointRadius);
+                int kpystart = std::max(0, kpy - keypointRadius);
+                int kpxend = std::min(depthWidth - 1, kpx + keypointRadius);
+                int kpyend = std::min(depthHeight - 1, kpy + keypointRadius);
+
+                std::uint32_t kpSum = 0;
+                std::uint16_t kpMin = 65535, max = 0;
```
lnotspotl: Just a nit pick, it's rather clear that 65535 is the maximum representable number using 16 bits, I'd make it a little more explicit though. 
```suggestion
                std::uint16_t kpMin = std::numeric_limits<uint16_t>::max(), max = std::numeric_limits<uint16_t>::min();
```

Also, is the naming here correct? There is another declaration of `max` a few lines above and it's not clear which one is intended to be used when
`float z = calculateDepth(calculationAlgorithm, validPixels, sum, counter, min, max);` happens

bindings/python/src/pipeline/datatype/SpatialImgDetectionsBindings.cpp
```diff
@@ -37,48 +38,127 @@ void bind_spatialimgdetections(pybind11::module& m, void* pCallstack) {
...
+             py::arg("label"),
+             DOC(dai, SpatialImgDetection, SpatialImgDetection))
+        .def("__repr__", &SpatialImgDetection::str)
+        .def_readwrite("label", &SpatialImgDetection::label)
+        .def_readwrite("labelName", &SpatialImgDetection::labelName)
+        .def_readwrite("confidence", &SpatialImgDetection::confidence)
+        .def_readwrite("xmin", &SpatialImgDetection::xmin, "Deprecation warning: use boundingBox instead")
+        .def_readwrite("ymin", &SpatialImgDetection::ymin)
+        .def_readwrite("xmax", &SpatialImgDetection::xmax)
+        .def_readwrite("ymax", &SpatialImgDetection::ymax)
```
lnotspotl: Should we also not include the deprecation warning for these?

examples/python/SpatialLocationCalculator/spatial_keypoints.py
```diff
@@ -0,0 +1,163 @@
...
+
+device = dai.Device()
+fps = 30
+platform = "RVC4"
+frame_type = dai.ImgFrame.Type.BGR888i
+model_name = "luxonis/yolov8-large-pose-estimation:coco-640x352"
+if device.getPlatform() == dai.Platform.RVC2:
+    model_name = "luxonis/yolov8-nano-pose-estimation:coco-512x288"
+    fps = 10
+    platform = "RVC2"
```
lnotspotl: Just FYI, there's also a method called `getPlatformAsString`, not change required though :)

examples/python/SpatialLocationCalculator/spatial_keypoints.py
```diff
@@ -0,0 +1,163 @@
...
+        assert isinstance(spatialData, dai.SpatialImgDetections)
+        assert isinstance(passthrough, dai.ImgFrame)
+        assert isinstance(depthFrame, dai.ImgFrame)
+
+        depthImg = depthFrame.getCvFrame()
+        colorizedDepth = cv2.applyColorMap(cv2.convertScaleAbs(depthImg, alpha=0.03), cv2.COLORMAP_JET)
+        image = passthrough.getCvFrame()
+
+        filter_keypoints = [0, 3, 4, 7, 8, 13, 14, 15, 16] # filter out nose, ears, elbows, knees, ankles
+        connected_keypoints = [[0, 1], [2, 3], [2, 4], [3, 5], [2, 6], [3, 7]] # indecies of keypoints that are connected with lines.
```
lnotspotl: Spelling mistake.

```suggestion
        connected_keypoints = [[0, 1], [2, 3], [2, 4], [3, 5], [2, 6], [3, 7]] # indices of keypoints that are connected with lines.
```

protos/SpatialImgDetections.proto
```diff
@@ -10,18 +9,23 @@ message SpatialImgDetections {
     common.Timestamp tsDevice = 3;
     repeated SpatialImgDetection detections = 4;
     common.ImgTransformation transformation = 5;
+    int64 segmentationMaskWidth = 6;
+    int64 segmentationMaskHeight = 7;
+    bytes data = 8;
```
lnotspotl: in `ProtoSerialize.cpp`, we copy `maskData` into `data`, can we maybe also call it the same way?

src/pipeline/datatype/SpatialImgDetections.cpp
```diff
@@ -1,11 +1,168 @@
...
+        return {};
+    }
+}
+
+std::vector<dai::Point3f> SpatialImgDetection::getKeypointSpatialCoordinates() const {
+    if(!this->keypoints.has_value()) {
+        return {};
+    }
+    return this->keypoints->getSpatialCoordinates();
+}
```
lnotspotl, line 130: To follow the  same style as the method above and bellow:

```suggestion
std::vector<dai::Point3f> SpatialImgDetection::getKeypointSpatialCoordinates() const {
    if(keypoints.has_value()) {
        return keypoints->getSpatialCoordinates();;
    } else {
        return {};
    }
}
```

src/pipeline/datatype/SpatialLocationCalculatorConfig.cpp
```diff
@@ -21,4 +21,62 @@ std::vector<SpatialLocationCalculatorConfigData> SpatialLocationCalculatorConfig
...
+    globalLowerThreshold = lowerThreshold;
+    globalUpperThreshold = upperThreshold;
+}
+
+void SpatialLocationCalculatorConfig::setCalculationAlgorithm(SpatialLocationCalculatorAlgorithm calculationAlgorithm) {
+    globalCalculationAlgorithm = calculationAlgorithm;
+}
+
+void SpatialLocationCalculatorConfig::setStepSize(int32_t stepSize) {
+    if(stepSize < -1 || stepSize >= 4000 || stepSize == 0) {
```
lnotspotl: where does the value of 4000 come from?

src/pipeline/node/DetectionParser.cpp
```diff
@@ -3,7 +3,7 @@
 #include <algorithm>
 #include <cctype>
 #include <cstddef>
-#include <cstdint>
+#include <iostream>
```
lnotspotl: Is this include necessary here? I see all the prints are commented out?

src/pipeline/node/SpatialLocationCalculator.cpp
```diff
@@ -1,17 +1,111 @@
...
     return properties;
 }
 
+void SpatialLocationCalculator::setRunOnHost(bool runOnHost) {
+    runOnHostVar = runOnHost;
+}
+
+/**
+ * Check if the node is set to run on host
+ */
```
lnotspotl: This is already in the header, is this necessary here?

src/pipeline/utilities/SpatialLocationCalculator/SpatialUtils.cpp
```diff
@@ -0,0 +1,408 @@
...
+
+                z = modeValue;
+            }
+            break;
+        }
+    }
+    return z;
+}
+
+dai::Point3f calculateSpatialCoordinates(float z, std::array<std::array<float, 3>, 3> intrinsicMatrix, dai::Point2f center) {
```
lnotspotl: I am not sure about the performance implications, this remark might be completely irrelevant, but I'd pass intrinsicMatrix by reference here, we don't modify it or use it anyway, just read from it

src/pipeline/utilities/SpatialLocationCalculator/SpatialUtils.cpp
```diff
@@ -0,0 +1,408 @@
...
+            uint16_t minVal = static_cast<uint16_t>(min);
+            uint16_t maxVal = static_cast<uint16_t>(max);
+            if(maxVal < minVal) {
+                z = 0;
+                break;
+            }
+            std::size_t valueRange = static_cast<std::size_t>(maxVal - minVal) + 1;
+
+            if(valueRange <= MAX_HISTOGRAM_RANGE) {
+                // Small range: build a dense histogram for cache-friendly mode lookup
```
lnotspotl, line 73: Love this, have you performed any speed tests to see how much faster this is?

src/pipeline/utilities/SpatialLocationCalculator/SpatialUtils.cpp
```diff
@@ -0,0 +1,408 @@
...
+
+    return dai::Point3f(x, y, z);
+}
+
+void computeSpatialData(std::shared_ptr<dai::ImgFrame> depthFrame,
+                        const std::vector<dai::SpatialLocationCalculatorConfigData>& configDataVec,
+                        std::vector<dai::SpatialLocations>& spatialLocations,
+                        std::shared_ptr<spdlog::async_logger> logger) {
+    logger->info("Compute spatial data!");
+    if(configDataVec.size() == 0) {
```
lnotspotl: Nitpick, for constency and good practices
```suggestion
    if(configDataVec.empty()) {
```

src/pipeline/utilities/SpatialLocationCalculator/SpatialUtils.cpp
```diff
@@ -0,0 +1,408 @@
...
+    std::int32_t depthHeight = depthFrame->getHeight();
+    uint8_t* maskPtr = nullptr;
+    if(useSegmentation) {
+        logger->debug("Using segmentation for spatial image calculations.");
+        std::size_t segmentationMaskWidth = imgDetections.getSegmentationMaskWidth();
+        std::size_t segmentationMaskHeight = imgDetections.getSegmentationMaskHeight();
+        std::optional<cv::Mat> optSeg = imgDetections.getCvSegmentationMask();
+        if(!optSeg.has_value()) {
+            logger->error("Segmentation mask expected but not available in ImgDetections.");
+            throw;
```
lnotspotl: What is the expected behavior of this throw? As far as I know, it rethrows the same exception caught when used within a try-catch block, however that is not the case here.

src/pipeline/utilities/SpatialLocationCalculator/SpatialUtils.cpp
```diff
@@ -0,0 +1,408 @@
...
+    if(!depthFrame->transformation.isValid()) {
+        logger->error("DepthFrame has invalid Image Transformations. Cannot get intrinsic matrix, skipping computation of spatial coordinates.");
+        return;
+    }
+
+    spatialLocations.resize(configDataVec.size());
+    std::int32_t depthWidth = depthFrame->getWidth();
+    std::int32_t depthHeight = depthFrame->getHeight();
+    for(int i = 0; i < static_cast<int>(configDataVec.size()); i++) {
+        const auto& cfg = configDataVec[i];
```
lnotspotl: Nitpick

```suggestion
    for(size_t i = 0; i < configDataVec.size(); i++) {
        const auto& cfg = configDataVec[i];
```

src/pipeline/utilities/SpatialLocationCalculator/SpatialUtils.hpp
```diff
@@ -0,0 +1,27 @@
+#pragma once
+#include <spdlog/async_logger.h>
+
+#include <depthai/pipeline/datatype/ImgFrame.hpp>
+#include <depthai/pipeline/datatype/SpatialLocationCalculatorData.hpp>
+
+#include "depthai/pipeline/datatype/SpatialImgDetections.hpp"
+#include "depthai/pipeline/node/SpatialLocationCalculator.hpp"
+#include "depthai/properties/SpatialLocationCalculatorProperties.hpp"
```
lnotspotl: circular include (in SpatialLocationCalculator), these two are not used here anyways so we can likely remove them, right?

src/pipeline/utilities/SpatialLocationCalculator/SpatialUtils.cpp
```diff
@@ -0,0 +1,408 @@
+#include "SpatialUtils.hpp"
+
+#include <algorithm>
+#include <array>
+#include <cstdint>
+#include <opencv2/opencv.hpp>
```
lnotspotl: This won't compile if `DEPTHAI_HAVE_OPENCV_SUPPORT` is `OFF`

protos/SpatialImgDetections.proto
aljazkonec1: I'm not sure what you mean, just renaming to maskData ?

src/pipeline/datatype/SpatialLocationCalculatorConfig.cpp
aljazkonec1: Largest image dimension. Moved into int MAX_STEP_SIZE = 4000; // largest dimension of the image

src/pipeline/utilities/SpatialLocationCalculator/SpatialUtils.cpp
aljazkonec1: I haven't done a specific speed test for these two, but doing some math:
1) Histogram path: O(valueRange + validPixels) and sequential reads (with large cache hits)
2) Else path: sort = O(validPixels * log validPixels) + O(validPixels)

ValueRange is small compared to validPixels so I would venture to guess the a few-times speedup.

src/pipeline/utilities/SpatialLocationCalculator/SpatialUtils.cpp
aljazkonec1: Thanks, created a thread local std::vector<uint8_t> maskData outside the if statement to keep the data allocated.

src/pipeline/utilities/SpatialLocationCalculator/SpatialUtils.cpp
aljazkonec1: Yes, forgot to update to throw runtime_error instead of just logger error.

src/pipeline/utilities/SpatialLocationCalculator/SpatialUtils.cpp
aljazkonec1: Thanks for catching!

src/pipeline/datatype/SpatialImgDetections.cpp
```diff
@@ -122,10 +122,12 @@ std::vector<SpatialKeypoint> SpatialImgDetection::getKeypoints() const {
 }
 
 std::vector<dai::Point3f> SpatialImgDetection::getKeypointSpatialCoordinates() const {
-    if(!this->keypoints.has_value()) {
+    if(keypoints.has_value()) {
+        return keypoints->getSpatialCoordinates();
+        ;
```
lnotspotl: extra semicolon

protos/SpatialImgDetections.proto
lnotspotl: yes, you've already made the changes, thanks! :100:

src/pipeline/node/DetectionParser.cpp
```diff
@@ -93,7 +93,6 @@ void DetectionParser::setConfig(const dai::NNArchiveVersionedConfig& config) {
 
     DAI_CHECK_V(numYoloHeads == 1, "NNArchive should contain exactly one YOLO head. Found {} YOLO heads.", numYoloHeads);  // no support for multi-head YOLO
     const auto head = (*model.heads)[yoloHeadIndex];
-
     if(head.parser == "YOLO" || head.parser == "YOLOExtendedParser") {
         properties.parser.nnFamily = DetectionNetworkType::YOLO;
```
moratom: Whitespace change

src/pipeline/utilities/SpatialLocationCalculator/SpatialUtils.hpp
```diff
@@ -0,0 +1,26 @@
...
+namespace SpatialUtils {
+
+void computeSpatialData(std::shared_ptr<dai::ImgFrame> depthFrame,
+                        const std::vector<dai::SpatialLocationCalculatorConfigData>& configDataVec,
+                        std::vector<dai::SpatialLocations>& spatialLocations,
+                        std::shared_ptr<spdlog::async_logger> logger);
+
+void computeSpatialDetections(std::shared_ptr<dai::ImgFrame> depthFrame,
+                              const std::shared_ptr<SpatialLocationCalculatorConfig> config,
+                              dai::ImgDetections& imgDetections,
```
moratom: Can `imgDetections` be const here?

src/pipeline/utilities/SpatialLocationCalculator/SpatialUtils.hpp
```diff
@@ -0,0 +1,26 @@
...
+                              dai::SpatialImgDetections& spatialDetections,
+                              std::shared_ptr<spdlog::async_logger> logger);
```
moratom: Better to do `const std::shared_ptr<spdlog::async_logger>&` as ownership is not problematic

protos/SpatialImgDetections.proto
```diff
@@ -10,18 +9,23 @@ message SpatialImgDetections {
...
+    float confidence = 2;
+    float xmin = 3;
+    float ymin = 4;
+    float xmax = 5;
+    float ymax = 6;
+    string labelName = 7;
+    common.RotatedRect boundingBox = 8;
+    common.SpatialKeypointsList keypoints = 9;
+    Point3f spatialCoordinates = 10;
+    SpatialLocationCalculatorConfigData boundingBoxMapping = 11;
```
moratom: We should check if visualizer is using this - as this is a breaking change.

tests/CMakeLists.txt
```diff
@@ -476,7 +476,7 @@ target_compile_definitions(neural_network_test PRIVATE BLOB_PATH="${mobilenet_bl
 dai_set_test_labels(neural_network_test ondevice rvc2_all ci)
 
 dai_add_test(spatial_location_calculator_test src/ondevice_tests/pipeline/node/spatial_location_calculator_test.cpp)
-dai_set_test_labels(spatial_location_calculator_test ondevice rvc2_all rvc4 ci)
+dai_set_test_labels(spatial_location_calculator_test ondevice onhost rvc2_all rvc4 ci)
```
moratom: `onhost` implies only on host, aka device is not needed to run the test.


Likely not the case here

include/depthai/pipeline/node/SpatialLocationCalculator.hpp
```diff
@@ -1,5 +1,7 @@
 #pragma once
 
+#include <spdlog/async_logger.h>
```
moratom: spdlog should not be in the header files, it's not part of the public API

include/depthai/pipeline/node/SpatialLocationCalculator.hpp
```diff
@@ -11,20 +13,26 @@ namespace dai {
...
-    Properties& getProperties();
+    Properties& getProperties() override;
 
    public:
     SpatialLocationCalculator() = default;
-    SpatialLocationCalculator(std::unique_ptr<Properties> props);
+    SpatialLocationCalculator(std::unique_ptr<Properties> props)
+        : DeviceNodeCRTP(std::move(props)), roiConfig(std::make_shared<SpatialLocationCalculatorConfig>(properties.roiConfig)) {}
 
+    std::shared_ptr<SpatialLocationCalculatorConfig> roiConfig;
```
moratom: `inputConfig` perhaps? Considering it doesn't only configure ROIs anymore

include/depthai/pipeline/datatype/SpatialLocationCalculatorConfig.hpp
```diff
@@ -66,38 +65,161 @@ struct SpatialLocationCalculatorConfigData {
...
+ * - Use keypoints for ImgDetections: false
+ * - Use segmentation for ImgDetections: true
+ *
+ * An optional list of per-ROI configurations is available via `config`. ROI
+ * settings override the corresponding global values where specified.
+ *
+ * @note The ROI-based API (`setROIs`, `addROI`) is scheduled for deprecation in future releases.
+ * Users are encouraged to utilize the global configuration
+ * methods instead.
  */
```
moratom: What do we mean here? The ROIs are a runtime thing, I think we can potentially make a new config with just the ROIs so these are decoupled, but you can't use the global configuration to add them right?

include/depthai/pipeline/datatype/SpatialLocationCalculatorConfig.hpp
```diff
@@ -66,38 +65,161 @@ struct SpatialLocationCalculatorConfigData {
...
+     * @warning Will be deprecated in future releases.
      */
     void setROIs(std::vector<SpatialLocationCalculatorConfigData> ROIs);
+
     /**
-     * Add a new ROI to configuration data.
-     * @param roi Configuration parameters for ROI (region of interest)
+     * Add a new region of interest (ROI) to configuration data.
+     * @param roi Configuration parameters for ROI
+     * @warning Will be deprecated in future releases.
```
moratom: I think we can deprecate now (and provide an alternative in the message)

include/depthai/pipeline/datatype/SpatialLocationCalculatorConfig.hpp
```diff
@@ -66,38 +65,161 @@ struct SpatialLocationCalculatorConfigData {
...
+     * @warning Only applicable to ImgDetections with keypoints.
+     */
+    bool getCalculateSpatialKeypoints() const;
+
+    /*
+     * Retrieve whether segmentation is used for spatial location calculation.
+     * @warning Only applicable to ImgDetections with segmentation masks.
+     */
+    bool getUseSegmentation() const;
+
```
moratom: Maybe we can also add `setCarrySegmentation` to allow for messages to useSegmentation but not carry it in the message for bandwidth.

include/depthai/pipeline/datatype/SpatialLocationCalculatorData.hpp
```diff
@@ -3,6 +3,7 @@
 #include <unordered_map>
 #include <vector>
 
+#include "depthai/common/Point3f.hpp"
```
moratom, line 6: Needed?

src/pipeline/datatype/SpatialLocationCalculatorConfig.cpp
```diff
@@ -2,6 +2,7 @@
 
 namespace dai {
 
+int MAX_STEP_SIZE = 4000;
```
moratom: `constexpr int`

src/pipeline/datatype/SpatialLocationCalculatorConfig.cpp
```diff
@@ -21,4 +22,62 @@ std::vector<SpatialLocationCalculatorConfigData> SpatialLocationCalculatorConfig
...
+    globalUpperThreshold = upperThreshold;
+}
+
+void SpatialLocationCalculatorConfig::setCalculationAlgorithm(SpatialLocationCalculatorAlgorithm calculationAlgorithm) {
+    globalCalculationAlgorithm = calculationAlgorithm;
+}
+
+void SpatialLocationCalculatorConfig::setStepSize(int32_t stepSize) {
+    if(stepSize < -1 || stepSize >= MAX_STEP_SIZE || stepSize == 0) {
+        throw std::invalid_argument("Step size must be -1 (AUTO) or between 1 and 3999");
```
moratom: Is the 4000 special in some way? (Seems oddly specific)

include/depthai/pipeline/node/SpatialLocationCalculator.hpp
```diff
@@ -36,22 +44,47 @@ class SpatialLocationCalculator : public DeviceNodeCRTP<DeviceNode, SpatialLocat
      */
     Input inputConfig{*this, {"inputConfig", DEFAULT_GROUP, false, 4, {{{DatatypeEnum::SpatialLocationCalculatorConfig, false}}}, DEFAULT_WAIT_FOR_MESSAGE}};
 
+    /**
+     * Input messages on which spatial location will be calculated.
+     * Possible datatypes are ImgDetections or Keypoints.
+     */
+    Input input{*this, {"input", DEFAULT_GROUP, true, 1, {{{DatatypeEnum::ImgDetections, false}}}, DEFAULT_WAIT_FOR_MESSAGE}};
```
moratom: I'd go with inputDetections & outputDetections to make it clear that those two match

examples/python/SpatialLocationCalculator/spatial_keypoints.py
```diff
@@ -0,0 +1,161 @@
...
+    monoRight = pipeline.create(dai.node.Camera).build(dai.CameraBoardSocket.CAM_C, sensorFps=fps)
+    stereo = pipeline.create(dai.node.StereoDepth)
+
+    monoLeftOut = monoLeft.requestOutput((w, h))
+    monoRightOut = monoRight.requestOutput((w, h))
+    monoLeftOut.link(stereo.left)
+    monoRightOut.link(stereo.right)
+    stereo.setRectification(True)
+    stereo.setExtendedDisparity(True)
+    stereo.setLeftRightCheck(True)
```
moratom: Optional NeuralDepth would be nice here, non blocker for a merge though

src/pipeline/node/SpatialLocationCalculator.cpp
```diff
@@ -1,17 +1,108 @@
...
+    logger->debug("Start SpatialLocationCalculator");
+
+    bool inputConfigSync = inputConfig.getWaitForMessage();
+    auto eepromData = getParentPipeline().getEepromData();
+    if(!eepromData) {
+        logger->warn("No calibration data, using default FOV!");
+    }
+
+    while(isRunning()) {
+        logger->debug("Running");
```
moratom: Maybe remove, or at least change to trace

src/pipeline/utilities/SpatialLocationCalculator/SpatialUtils.cpp
```diff
@@ -0,0 +1,428 @@
...
+
+        spatialLocations[i] = spatialData;
+    }
+}
+
+void computeSpatialDetections(std::shared_ptr<dai::ImgFrame> depthFrame,
+                              const std::shared_ptr<SpatialLocationCalculatorConfig> config,
+                              dai::ImgDetections& imgDetections,
+                              dai::SpatialImgDetections& spatialDetections,
+                              std::shared_ptr<spdlog::async_logger> logger) {
```
moratom: const shared_ptr<> &
(for logger, config, depthFrame & imgDetections) as only spatialDetections need to be mutable here

src/pipeline/utilities/SpatialLocationCalculator/SpatialUtils.cpp
```diff
@@ -0,0 +1,428 @@
...
+        throw std::runtime_error("No transformation set on ImgDetections. Cannot compute spatial coordinates.");
+    }
+
+    std::vector<dai::ImgDetection> imgDetectionsVector = imgDetections.detections;
+    if(imgDetectionsVector.size() == 0) {
+        return;
+    }
+    spatialDetections.detections.resize(imgDetectionsVector.size());
+
+    std::optional<std::vector<std::uint8_t>> optMaskData = imgDetections.getMaskData();
```
moratom: This creates a copy, better to avoid if we can

src/pipeline/utilities/SpatialLocationCalculator/SpatialUtils.cpp
```diff
@@ -0,0 +1,428 @@
...
+        boundingBoxMapping.depthThresholds.lowerThreshold = lowerThreshold;
+        boundingBoxMapping.depthThresholds.upperThreshold = upperThreshold;
+        boundingBoxMapping.calculationAlgorithm = calculationAlgorithm;
+        spatialDetection.boundingBoxMapping = boundingBoxMapping;
+        spatialDetections.detections[i] = spatialDetection;
+    }
+
+    if(optMaskData) {
+        spatialDetections.setSegmentationMask(*optMaskData, imgDetections.getSegmentationMaskWidth(), imgDetections.getSegmentationMaskHeight());
+    }
```
moratom: This creates another copy, we should likely just do `spatialDetection.data = imgDetections.data` as `data` is a shared pointer and can then be reused between the two.
+ as mentioned in the header, I think it would be nice to be able to not attach the segmentation mask while still using it to improve the spatial detections

src/pipeline/utilities/SpatialLocationCalculator/SpatialUtils.cpp
```diff
@@ -0,0 +1,428 @@
...
+            throw std::runtime_error("Segmentation mask size does not match depth frame size, and OpenCV support is not available for resizing.");
+#endif
+        } else {
+            if(!optMaskData) {
+                throw std::runtime_error("Segmentation enabled but mask data missing");  // should never trigger
+            }
+            maskPtr = optMaskData->data();
+        }
+    }
+
```
moratom: Do we need to match the depth size?
We could map the indexes to the mask instead here to improve on the performance.

src/pipeline/node/SpatialLocationCalculator.cpp
```diff
@@ -1,17 +1,108 @@
...
+                utilities::SpatialUtils::computeSpatialDetections(imgFrame, roiConfig, *imgDetections, *outputSpatialImgDetections, logger);
+
+                auto stop = high_resolution_clock::now();
+                auto timeToComputeSpatialDetections = duration_cast<microseconds>(stop - start);
+                logger->trace("Time to compute spatial detections: {} us", timeToComputeSpatialDetections.count());
+            }
+            outputSpatialImgDetections->setSequenceNum(imgFrame->getSequenceNum());
+            outputSpatialImgDetections->setTimestampDevice(imgFrame->getTimestampDevice());
+            outputSpatialImgDetections->setTimestamp(imgFrame->getTimestamp());
+            outputSpatialImgDetections->transformation = imgFrame->transformation;
```
moratom: Let's double check this, I think if depth is not aligned to RGB we need to copy those from ImgDetections, not the input depth.

In case depth is aligned it generally doesn't matter.

src/pipeline/datatype/SpatialLocationCalculatorConfig.cpp
aljazkonec1: Its half of OAK4 camera sensor resolution "Max Resolution 48MP (8000x6000)"

include/depthai/pipeline/datatype/SpatialLocationCalculatorData.hpp
aljazkonec1: yes, line 56:
```
    /**
     *  Spatial coordinates - x,y,z; x,y are the relative positions of the center of ROI to the center of depth map
     */
    dai::Point3f spatialCoordinates;
```

protos/SpatialImgDetections.proto
aljazkonec1: Fixed [here](https://github.com/luxonis/depthai-core/pull/1570/commits/c2b7c8578b27f846983ea57e6766631aa76e6377)

examples/python/SpatialLocationCalculator/spatial_segmentation.py
```diff
@@ -0,0 +1,103 @@
...
+                outer_points = outer_points.getPoints()
+                outer_points = [[int(p.x), int(p.y)] for p in outer_points]
+                outer_points = np.array(outer_points, dtype=np.int32)
+                cv2.polylines(image, [outer_points], isClosed=True, color=(0, 255, 0), thickness=2)
+
+                # depth
+                depth_coordinate = det.spatialCoordinates
+                depth = depth_coordinate.z
+                text = f"X: {int(depth_coordinate.x / 10 )} cm, Y: {int(depth_coordinate.y / 10)} cm, Z: {int(depth / 10)} cm"
+                cv2.putText(image, text, outer_points[0], cv2.FONT_HERSHEY_PLAIN, 1, (232,36,87), 2)
```
moratom: Maybe split visualization in a separate function so "main" is cleaner

examples/python/SpatialLocationCalculator/spatial_segmentation.py
```diff
@@ -0,0 +1,103 @@
...
+    spatial_calculator = pipeline.create(dai.node.SpatialLocationCalculator)
+    spatial_calculator.initialConfig.setUseSegmentation(True)
+    align.outputAligned.link(spatial_calculator.inputDepth)
+    det_nn.out.link(spatial_calculator.inputDetections)
+
+    cam_queue = det_nn.passthrough.createOutputQueue()
+    spatial_output_queue = spatial_calculator.outputDetections.createOutputQueue()
+    depth_queue = spatial_calculator.passthroughDepth.createOutputQueue()
+
+    input_config_queue = spatial_calculator.inputConfig.createInputQueue()
```
moratom: camelCase should be used consistently (examples included)

src/pipeline/datatype/ImgTransformations.cpp
```diff
@@ -459,4 +459,29 @@ dai::RotatedRect ImgTransformation::remapRectFrom(const ImgTransformation& from,
...
+    for(int i = 0; i < 3; ++i) {
+        for(int j = 0; j < 3; ++j) {
+            float expectedValue = (i == j) ? 1.0f : 0.0f;
+            if(!approxEqual(relativeTransform[i][j] / scale, expectedValue)) {
+                return false;
+            }
+        }
+    }
+    return true;
+}
```
moratom, line 449: Would be good to add test cases for that.

src/pipeline/datatype/ImgTransformations.cpp
```diff
@@ -459,4 +459,29 @@ dai::RotatedRect ImgTransformation::remapRectFrom(const ImgTransformation& from,
     return transformed;
 }
 
+bool ImgTransformation::isAlignedTo(const ImgTransformation& to) const {
+    if(width != to.width || height != to.height) return false;
+    if(this->distortionModel != to.distortionModel) return false;
```
moratom, line 427: Distortion coefficients should be checked too

examples/cpp/SpatialLocationCalculator/spatial_segmentation.cpp
```diff
@@ -0,0 +1,149 @@
...
+        auto monoLeft = pipeline.create<dai::node::Camera>();
+        monoLeft->build(dai::CameraBoardSocket::CAM_B, std::nullopt, fps);
+        auto monoRight = pipeline.create<dai::node::Camera>();
+        monoRight->build(dai::CameraBoardSocket::CAM_C, std::nullopt, fps);
+
+        auto stereo = pipeline.create<dai::node::StereoDepth>();
+        stereo->setSubpixel(false);
+        stereo->setRectification(true);
+        stereo->setExtendedDisparity(true);
+        stereo->setLeftRightCheck(true);
```
moratom: any reason to diverge from defaults? Maybe we can use the default preset?

examples/cpp/SpatialLocationCalculator/spatial_segmentation.cpp
```diff
@@ -0,0 +1,149 @@
...
+
+        auto align = pipeline.create<dai::node::ImageAlign>();
+        stereo->depth.link(align->input);
+        detectionNetwork->passthrough.link(align->inputAlignTo);
+
+        auto spatialCalculator = pipeline.create<dai::node::SpatialLocationCalculator>();
+        spatialCalculator->initialConfig->setUseSegmentation(true);
+        align->outputAligned.link(spatialCalculator->inputDepth);
+        detectionNetwork->out.link(spatialCalculator->inputDetections);
+
```
moratom, line 112: Once we have the `SpatialDetectionNetwork` ready we should simplify this example by using it IMO.

examples/python/SpatialLocationCalculator/spatial_segmentation.py
```diff
@@ -0,0 +1,103 @@
...
+
+        cv2.imshow("Depth", colorizedDepth)
+        cv2.imshow("Spatial detections", concatenated_frame)
+        key = cv2.waitKey(1)
+        if key == ord('q'):
+            break
+
+        elif key == ord('s'):
+            calculatorConfig.setUseSegmentation(not calculatorConfig.getUseSegmentation())
+            input_config_queue.send(calculatorConfig)
```
moratom: camelCase should be used consistnetly (and the comments from the C++ example apply here as well)

examples/python/SpatialLocationCalculator/spatial_keypoints.py
```diff
@@ -0,0 +1,161 @@
...
+h = nnArchive.getInputHeight()
+assert w is not None
+assert h is not None
+
+with dai.Pipeline(device) as pipeline:
+    print("Creating pipeline...")
+
+    camera_node = pipeline.create(dai.node.Camera).build(dai.CameraBoardSocket.CAM_A)
+    cam_out =  camera_node.requestOutput((w, h), frame_type, fps= fps, enableUndistortion=True)
+    monoLeft = pipeline.create(dai.node.Camera).build(dai.CameraBoardSocket.CAM_B, sensorFps=fps)
```
moratom: camelCase

include/depthai/pipeline/datatype/SpatialLocationCalculatorData.hpp
```diff
@@ -52,7 +53,7 @@ struct SpatialLocations {
     /**
      *  Spatial coordinates - x,y,z; x,y are the relative positions of the center of ROI to the center of depth map
      */
-    Point3f spatialCoordinates;
+    dai::Point3f spatialCoordinates;
```
moratom: Shouldn't be needed, it's in the dai namespace.

include/depthai/pipeline/node/SpatialLocationCalculator.hpp
```diff
@@ -11,20 +11,26 @@ namespace dai {
...
-    Properties& getProperties();
+    Properties& getProperties() override;
 
    public:
     SpatialLocationCalculator() = default;
-    SpatialLocationCalculator(std::unique_ptr<Properties> props);
+    SpatialLocationCalculator(std::unique_ptr<Properties> props)
+        : DeviceNodeCRTP(std::move(props)), calculationConfig(std::make_shared<SpatialLocationCalculatorConfig>(properties.roiConfig)) {}
 
+    std::shared_ptr<SpatialLocationCalculatorConfig> calculationConfig;
```
moratom: why is this duplicated with initialConfig?

protos/SpatialImgDetections.proto
```diff
@@ -1,27 +1,25 @@
...
 message SpatialImgDetections {
     int64 sequenceNum = 1;
     common.Timestamp ts = 2;
     common.Timestamp tsDevice = 3;
     repeated SpatialImgDetection detections = 4;
     common.ImgTransformation transformation = 5;
+    int64 segmentationMaskWidth = 6;
+    int64 segmentationMaskHeight = 7;
+    bytes maskData = 8;
 }
```
moratom, line 16: Backwards compatible, nice 👍

src/pipeline/datatype/SpatialLocationCalculatorConfig.cpp
moratom: I would remove this, we can throw in runtime based on size if a step size bigger than half the resolution doesn't make sense.

src/pipeline/node/SpatialLocationCalculator.cpp
```diff
@@ -1,17 +1,118 @@
...
+            }
+        }
+        if(hasConfig) {
+            calculationConfig = config;
+        }
+
+        imgFrame = inputDepth.get<dai::ImgFrame>();
+        if(imgFrame == nullptr) {
+            logger->critical("Invalid input depth.");
+        }
```
moratom: If this happens we'll crash two lines later -> we should skip processing in this case.

src/pipeline/node/SpatialLocationCalculator.cpp
```diff
@@ -1,17 +1,118 @@
...
+        if(imgFrame->getType() != dai::ImgFrame::Type::RAW16) {
+            logger->critical("Invalid frame type for depth image. Depth image must be RAW16 type, got {}", static_cast<int>(imgFrame->getType()));
+        }
```
moratom: We should likely skip processing in this case too

src/pipeline/node/SpatialLocationCalculator.cpp
```diff
@@ -1,17 +1,118 @@
...
+            if(imgDetections != nullptr) {
+                start = high_resolution_clock::now();
+                utilities::SpatialUtils::computeSpatialDetections(*imgFrame, *calculationConfig, *imgDetections, *outputSpatialImgDetections, logger);
+
+                auto stop = high_resolution_clock::now();
+                auto timeToComputeSpatialDetections = duration_cast<microseconds>(stop - start);
+                logger->trace("Time to compute spatial detections: {} us", timeToComputeSpatialDetections.count());
+            }
+            outputSpatialImgDetections->setSequenceNum(imgDetections->getSequenceNum());
+            outputSpatialImgDetections->setTimestampDevice(imgDetections->getTimestampDevice());
```
moratom: if imgDetections == nullptr -> imgDetections->getSequenceNum() will segfault

src/pipeline/utilities/SpatialLocationCalculator/SpatialUtils.cpp
```diff
@@ -0,0 +1,417 @@
...
+        spatialDetection.boundingBoxMapping = boundingBoxMapping;
+        spatialDetections.detections[i] = spatialDetection;
+    }
+
+    if(passthroughSegmentation) {
+        spatialDetections.data = imgDetections.data;
+        spatialDetections.segmentationMaskWidth = imgDetections.getSegmentationMaskWidth();
+        spatialDetections.segmentationMaskHeight = imgDetections.getSegmentationMaskHeight();
+    }
+}
```
moratom, line 458: Could be split into multiple functions to improve readibility

src/pipeline/utilities/SpatialLocationCalculator/SpatialUtils.cpp
```diff
@@ -0,0 +1,417 @@
...
+        for(int y = ystart; y < yend; y += stepSize) {
+            for(int x = xstart; x < xend; x += stepSize) {
+                int index = y * depthWidth + x;
+                uint16_t px = plane[index];
+
+                bool usePixel = true;
+                if(useSegmentation) {
+                    dai::Point2f maskPoint{static_cast<float>(x), static_cast<float>(y)};
+                    if(!areAligned) {
+                        maskPoint = depthTransformation->remapPointTo(*detectionsTransformation, maskPoint);
```
moratom: Maybe worth adding a comment here that this is expansive in a hot loop, in case we'll want to improve (same for keypoints)

examples/cpp/SpatialLocationCalculator/spatial_segmentation.cpp
aljazkonec1: I would make that a separate example as we still want to showcase SpatialLocationCalculator because it can be used with all detection models (yunet, scrfd, palm detection, text detection, ...)

examples/python/SpatialLocationCalculator/spatial_segmentation.py
aljazkonec1: [fix](https://github.com/luxonis/depthai-core/pull/1570/changes/92198c7811749276d8fd00772ac45768ff9cb3f8)

examples/python/SpatialLocationCalculator/spatial_segmentation.py
aljazkonec1: [Small helper function](https://github.com/luxonis/depthai-core/pull/1570/changes/3e64dcec02d4dd1ae4553dea142babe16e2b2bf4), but I would leave the rest as is as it showcases many nice functions SpatialLocationCalculator and SpatialImgDetections offer

src/pipeline/datatype/ImgTransformations.cpp
aljazkonec1: [Test case](https://github.com/luxonis/depthai-core/pull/1570/changes/047e0726ce7aa7699880bc6221b0394e2fad4858)

src/pipeline/datatype/ImgTransformations.cpp
aljazkonec1: Distortion coefficients [fix](https://github.com/luxonis/depthai-core/pull/1570/changes/047e0726ce7aa7699880bc6221b0394e2fad4858)

src/pipeline/datatype/SpatialLocationCalculatorConfig.cpp
aljazkonec1: [Moved to runtime](https://github.com/luxonis/depthai-core/pull/1570/changes/6f8b6b95e9de0e2183568b7fa6fa9807d32d86f1)

examples/cpp/SpatialLocationCalculator/spatial_segmentation.cpp
aljazkonec1: [Removed](https://github.com/luxonis/depthai-core/pull/1570/changes/59c5f34d13d71aaf5a5b60544d69f2bdcadea390)

examples/python/SpatialLocationCalculator/spatial_segmentation.py
aljazkonec1: [Fixed](https://github.com/luxonis/depthai-core/pull/1570/changes/59c5f34d13d71aaf5a5b60544d69f2bdcadea390)

examples/python/SpatialLocationCalculator/spatial_keypoints.py
aljazkonec1: [Fixed](https://github.com/luxonis/depthai-core/pull/1570/changes/59c5f34d13d71aaf5a5b60544d69f2bdcadea390)

include/depthai/pipeline/datatype/SpatialLocationCalculatorData.hpp
aljazkonec1: [Removed](https://github.com/luxonis/depthai-core/pull/1570/changes/6f8b6b95e9de0e2183568b7fa6fa9807d32d86f1)

include/depthai/pipeline/node/SpatialLocationCalculator.hpp
aljazkonec1: Moved to private [here](https://github.com/luxonis/depthai-core/pull/1570/changes/7376acaffb8fa6aaa85ec06e49f000c8b189527b)

src/pipeline/node/SpatialLocationCalculator.cpp
aljazkonec1: [Added skip](https://github.com/luxonis/depthai-core/pull/1570/changes/a005a288da999b8eef2f20d48906f7d2ad3fa3d7)

src/pipeline/node/SpatialLocationCalculator.cpp
aljazkonec1: [Added skip](https://github.com/luxonis/depthai-core/pull/1570/changes/a005a288da999b8eef2f20d48906f7d2ad3fa3d7)

src/pipeline/node/SpatialLocationCalculator.cpp
aljazkonec1: Thanks! [FIxed](https://github.com/luxonis/depthai-core/pull/1570/changes/a005a288da999b8eef2f20d48906f7d2ad3fa3d7)

src/pipeline/utilities/SpatialLocationCalculator/SpatialUtils.cpp
aljazkonec1: Thanks! [Split up the function](https://github.com/luxonis/depthai-core/pull/1570/changes/976599bbf175889deee4503314ccaf2e4eba8645)

PR #1586, opened 2025-12-15
Fix MacOS wheels by disabling symbol stripping

lnotspotl:
## Purpose
Stripping symbols in the combiner script renders macos wheels unusable, importing installed depthai leads to the python interpreter being killed. This PR fixes that.

### DIFF
--- a/bindings/python/ci/combine_wheels.py
+++ b/bindings/python/ci/combine_wheels.py
@@ -448,7 +448,10 @@ def _combine_wheels_macos(input_folder, output_folder, strip):
 
     if strip:
         _logger.warning("Stripping debug symbols is not supported on MacOS")
-        strip = False # wheels do not work on some users' machines, TODO: figure out why
+        # wheels do not work on some users' machines, TODO: figure out why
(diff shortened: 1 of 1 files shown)


bindings/python/ci/combine_wheels.py
```diff
@@ -517,9 +524,9 @@ def _combine_wheels_macos(input_folder, output_folder, strip):
                     lib_path = os.path.join(wheel_extract_dir, palace_dylibs, lib)
                     subprocess.run(["strip", "-S", lib_path], check=True)
 
-            # Strip debug symbols for the cpython library
-            # TODO: Invalid signature error, fix this
-            _logger.debug(f"Found cpython lib: {cpython_lib}, but not stripping debug symbols for it")
```
moratom, line 522: Do we want to have this as a waring considering it runs every time?

moratom (commented):
Thanks @lnotspotl !

Any idea why the smoke test didn't catch the issue?
Locally `import depthai` was enough to triggger it.

PR #1587, opened 2025-12-15
CalibrationHandler Cycling Calibration issue

MaticTonin:
Idea is to prevent user to even set a calibration which has a cycling calibration set in it. So the validateExtrisniscGraph function was added which:
- reads EEROM data
- check if all toSockets created in extrinsics calibration even exists (dangling reference)
- checks if there is a cycling connection present to any origin.

The function will be present always on the setExtrisnics, while in constructors there is a toggle of using it or not.

Additional tests were added, only questions I have is, do we wanna add python bindings test for the calibrationHandler as well or not? @lnotspotl

### DIFF
--- a/examples/cpp/AprilTags/CMakeLists.txt
+++ b/examples/cpp/AprilTags/CMakeLists.txt
@@ -12,7 +12,7 @@ if(DEPTHAI_FETCH_ARTIFACTS)
         SHA1 "6818a531e71948bd28e1f0ab3e76b18aff6150fb"
         LOCATION april_tags
     )
-endif
+endif()
 
 dai_add_example(april_tags april_tags.cpp ON OFF)
 dai_set_example_test_labels(april_tags ondevice rvc2_all rvc4 rvc4rgb ci)


include/depthai/device/CalibrationHandler.hpp
```diff
@@ -38,31 +38,35 @@ class CalibrationHandler {
      * eeprom json file created from calibration procedure.
      *
      * @param eepromDataPath takes the full path to the json file containing the calibration and device info.
+     * @param boardConfigPath Enable internal check for extrinsics cycling links or dangling references.
```
lnotspotl: Inconsistent naming: rename `boardConfigPath` to `validateExtrinsics`

include/depthai/device/CalibrationHandler.hpp
```diff
@@ -38,31 +38,35 @@ class CalibrationHandler {
...
-    explicit CalibrationHandler(std::filesystem::path eepromDataPath);
+    explicit CalibrationHandler(std::filesystem::path eepromDataPath, bool validateExtrinsics = false);
 
     /**
      * Construct a new Calibration Handler object using the board
      * config json file and .calib binary files created using gen1 calibration.
      *
      * @param calibrationDataPath Full Path to the .calib binary file from the gen1 calibration. (Supports only Version 5)
      * @param boardConfigPath Full Path to the board config json file containing device information.
+     * @param boardConfigPath Enable internal check for extrinsics cycling links or dangling references.
```
lnotspotl: same as obove (there's other places too)

src/device/CalibrationHandler.cpp
```diff
@@ -49,7 +49,57 @@ void invertSe3Matrix4x4InPlace(std::vector<std::vector<float>>& mat) {
 }
 }  // namespace
 
-CalibrationHandler::CalibrationHandler(std::filesystem::path eepromDataPath) {
+CalibrationHandler::ExtrinsicGraphValidationResult CalibrationHandler::validateExtrinsicGraph() const {
+    std::unordered_set<CameraBoardSocket> globalVisited;
```
lnotspotl: necessary? It does not seem to be used anywhere.

src/device/CalibrationHandler.cpp
```diff
@@ -49,7 +49,57 @@ void invertSe3Matrix4x4InPlace(std::vector<std::vector<float>>& mat) {
 }
 }  // namespace
 
-CalibrationHandler::CalibrationHandler(std::filesystem::path eepromDataPath) {
+CalibrationHandler::ExtrinsicGraphValidationResult CalibrationHandler::validateExtrinsicGraph() const {
```
lnotspotl, line 52: Will this work if there's two disconnected subgraphs, for instance CAM_A -> CAM_B -> CAM_C and then CAM_D -> CAM_E?

Also, can we detect self cycles, that is instances such as CAM_A -> CAM_A?

tests/src/onhost_tests/calibration_handler_test.cpp
```diff
@@ -1,13 +1,14 @@
 #include <catch2/catch_all.hpp>
 #include <depthai/device/CalibrationHandler.hpp>
+#include <fstream>
```
lnotspotl: Necessary include?

tests/src/onhost_tests/calibration_handler_test.cpp
```diff
@@ -310,28 +338,49 @@ TEST_CASE("Directly linked cameras translation", "[getCameraTranslationVector]")
 }
 
 TEST_CASE("Cyclic connection throws", "[getCameraTranslationVector]") {
```
lnotspotl: How is this test different from `TEST_CASE("Cyclic connection throws", "[getCameraExtrinsics]")`, also the test case description seems wrong (we're not testing the getCameraTranslationVector method here (removed))

tests/src/onhost_tests/calibration_handler_test.cpp
```diff
@@ -432,32 +481,64 @@ TEST_CASE("Directly linked cameras rotation", "[getCameraRotationMatrix]") {
 }
 
 TEST_CASE("Cyclic connection throws", "[getCameraRotationMatrix]") {
```
lnotspotl: missing getCameraRotationMatrix test?

include/depthai/device/CalibrationHandler.hpp
```diff
@@ -38,31 +38,35 @@ class CalibrationHandler {
      * eeprom json file created from calibration procedure.
      *
      * @param eepromDataPath takes the full path to the json file containing the calibration and device info.
+     * @param boardConfigPath Enable internal check for extrinsics cycling links or dangling references.
      */
-    explicit CalibrationHandler(std::filesystem::path eepromDataPath);
+    explicit CalibrationHandler(std::filesystem::path eepromDataPath, bool validateExtrinsics = false);
```
lnotspotl: I think we should validate these by default in the future and so how about we use std::optional here and if not set (std::nullopt), we will issue a warning that this will be changed in a future release (e.g. 3.5.0) to true (automatic checking).

If case of `std::nullopt`, we will refrain from validating the extrinsincs. Also, maybe instead of `validateExtrinsics`, could we rename it to `validateCalibration` such that in the future we can keep on adding tests (e.g. for the intrinsics and what not, data that's not yet part eeprom but might be in the future)

src/device/CalibrationHandler.cpp
MaticTonin: Wdym? Mybe a comment to different function since this one is main logic of the PR?

src/device/CalibrationHandler.cpp
MaticTonin: Hm, interesting case! Will add a test for it and see if we manage to catch it!

tests/src/onhost_tests/calibration_handler_test.cpp
MaticTonin: Will remove those cycling connection tests for seperate getCameraTranslationVector and getCameraRotationMatrix, since they are redundant now.

tests/src/onhost_tests/calibration_handler_test.cpp
MaticTonin: Which one exacly?

src/device/CalibrationHandler.cpp
MaticTonin: Will this work if there's two disconnected subgraphs, for instance CAM_A -> CAM_B -> CAM_C and then CAM_D -> CAM_E?

Based on the implementation, those will for now not be caught, due to links always having a reference point, even tho there are then two points of origin and as well no references between the CAM_ABC to CAM_DE.
We can expand, this, so this will be caught as well, but then all pipelines would have to have "closed loop reference with one origin" so that every link can be calculated. 

CAM_A -> CAM_A is fine, since it will return idenity but will add test.

include/depthai/device/CalibrationHandler.hpp
MaticTonin: Adressed https://github.com/luxonis/depthai-core/commit/0e74ce163c3f306ae7e20761cf48f32e2349d129

include/depthai/device/CalibrationHandler.hpp
MaticTonin: Adressed and renamed, good point!
https://github.com/luxonis/depthai-core/commit/0e74ce163c3f306ae7e20761cf48f32e2349d129

include/depthai/device/CalibrationHandler.hpp
MaticTonin: Adressed https://github.com/luxonis/depthai-core/commit/0e74ce163c3f306ae7e20761cf48f32e2349d129

tests/src/onhost_tests/calibration_handler_test.cpp
MaticTonin: Removed, legacy while testing https://github.com/luxonis/depthai-core/commit/0e74ce163c3f306ae7e20761cf48f32e2349d129

tests/src/onhost_tests/calibration_handler_test.cpp
MaticTonin: Removed, added a setCalibration test rather
https://github.com/luxonis/depthai-core/commit/0e74ce163c3f306ae7e20761cf48f32e2349d129

src/device/CalibrationHandler.cpp
```diff
@@ -49,7 +49,82 @@ void invertSe3Matrix4x4InPlace(std::vector<std::vector<float>>& mat) {
...
+        std::unordered_set<CameraBoardSocket> localVisited;
+        CameraBoardSocket current = start;
+
+        while(true) {
+            if(localVisited.count(current)) {
+                return {CalibrationHandler::ExtrinsicGraphError::CycleDetected, current, eepromData.cameraData.at(current).extrinsics.toCameraSocket};
+            }
+
+            localVisited.insert(current);
+            globalVisited.insert(current);
```
lnotspotl: I do not understand the merit of this variable, we seem to be inserting into this set but we never read from it or do anything else that's useful with it.

src/device/CalibrationHandler.cpp
```diff
@@ -60,15 +135,23 @@ CalibrationHandler::CalibrationHandler(std::filesystem::path eepromDataPath) {
     }
     nlohmann::json jsonData = nlohmann::json::parse(jsonStream);
     eepromData = jsonData;
+    if(validateCalibration) {
```
lnotspotl: now that `validateCalibration` is of type std::optional, we should  do `validateCalibration.value_or(false)` - operator bool checks whether the variable has been set

bindings/python/src/CalibrationHandlerBindings.cpp
```diff
@@ -25,11 +25,19 @@ void CalibrationHandlerBindings::bind(pybind11::module& m, void* pCallstack) {
...
-
-        .def_static("fromJson", &CalibrationHandler::fromJson, DOC(dai, CalibrationHandler, fromJson))
+        .def(py::init<std::filesystem::path, bool>(),
+             py::arg("eepromDataPath"),
+             py::arg("validateExtrinsics") = false,
+             DOC(dai, CalibrationHandler, CalibrationHandler, 2))
+        .def(py::init<std::filesystem::path, std::filesystem::path, bool>(),
+             py::arg("calibrationDataPath"),
+             py::arg("boardConfigPath"),
+             py::arg("validateExtrinsics") = false,
```
lnotspotl: let's do std::nullopt here instead of false (in case we decide to change the default value to true)

src/device/CalibrationHandler.cpp
MaticTonin: For global variable I agree, it is a legacy of my trying of implementation of cycling dependency

src/device/CalibrationHandler.cpp
MaticTonin: Added.

bindings/python/src/CalibrationHandlerBindings.cpp
MaticTonin: Added!

PR #1591, opened 2025-12-17
Add back docker containers

moratom:
## Purpose
Add a workflow to build development docker containers with preinstalled DepthAI C++ and Python libraries

None / not applicable

## Deployment Plan

A Docker container will get pushed to Docker hub with every push on `develop` `main` and `tags`

## Testing & Validation

Tested manually & added basic smoke tests to the container build itself to make sure a simple app can be compiled.
A container from the branch can be tested with:
```
docker pull luxonis/depthai-library:a1b078ba480981b9e81251bfd37799be644052b9
```

### DIFF
--- a/.github/workflows/containers.yml
+++ b/.github/workflows/containers.yml
@@ -1,26 +1,23 @@
 name: Publish Docker image
 on:
   workflow_dispatch:
-  # TODO(Morato) - re-enable docker upload
-  # push:
-  #   branches:
-  #   - main
-  #   - develop
-  #   tags:
-  #   - 'v2*'
+  push:
+    branches:
+    - main
+    - develop
+    tags:
+    - 'v3*'
 jobs:
 
   build:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-latest, [Linux, ARM], [Linux, ARM64]]
+        os: [ubuntu-latest, ubuntu-24.04-arm]
         include:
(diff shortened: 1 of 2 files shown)


.github/workflows/containers.yml
```diff
@@ -1,26 +1,23 @@
...
-  # push:
-  #   branches:
-  #   - main
-  #   - develop
-  #   tags:
-  #   - 'v2*'
+  push:
+    branches:
+    - main
+    - develop
```
lnotspotl, line 7: Do we want this to be dispatched on every push to `develop`?

ci/Dockerfile
```diff
@@ -0,0 +1,128 @@
...
+        python3-setuptools \
+        python3-wheel && \
+    rm -rf /var/lib/apt/lists/*
+
+RUN python3 -m pip install --break-system-packages opencv-python-headless
+
+# ------------------------------------------------------------
+# Copy final installed artifacts only
+# ------------------------------------------------------------
+COPY --from=builder /usr/local /usr/local
```
lnotspotl, line 86: I assume the python wheels also gets installed somewhere in /usr/local, correct?

.github/workflows/containers.yml
moratom: We've had this configured like so in `v2`, we'll reduce the rate if we'll start to hit issues with storage.

PR #1595, opened 2025-12-18
Feature/snaps callbacks

aljazdu:
## Purpose
This PR solves multiple minor issues with Snaps such as: missing callbacks for snap uploads, exposed device serial numbers when sending snaps, minor update of certain errors messages.

## Specification
Added CallBacks to SendSnap(), which returns SendSnapCallbackResult which contains snap payload, name, timestamp, id, as well as SendSnapCallbackStatus which describes the status of snap upload.

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation
This was tested using the updated examples. both in python and C++.

### DIFF
--- a/examples/cpp/Events/events.cpp
+++ b/examples/cpp/Events/events.cpp
@@ -59,8 +59,10 @@ int main() {
 
     pipeline.start();
 
+    std::cout << "Press 's' to send a snap, and 'q' to quit" << std::endl;
     while(pipeline.isRunning()) {
-        if(cv::waitKey(1) != -1) {
+        auto key = cv::waitKey(1);
+        if(key == 'q') {
             break;
         }
 
@@ -98,7 +100,7 @@ int main() {
         }
 
         // Trigger sendSnap()
-        if(cv::waitKey(1) == 's') {
+        if(key == 's') {
             auto localSnapID = eventsManager->sendSnap("ImageDetection",
                                                        std::nullopt,
                                                        inRgb,
--- a/examples/python/Events/events.py
+++ b/examples/python/Events/events.py
@@ -48,8 +48,10 @@ with dai.Pipeline() as pipeline:
         return (np.clip(np.array(bbox), 0, 1) * normVals).astype(int)
 
 
+    print("Press 's' to send a snap, and 'q' to quit")
     while pipeline.isRunning():
-        if cv2.waitKey(1) != -1:
+        key = cv2.waitKey(1)
+        if key == ord("q"):
             pipeline.stop()
             break
 
@@ -88,7 +90,7 @@ with dai.Pipeline() as pipeline:
             cv2.imshow("rgb", frame)
 
         # Trigger sendSnap()
-        if cv2.waitKey(1) == ord("s"):
+        if key == ord("s"):
             localSnapID = eventMan.sendSnap("ImageDetection", None, inRgb, inDet, ["EventsExample", "Python"], {"key_0" : "value_0", "key_1" : "value_1"}, 
                               uploadSuccessCallback, uploadFailureCallback)
             if localSnapID != None:
--- a/examples/python/Events/events_file_group.py
+++ b/examples/python/Events/events_file_group.py
@@ -47,9 +47,10 @@ with dai.Pipeline() as pipeline:
         return (np.clip(np.array(bbox), 0, 1) * normVals).astype(int)
 
 
+    print("Press 'q' to quit")
     counter = 0
     while pipeline.isRunning():
-        if cv2.waitKey(1) != -1:
+        if cv2.waitKey(1) == ord("q"):
             pipeline.stop()
             break
 
--- a/examples/cpp/Events/events_file_group.cpp
+++ b/examples/cpp/Events/events_file_group.cpp
@@ -61,7 +61,7 @@ int main() {
 
     int counter = 0;
     while(pipeline.isRunning()) {
-        if(cv::waitKey(1) != -1) {
+        if(cv::waitKey(1) == 'q') {
             break;
         }
 


examples/cpp/Events/events.cpp
```diff
@@ -72,7 +99,14 @@ int main() {
...
         if(cv::waitKey(1) == 's') {
-            eventsManager->sendSnap("ImageDetection", std::nullopt, inRgb, inDet, {"EventsExample", "C++"}, {{"key_0", "value_0"}, {"key_1", "value_1"}});
+            eventsManager->sendSnap("ImageDetection",
+                                    std::nullopt,
+                                    inRgb,
+                                    inDet,
+                                    {"EventsExample", "C++"},
+                                    {{"key_0", "value_0"}, {"key_1", "value_1"}},
+                                    uploadSuccessCallback,
+                                    uploadFailureCallback);
```
moratom: What's the way to go if one wants to match which Snap exactly failed to upload?
Is the only way to go to use a unique name here?

Maybe we can return a snapID here so this is not a requirement?

aljazdu: Currently a snap name and a timestamp are returned for both the successful and unsuccessful snaps, while the snapID (assigned by the hub) is returned only for the successful snaps (hub implementation). I have raised this concern with their team, and they agreed that snap ids (and consequent identification) is only required for the successful snaps: 

- "I think keeping the id as null for failed snaps is fine" & 

- "yeah I think since they are not in Hub, its fine, otherwise you would need some local IDs you return from sendSnap"

bindings/python/src/utility/EventsManagerBindings.cpp
```diff
@@ -7,6 +7,10 @@
 
 void EventsManagerBindings::bind(pybind11::module& m, void* pCallstack) {
     using namespace dai;
+
+    // Type definitions
+    py::enum_<dai::utility::SendSnapCallbackStatus> sendSnapCallbackStatus(m, "SendSnapCallbackStatus", DOC(dai, SendSnapCallbackStatus));
```
lnotspotl: maybe we can also move this a bit down to the DEPTHAI_ENABLE_EVENTS_MANAGER section?

src/utility/EventsManager.cpp
```diff
@@ -848,7 +961,7 @@ void EventsManager::uploadCachedData() {
                     fileGroup->fileData.push_back(fileData);
                 }
             }
-            snapData->event = event;
+            snapData->eventData->event = event;
```
lnotspotl, line 968: Isn't eventData here still a null pointer?

src/utility/EventsManager.cpp
```diff
@@ -582,6 +644,21 @@ void EventsManager::uploadEventBatch() {
         logger::error("Failed to send event, status code: {}", response.status_code);
         // In case the eventBuffer gets too full (dropped connection), cache the events or drop them
         if(eventBuffer.size() >= EVENT_BUFFER_MAX_SIZE) {
+            // failureCallback
+            for(size_t index = 0; index < eventBuffer.size(); ++index) {
```
lnotspotl, line 649: Shouldn't eventBufferMutex be locked here?

src/utility/EventsManager.cpp
```diff
@@ -419,6 +434,20 @@ void EventsManager::uploadFileBatch(std::deque<std::shared_ptr<SnapData>> inputS
             eventBufferCondition.wait_for(lock, duration, [this]() { return stopUploadThread.load(); });
             // After retrying a defined number of times, we can determine the connection is not established, cache if enabled
             if(retryAttempt >= uploadRetryPolicy.maxAttempts) {
+                for(size_t index = 0; index < inputSnapBatch.size(); ++index) {
+                    if(inputSnapBatch.at(index)->eventData->callbacks.has_value()) {
+                        auto event = inputSnapBatch.at(index)->eventData->event;
+                        std::string serializedPayload;
+                        event->SerializeToString(&serializedPayload);
+                        inputSnapBatch.at(index)->eventData->callbacks.value().second(
```
lnotspotl: What is callbacks.value().second is a nullptr (no callback set)? Maybe instead of using a pair, we can introduce yet another helper struct, something like 

```
struct SnapCallbacks {
    std::function<void(SendSnapCallbackResult)> onSuccess;
    std::function<void(SendSnapCallbackResult)> onFailure;
};
```

src/utility/EventsManager.cpp
```diff
@@ -444,7 +473,18 @@ void EventsManager::uploadFileBatch(std::deque<std::shared_ptr<SnapData>> inputS
                     std::string rejectionReason = dai::proto::event::RejectedFileGroupReason_descriptor()
                                                       ->FindValueByNumber(static_cast<int>(prepareGroupResult.rejected().reason()))
                                                       ->name();
-                    logger::info("A group has been rejected because of {}", rejectionReason);
+                    logger::error("A group has been rejected because of {}", rejectionReason);
+                    if(snapData->eventData->callbacks.has_value()) {
+                        auto event = snapData->eventData->event;
+                        std::string serializedPayload;
+                        event->SerializeToString(&serializedPayload);
+                        snapData->eventData->callbacks.value().second(SendSnapCallbackResult{event->name(),
```
lnotspotl: callbacks.value().second could potentially be a null pointer

src/utility/EventsManager.cpp
```diff
@@ -241,6 +250,12 @@ EventsManager::EventsManager(bool uploadCachedOnStart)
     auto containerId = utility::getEnvAs<std::string>("OAKAGENT_CONTAINER_ID", "");
     sourceAppId = appId == "" ? containerId : appId;
     sourceAppIdentifier = utility::getEnvAs<std::string>("OAKAGENT_APP_IDENTIFIER", "");
+    auto connectedDevices = DeviceBase::getAllConnectedDevices();
+    if(!connectedDevices.empty()) {
+        sourceSerialNumber = connectedDevices[0].getDeviceId();
```
lnotspotl: Is it the case that device 0 is the one used? What's the merit of logging if serial number?

examples/python/Events/events.py
```diff
@@ -6,6 +6,23 @@
 import time
 
 
+# Callback functions
+def upload_success_callback(send_snap_result):
```
lnotspotl: can we maybe add an assert on the uploadStatus so that it's clear that the only thing that it can be is SUCCESS?

lnotspotl (commented):
.

src/utility/EventsManager.cpp
aljazdu: You are correct, it is.

src/utility/EventsManager.cpp
aljazdu: It should.

examples/cpp/Events/events.cpp
aljazdu: This has been solved with the use of localIDs

bindings/python/src/utility/EventsManagerBindings.cpp
aljazdu: Agree, fixed.

src/utility/EventsManager.cpp
aljazdu: Agree, fixed.

examples/python/Events/events.py
aljazdu: Done.

src/utility/EventsManager.cpp
aljazdu: Changed to first available device.

src/utility/EventsManager.cpp
aljazdu: Agree that the use of std::pair was not optimal. Introduced separate members for onSuccess and onFailure callbacks instead of common member (std::pair). 
Avoided introducing another helper struct as it was not the best approach.

examples/python/Events/events.py
moratom: also camelCase for consistnecy instead of snake_case (for failure as well)

src/utility/EventsManager.cpp
moratom: This might be missleading, I would always leave it empty for now as there's no gurantee that the first connected device will be the one used in the application (there could even be more devices in the application).

examples/python/Events/events_file_group.py
```diff
@@ -82,6 +101,7 @@ def frameNorm(frame, bbox):
 
             fileGroup = dai.FileGroup()
             fileGroup.addImageDetectionsPair(fileName, inRgb, borderDetections)
-            eventMan.sendSnap("LowConfidenceDetection", fileGroup, ["EventsExample", "Python"], {"key_0" : "value_0", "key_1" : "value_1"})
+            eventMan.sendSnap("LowConfidenceDetection", fileGroup, ["EventsExample", "Python"], {"key_0" : "value_0", "key_1" : "value_1"}, 
+                              upload_success_callback, upload_failure_callback)
```
moratom: How can I get the localID out, to later match what snap has failed/succeded?

aljazdu: sendSnap() has been updated to return the localSnapID to the caller  (which was missing before), in the form of std::optional<std::string> - has value for successfully added snaps.
Examples have been updated as well.

src/utility/EventsManager.cpp
aljazdu: Fixed, so it's empty for the time being.

examples/python/Events/events.py
aljazdu: Fixed in the new commit.

moratom (approved):

aljazdu:
> However, the issue with me uploading snaps super quickly persists, if I spam `s`, the app just crashes, any idea why that could be?

I will look into it and let you know. In the mean time, you can use the file_group example if you want to automate sending the snaps (it sends a snap when a detection confidence is within a certain threshold). It can be used to easily send a large number of snaps.

PR #1596, opened 2025-12-18
Refactor CrashDump

lnotspotl:
## Purpose
This PR refactors the contemporary CrashDump data structure to facilitate work with crashdumps from various sources, including those generated by RVC2 and RVC4 devices. The newly introduced virtual `CrashDump` class allows adding new types of crashdumps by creating a subclass and implementing just a handful of virtual methods, notably for archiving and loading a crashdump to and from an archive, respectively. There are currently two crashdump implementations, specifically `CrashDumpRVC2` and `CrashDumpRVC4`.

A new method was introduced to the `DeviceBase` class, called `registerCrashdumpCallback` that allows users to add their custom callback function called with a collected crashdump as its only argument, allowing users to add their own extra information or do further processing before letting depthai automatically store the updated crashdump in a cache folder.

```python3
import os
os.environ["DEPTHAI_CRASH_DEVICE"] = "1"

import time
import depthai as dai

device = dai.Device()
def my_callback(dump): # dump, a reference (shared_ptr) to the crashdump
    print(f"Got dump: {dump.getPlatform()}")

    device.registerCrashdumpCallback(lambda x: None)

    dump.extra["current_time"] = time.time()  # has to be convertible by `to_json`, check `pybind11_json.hpp`
    dump.extra["current_time2"] = time.time()
    dump["current_time3"] = time.time() # same as dump.extra["current_time3"] = ...

    extra = dump.extra
    extra["dds"] = "beee"
    dump.extra = extra

device.registerCrashdumpCallback(my_callback)
device.crashDevice()
time.sleep(60)
```

Later on, one can load a stored crashdump back into memory via
```python3
import depthai as dai
dump = dai.CrashDump.load("/bla/bla/crash_dump.tar.gz")
```

Compression utilities were enhanced to include a function for compressed file archiving, the deflating function stays untouched and supports unpacking of both compressed (`.tar.gz`) and uncompressed (`.tar`) files.

Finally, this PR adds a way to programatically check whether or not a device has crashed. Note, however, that a device might crash following which the re-connection mechanism might establish a new connection, thereby making the device work properly again. In that case, only for the duration of the reconnection routine does the newly introduced `hasCrashed` method return True, after a successful reconnection, this method returns False again.

Both `hasCrashed` and crashdump collection were tested on RVC2 and RVC4 devices, each in combination with all supported operating systems, that is Windows, MacOS, and Ubuntu Linux. 

CU task: https://app.clickup.com/t/86c5e030a

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -775,4 +775,4 @@ dai_set_test_labels(auto_calibration_test ondevice rvc4 rvc2_all ci)
 
 # Crashdump test
 dai_add_test(crashdump_test src/ondevice_tests/crashdump_test.cpp)
-dai_set_test_labels(crashdump_test ondevice rvc2_all rvc4)
+dai_set_test_labels(crashdump_test ondevice rvc2_all rvc4 ci)


src/device/DeviceBase.cpp
```diff
@@ -546,7 +547,8 @@ void DeviceBase::closeImpl() {
                 if(found && (rebootingDeviceInfo.state == X_LINK_UNBOOTED || rebootingDeviceInfo.state == X_LINK_BOOTLOADER)) {
                     pimpl->logger.trace("Found rebooting device in {}ns", duration_cast<nanoseconds>(steady_clock::now() - t1).count());
                     DeviceBase rebootingDevice(config, rebootingDeviceInfo, firmwarePath, true);
-                    if(rebootingDevice.hasCrashDump()) {
+                    crashed = rebootingDevice.hasCrashDump();
+                    if(crashed) {
                         auto dump = rebootingDevice.getCrashDump();
                         logCollection::logCrashDump(pipelineSchema, dump, deviceInfo);
                     } else {
```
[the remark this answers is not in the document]
aljazdu: @lnotspotl to my understanding, Copilot's comment is wrong, since we are tying the `crashed` detection to the existence of `crashDump`. Correct?

src/device/DeviceBase.cpp
```diff
@@ -1105,6 +1112,7 @@ void DeviceBase::monitorCallback(std::chrono::milliseconds watchdogTimeout, Prev
             }
             if(reconnectionCallback) reconnectionCallback(ReconnectionStatus::RECONNECTED);
             pimpl->logger.warn("Reconnection successful\n");
+            crashed = hasCrashDump();
```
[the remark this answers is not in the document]
aljazdu: Hijacking Copilot's comment again. @lnotspotl , this is meant to reset the crashed flag after a successful reconnection,  right? 
This is a nitpick, but since this line gets executed only in case of a successful reconnection (break in case of unsuccessful reconnection), I would move it next to the 
`reconnected = true;`
line, so it's grouped together, for easier understanding.

src/utility/Platform.cpp
```diff
@@ -155,22 +167,28 @@ void setThreadName(JoiningThread& thread, const std::string& name) {
...
+    } else {
+        tmpPath = std::filesystem::temp_directory_path();
+    }
+    return tmpPath;
 #else
+    std::string tmpPath;
     char tmpTemplate[] = "/tmp/depthai_XXXXXX";
     char* tmpName = mkdtemp(tmpTemplate);
     if(tmpName == nullptr) {
         tmpPath = "/tmp";
```
aljazdu: Suppose that `tmpPath` gets set to `"/tmp"` . Loading a crash dump uses `fs::removeAll(tempDir)` , this would remove the entire tmp dir, containing other system and application tmp files.

lnotspotl: True, good catch. Maybe we should throw right away? I think the temp dir creation is very unlikely to fail but in case it does, I'd stop the program execution right away instead of trying to work around it. Thoughts?

aljazdu: Agree with you, it would be better to stop it in this case.

include/depthai/device/CrashDump.hpp
```diff
@@ -2,16 +2,125 @@
...
+    /* Extra information - user-specified additional information */
+    nlohmann::json extra = nlohmann::json::object();
+
+   protected:
+    // Writers
+    void writeMetadata(nlohmann::json& json) const;
+
+    // Readers
+    void readMetadata(const nlohmann::json& json);
+};
```
moratom, line 120: Can we add crashId? It's returned by the server when the crash is uploaded.

src/device/CrashDumpManager.hpp
```diff
@@ -0,0 +1,48 @@
...
+    bool hasCrashDump();
+
+   private:
+    std::unique_ptr<CrashDump> collectDumpRVC2(bool clear);
+    std::unique_ptr<CrashDump> collectDumpRVC4();
+
+    DeviceBase* devicePtr;
+};
+
+}  // namespace dai
```
moratom, line 48: Intended to be used by customers?

include/depthai/device/DeviceBase.hpp
```diff
@@ -414,7 +428,7 @@ class DeviceBase {
     /**
      * Retrieves crash dump for debugging.
      */
-    dai::CrashDump getCrashDump(bool clearCrashDump = true);
+    CrashDumpRVC2::CrashReportCollection getCrashDump(bool clearCrashDump = true);
```
moratom: I think it make sense to return the new high level object here instead?
(with commit hash, ...)

include/depthai/device/DeviceBase.hpp
```diff
@@ -778,6 +792,27 @@ class DeviceBase {
...
     bool isClosed() const;
 
+    /**
+     * Register a callback that will be called when the device crashes.
+     * The callback receives a shared pointer to the CrashDump, which can be modified
+     * before it is stored. Only one callback can be registered at a time.
+     *
+     * @param callback Callback to call when a crash dump is collected
+     */
+    void registerCrashdumpCallback(std::function<void(std::shared_ptr<CrashDump>)> callback);
```
moratom, line 829: We can add a nice test now that we have both a way to crash device and register a callback for it.
Just make sure the crashdump upload is disabled in the test, so we don't spam the CI.

src/device/DeviceBase.cpp
```diff
@@ -451,6 +453,23 @@ unsigned int getCrashdumpTimeout(XLinkProtocol_t protocol) {
     return timeoutMs;
 }
 
+void DeviceBase::collectAndLogCrashDump(DeviceBase* device) {
```
moratom, line 460: How come we explicitly pass in a pointer and not use the implicit `this`?

src/device/DeviceBase.cpp
```diff
@@ -451,6 +453,23 @@ unsigned int getCrashdumpTimeout(XLinkProtocol_t protocol) {
...
+    std::shared_ptr<CrashDump> crashDump = dai::CrashDumpManager(device).collectCrashDump();
+    if(crashDump) {
+        decltype(crashdumpCallback) callbackCopy;
+
+        // Create a copy of the callback function to avoid race conditions
+        // (user could set a new callback from within the callback, leading to a deadlock)
+        {
+            std::unique_lock<std::mutex> l(crashdumpCallbackMtx);
+            callbackCopy = crashdumpCallback;
+        }
```
moratom, line 472: IMO not necessary (if one tries to register another callback inside the callback I think he's in a deadlock anyhow).

src/device/DeviceGate.cpp
```diff
@@ -299,51 +299,36 @@ std::optional<std::vector<uint8_t>> DeviceGate::getFile(const std::string& fileU
...
-                }
-                spdlog::warn("FW crashed - trying to get out the crash dump");
-                std::this_thread::sleep_for(std::chrono::seconds(3));  // Allow for the generation of the crash dump and the log file
-                std::string crashDumpName;
-                spdlog::warn("Getting the crash dump out - this can take up to a minute, because it first needs to be compressed.");
-                return getCrashDump();
+                return;  // Session is destroyed, we can exit
         }
+        std::this_thread::sleep_for(std::chrono::milliseconds(500));
     }
```
moratom, line 760: I think the new behavior only checks for crashed, but not destroyed, is this intentional?

src/utility/Platform.cpp
lnotspotl: Addressed in [973145d](https://github.com/luxonis/depthai-core/pull/1596/commits/973145d8aa026a159a4b2f903455fc706a063413)

src/device/CrashDumpManager.hpp
lnotspotl: You are right, I moved it to `src`

src/device/DeviceBase.cpp
lnotspotl: What if I set another callback from within the callback, the first callback being useful while the latter acting as a passthrough? 

```python3
def callback1(dump):
    device.registerCrashdumpCallback(lambda x: None)
    print("First callback, doing something useful")
```

moratom: Won't `device.registerCrashdumpCallback(...)` try to capture the same mutex as the the mutex that's already captured to run the `callback1`?

lnotspotl: yeah, before the callback is called, we create a copy of it and so this should not pose an issue

PR #1607, opened 2026-01-07
Remove GH spamming of precheck when PR is open

MaticTonin:
The main idea is to remove spamming of GH actions when PR is open and when commits are pushed on it, since there was a lot of cases when we have branch labeled as testable and then push additional commits on it due to review changes, example below:

image

This should as well limit the usage of our GH runnners, but would mean that retesting of the actions would require to add + remove the label since GH no longer checks new pushed changes.

### DIFF
--- a/.github/workflows/main.workflow.yml
+++ b/.github/workflows/main.workflow.yml
@@ -1,5 +1,9 @@
 name: DepthAI Core CI/CD
 
+concurrency:
+  group: depthai-core-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
 on:
   workflow_dispatch:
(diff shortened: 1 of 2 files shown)


aljazdu (approved):
Changes LGTM, as I agree that the spamming of empty GH workflows was annoying. The perfect solution would entail a workflow run only when a combination of **synchronize** _and_ **labeled** would be in effect, but since we are limited by GH workflows this can not be achieved - as it is not supported.
Since re-labeling of testable will be "annoying" in this case, I would suggest that we stick to labeling when PR is already reviewed - perhaps a check for run workflows can be added at some point to avoid merging without running the workflows.

PR #1610, opened 2026-01-08
Fix: Nullptr derefrence in calibration init

aljazkonec1:
## Purpose

If stereo calibration changes while running, a nullptr dereference would happen. This could be seen by running [dynamic_calibration.py](https://github.com/luxonis/depthai-core/blob/3df0856e11902b6c8cc5442b7de8ab74efb94569/examples/python/DynamicCalibration/calibration_dynamic.py#L101)  and moving the RVC4 camera to trigger a calibration update.

## Specification

 The fix is purely on RVC4 device side.

RVC4 device side.

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable


moratom:
> I would wait with merging, so that we introduce as well here a test to catch this error. so it won't repeat again.

Can you merge your change to enable the test on RVC4 here? That should cover it.

MaticTonin:
WIll add that as well yes.

PR #1613, opened 2026-01-08
Add RVC4 StereoDepth presents (these shall be platform specific)

lnotspotl:
Clickup task: https://app.clickup.com/t/86c7bvth1

### DIFF
--- a/src/pipeline/node/StereoDepth.cpp
+++ b/src/pipeline/node/StereoDepth.cpp
@@ -396,6 +396,10 @@ void StereoDepth::setRvc2ProfilePreset(PresetMode mode) {
 
             setPostProcessingHardwareResources(3, 3);
         } break;
+
+        default:
+            DAI_CHECK_V(false, "Unsupported preset mode: {}", static_cast<int>(mode));
+            break;
     }
 }
 
@@ -476,6 +480,10 @@ void StereoDepth::setRvc4ProfilePreset(PresetMode mode) {
             initialConfig->setExtendedDisparity(true);
             initialConfig->setLeftRightCheck(true);
         } break;
+
+        default:
+            DAI_CHECK_V(false, "Unsupported preset mode: {}", static_cast<int>(mode));
+            break;
     }
 }
 


tests/src/ondevice_tests/stereo_depth_node_test.cpp
```diff
@@ -31,12 +31,12 @@ void testStereoDepthPreset(dai::node::StereoDepth::PresetMode preset, dai::Proce
     }
 }
 
-TEST_CASE("Test StereoDepth node FAST_ACCURACY preset") {
```
MaticTonin, line 34: Wouldnt at this stage be better to basicly loop through all nodes Presets of each node having its own test? sounds more robust and as well easier to maintain.

src/pipeline/node/StereoDepth.cpp
```diff
@@ -375,6 +399,86 @@ void StereoDepth::setDefaultProfilePreset(PresetMode mode) {
...
+            initialConfig->postProcessing.holeFilling.minValidDisparity = 1;
+            initialConfig->postProcessing.spatialFilter.enable = false;
+            initialConfig->postProcessing.speckleFilter.enable = true;
+            initialConfig->postProcessing.speckleFilter.speckleRange = 200;
+
+            initialConfig->setConfidenceThreshold(97);
+            initialConfig->setMedianFilter(MedianFilter::MEDIAN_OFF);
+            initialConfig->setExtendedDisparity(true);
+            initialConfig->setLeftRightCheck(true);
+        } break;
```
MaticTonin, line 482: Do we need to as well handle cases where some Presets only exists for RVC4 and not for RVC2? If so, additional case would be required to say something like:
"warning: the following Preset is does not exists on this platform, reverting to default"

moratom: I think all are handled now, but better to add a default handle here as you mentioned

tests/src/ondevice_tests/stereo_depth_node_test.cpp
lnotspotl: I think I'd keep it separate, honestly. In case of a failure, one can selectively pick a test that fails while ignoring the rest. If we were to put everything in a loop, then we would not have such an option.

src/pipeline/node/StereoDepth.cpp
lnotspotl: There's also a compiler flag that we're using that makes the compiler throw an error whenever it notices an unhandled enum case: `-Werror=switch-enum`

tests/src/ondevice_tests/stereo_depth_node_test.cpp
MaticTonin: Good point! Then we can keep it seperate for now.

src/pipeline/node/StereoDepth.cpp
MaticTonin: Looks good to me!

PR #1631, opened 2026-01-15
FPS calculation fixes for pipeline debugging

asahtik:
## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

### DIFF
--- a/include/depthai/pipeline/MessageQueue.hpp
+++ b/include/depthai/pipeline/MessageQueue.hpp
@@ -88,9 +88,11 @@ class MessageQueue : public std::enable_shared_from_this<MessageQueue> {
         return *this;
     }
 
-    static std::unordered_map<std::string, std::shared_ptr<ADatatype>> getAny(std::unordered_map<std::string, MessageQueue&> queues, std::optional<std::chrono::milliseconds> timeout = std::nullopt);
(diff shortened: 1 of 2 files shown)


src/pipeline/MessageQueue.cpp
```diff
@@ -215,7 +215,7 @@ void MessageQueue::notifyCondVars() {
 
 MessageQueue::QueueException::~QueueException() noexcept = default;
 
-std::unordered_map<std::string, std::shared_ptr<ADatatype>> MessageQueue::getAny(std::unordered_map<std::string, MessageQueue&> queues) {
```
moratom, line 218: We should add a test w/ the timeout

src/pipeline/MessageQueue.cpp
```diff
@@ -243,19 +243,35 @@ std::unordered_map<std::string, std::shared_ptr<ADatatype>> MessageQueue::getAny
...
+                            return true;
+                        }
+                    }
+                    for(auto& kv : queues) {
+                        if(kv.second.has()) {
+                            return true;
+                        }
+                    }
+                    return false;
+                });
```
moratom: Can we define the lambda outside and reuse it?

PR #1634, opened 2026-01-16
Feature/gate node

JakubFara:
# Gate Node

## Purpose

* Regulate the queue during runtime.

## Possible Improvements

* Is it possible to create a command that sends a certain number of images. This might be a useful feature to have.

## Examples

* `examples/cpp/Gate/gate_example.cpp`
* `examples/python/Gate/gate_example.py`

## Tests

* `tests/src/ondevice_tests/pipeline/node/gate_node_tests.cpp`

## FW
- implemented for RVC2 and RVC4

### DIFF
--- a/cmake/Depthai/DepthaiDeviceRVC4Config.cmake
+++ b/cmake/Depthai/DepthaiDeviceRVC4Config.cmake
@@ -3,4 +3,4 @@
 set(DEPTHAI_DEVICE_RVC4_MATURITY "snapshot")
 
 # "version if applicable"
-set(DEPTHAI_DEVICE_RVC4_VERSION "0.0.1+a412c85954366d09defede43d775cd676d68173f")
+set(DEPTHAI_DEVICE_RVC4_VERSION "0.0.1+5d393dbecabf3343abfa9fd0746cbd82ba3b99ae")
--- a/cmake/Depthai/DepthaiDeviceSideConfig.cmake
+++ b/cmake/Depthai/DepthaiDeviceSideConfig.cmake
@@ -2,7 +2,7 @@
 set(DEPTHAI_DEVICE_SIDE_MATURITY "snapshot")
 
 # "full commit hash of device side binary"
-set(DEPTHAI_DEVICE_SIDE_COMMIT "c2eb5a3661dbe60ee3a67f00a24d269a7ecce8e7")
+set(DEPTHAI_DEVICE_SIDE_COMMIT "6d8681c877c0329ae9283946ba1e3af8d8a08214")
 
 # "version if applicable"
 set(DEPTHAI_DEVICE_SIDE_VERSION "")


bindings/python/src/pipeline/datatype/GateControl.cpp
MaticTonin, line 1: Probably a file duplication since same bindings are created as well in GateControlBindings.cpp?

examples/cpp/Gate/gate_example.cpp
```diff
@@ -0,0 +1,60 @@
...
+    auto cameraQueue = gate->output.createOutputQueue();
+
+    auto gateControlQueue = gate->inputControl.createInputQueue();
+
+    auto gateControl = std::make_shared<dai::GateControl>();
+    gateControl->stop();
+
+    pipeline.start();
+
+    auto start_time = std::chrono::steady_clock::now();
```
MaticTonin: camelCase

examples/python/Gate/gate_example.py
```diff
@@ -0,0 +1,55 @@
...
+    cameraQueue = gate.output.createOutputQueue()
+    gateControlQueue = gate.inputControl.createInputQueue()
+
+    gateControl = dai.GateControl()
+    gateControl.stop()
+
+    pipeline.start()
+
+    ellapsed = 0
+    start_time = time.monotonic()
```
MaticTonin: camelCase

examples/python/Gate/gate_example.py
```diff
@@ -0,0 +1,55 @@
...
+import cv2 
+
+deviceInfo = dai.DeviceInfo()
+device = dai.Device(deviceInfo)
+
+withGate = True 
+
+with dai.Pipeline(device) as pipeline:
+    camera = pipeline.create(dai.node.Camera).build()
+    cameraOut = camera.requestFullResolutionOutput(fps=30)
```
MaticTonin: As discussed, it would be better to lower resolution to achieve higher FPS as done in C++ example

examples/python/Gate/gate_example.py
```diff
@@ -0,0 +1,55 @@
+import os
+import sys
+import time
+import depthai as dai
+import cv2 
+
+deviceInfo = dai.DeviceInfo()
+device = dai.Device(deviceInfo)
```
MaticTonin: Probably not required to do deviceInfo.

include/depthai/pipeline/node/Gate.hpp
```diff
@@ -0,0 +1,39 @@
...
+
+    Input input{*this, {"input", DEFAULT_GROUP, false, 4, {{{DatatypeEnum::Buffer, true}}}, DEFAULT_WAIT_FOR_MESSAGE}};
+
+    Output output{*this, {"output", DEFAULT_GROUP, {{{DatatypeEnum::Buffer, true}}}}};
+
+    Input inputControl{*this, {"inputControl", DEFAULT_GROUP, false, 4, {{{DatatypeEnum::GateControl, false}}}, DEFAULT_WAIT_FOR_MESSAGE}};
+
+    void setRunOnHost(bool runOnHost);
+
+    bool runOnHost() const override;
```
MaticTonin, line 68: docs strings missing for exposed public API features, follow https://github.com/luxonis/depthai-core/blob/0cfd464dc30f32c45ac83f0dd3ba4d813131d952/include/depthai/pipeline/node/ImageAlign.hpp example

tests/src/ondevice_tests/pipeline/node/gate_note_tests.cpp
MaticTonin, line 1: camelCase in whole test

MaticTonin (changes requested):
Looks good to me, some small comments and we should not forget that we must as well update the FW variables in CMake.

include/depthai/pipeline/datatype/GateControl.hpp
```diff
@@ -0,0 +1,35 @@
...
+    bool value = true;
+
+    void start() {
+        value = true;
+    }
+
+    void stop() {
+        value = false;
+    }
+
```
moratom, line 24: nit: I'd go with open&close (to match kind of what "Gate" does)

include/depthai/pipeline/node/Gate.hpp
```diff
@@ -0,0 +1,39 @@
...
+    using DeviceNodeCRTP::DeviceNodeCRTP;
+
+    Input input{*this, {"input", DEFAULT_GROUP, false, 4, {{{DatatypeEnum::Buffer, true}}}, DEFAULT_WAIT_FOR_MESSAGE}};
+
+    Output output{*this, {"output", DEFAULT_GROUP, {{{DatatypeEnum::Buffer, true}}}}};
+
+    Input inputControl{*this, {"inputControl", DEFAULT_GROUP, false, 4, {{{DatatypeEnum::GateControl, false}}}, DEFAULT_WAIT_FOR_MESSAGE}};
+
+    void setRunOnHost(bool runOnHost);
+
```
moratom, line 63: maybe worth adding an initialControl, to have the initial state of the node configurable

include/depthai/properties/GateProperties.hpp
```diff
@@ -0,0 +1,19 @@
...
+#include "depthai/properties/Properties.hpp"
+
+namespace dai {
+
+/**
+ * Specify properties for Gate.
+ */
+
+struct GateProperties : PropertiesSerializable<Properties, GateProperties> {
+    int sleepingTimeMs = 50;  // sleep 50 ms between each inputControl reading
```
moratom: We should rather use `getAny` function from MessageGroup that was added for pipeline debugging here and we can then respond to any event.

include/depthai/pipeline/node/Gate.hpp
JakubFara: fixed here [02e9d9af2](https://github.com/luxonis/depthai-core/pull/1634/commits/02e9d9af2c28517eeccb0bef15d30c688756d78b) `Add initialConfig`

include/depthai/pipeline/datatype/GateControl.hpp
JakubFara: [f3d86e7b3](https://github.com/luxonis/depthai-core/pull/1634/commits/f3d86e7b35d14483cb910ad2d826dc479de5112a) Add test for N messages

include/depthai/properties/GateProperties.hpp
JakubFara: Now I am using
` auto inData = input.get();`
what is the difference between `get `and `getAny` ?

tests/src/ondevice_tests/pipeline/node/gate_note_tests.cpp
JakubFara: fixed here: [c3beccd2c](https://github.com/luxonis/depthai-core/pull/1634/commits/c3beccd2c5a30a454f3ab01234147e28f1126fe7) _Fix camelCase_

examples/python/Gate/gate_example.py
JakubFara: fixed here: [c3beccd2c](https://github.com/luxonis/depthai-core/pull/1634/commits/c3beccd2c5a30a454f3ab01234147e28f1126fe7) _Fix camelCase_

examples/cpp/Gate/gate_example.cpp
JakubFara: fixed here: [c3beccd2c](https://github.com/luxonis/depthai-core/pull/1634/commits/c3beccd2c5a30a454f3ab01234147e28f1126fe7) _Fix camelCase_

bindings/python/src/pipeline/datatype/GateControl.cpp
JakubFara: fixed here: [ef7363c8e](https://github.com/luxonis/depthai-core/pull/1634/commits/ef7363c8e7c2ef504998041305a3884d135d554b) _Remove diplicit bindings_

examples/python/Gate/gate_example.py
JakubFara: fixed here:[ 7c517d38](https://github.com/luxonis/depthai-core/pull/1634/commits/7c517d3818286953b4a6a17629eccef9829fcce5)

include/depthai/properties/GateProperties.hpp
moratom: I assume we need getAny between an `inputConfig` & `input` as you otherwise cannot close the gate before one message arrives.

Why is the sleepingTime needed?

bindings/python/src/pipeline/datatype/GateControlBindings.cpp
```diff
@@ -0,0 +1,36 @@
...
+    Callstack* callstack = (Callstack*)pCallstack;
+    auto cb = callstack->top();
+    callstack->pop();
+    cb(m, pCallstack);
+    ///////////////////////////////////////////////////////////////////////
+
+    gateControl.def(py::init<>())
+        .def(py::init<bool, int>(), "open"_a, "numMessages"_a)
+        // Member variables
+        .def_readwrite("open", &GateControl::open, DOC(dai, GateControl, open))
```
moratom, line 28: no `close`?

include/depthai/properties/GateProperties.hpp
moratom: To ilustrate my point, let's say I have an output that sends one message every 10 seconds.

Current behavior, if I'm understanding the code correctly is:
1. The `Gate` is open initally and passight through frames
2. 5 seconds after one of the messages was passed, someone sends a command to `Gate` to `close`
3. `Gate` waits for `get()` for 5 more seconds, still send the frame over and then close itself.

I suggest we do something like:
```
inputMessage = MessageQueue::getAny(inputControl, inputBuffer);
if inputMessage is inputControl -> change the behavior, settings respectively
if inputMessage is buffer -> send or don't send the message through (depending on whether the gate is closed or opened)
```

The behavir would then be:
1. The `Gate` is open initally and passight through frames
2. 5 seconds after one of the messages was passed, someone sends a command to `Gate` to `close`
3. `Gate` closes itself, no more messages are sent.

Steps 1. and 2. are the same as now.

tests/src/ondevice_tests/pipeline/node/gate_note_tests.cpp
```diff
@@ -0,0 +1,83 @@
...
+    // Test Parameters
+    const int num_cycles = 4;            // Number of times to toggle
+    const double period_duration = 2.0;  // 2 seconds per state
+    const double wait_delay = 0.05;      // 0.05s stabilization wait
+    bool current_gate_state = true;      // Start with Gate ON as requested
+
+    auto start_time = std::chrono::steady_clock::now();
+    int cycle_count = 0;
+    int frames_in_period = 0;
+    bool is_measuring = false;
```
moratom: camelCase

bindings/python/src/pipeline/datatype/GateControlBindings.cpp
JakubFara: `open` is a `bool` value. `open=false` is close

tests/src/ondevice_tests/pipeline/node/gate_note_tests.cpp
JakubFara: [c3beccd2c](https://github.com/luxonis/depthai-core/pull/1634/commits/c3beccd2c5a30a454f3ab01234147e28f1126fe7) _Fix camelCase_

include/depthai/properties/GateProperties.hpp
JakubFara: Your solution has the disadvantage that when the gate is closed, messages still keep coming in, and you need to use a dynamic_cast to distinguish between GateControl and ImgFrame.

what about this: 
```
 std::shared_ptr<GateControl> Gate::sendMessages() {
     while(true) {
+        auto inData = input.get();
         auto control = inputControl.tryGet<GateControl>();
         if(control) {
             return control;
         }
-        auto inData = input.get();
         output.send(inData);
     }
 }
```

JakubFara: sleepingTime has been removed

moratom: IIRC it's not dynamic_cast that would be used, you pass the two queues in with a `std::map<std::string, MessageQueue>`.

Very good point on still getting messages though!
How about `getAny` when the gate is open but `get` on the input control when it's closed?

I think more robust than above (as in theory you could get two input controls in the mean time between `gets`, etc). This is very much an edge edge case case, but still.

JakubFara: Also 1 disadvantage is that you need to have a non-blocking queue. So it might happen that your control will be removed if the queues are merged into 1.

moratom: It's not one queue, you keep two inputs, it's only the implementation that changes:
https://github.com/luxonis/depthai-core/blob/fa1ae8626d3720c7e76f3aa62eb76d2cd7f5c8e3/include/depthai/pipeline/MessageQueue.hpp#L92

JakubFara: Is is possible to do a similar thing on RVC2?

asahtik: `getAny` is not implemented on RVC2 (yet?)

JakubFara: I have implemented it with `getAny` here (@moratom please take a look if it is correct):

- [Add reference to unordered map in getAny](https://github.com/luxonis/depthai-core/pull/1634/commits/8df5fab917adee0859a5a1bc3bb7e7d05b746a55)
[8df5fab](https://github.com/luxonis/depthai-core/pull/1634/commits/8df5fab917adee0859a5a1bc3bb7e7d05b746a55)

- [Use getAny](https://github.com/luxonis/depthai-core/pull/1634/commits/c6469d8ac025e4ec0aba170f44ab6bca750eaff9)
[c6469d8](https://github.com/luxonis/depthai-core/pull/1634/commits/c6469d8ac025e4ec0aba170f44ab6bca750eaff9)

moratom: Good question.
CC: @asahtik on this, should be possible.

If PITA I'm fine with going with your approach though.

asahtik: There is something that can maybe be used to get similar functionality though (haven't tried it yet).
You could set a `SemSync` struct to the receivers you want to wait on:
```
input1.sync = &semSync
input2.sync = &semSync
```
`semSync` should get notified when a new msg arrives and you can wait on that with `semSync.Wait()`.

`PlgNeuralNetwork` sort of does this but it sets the semSync when the inputs are being created (maybe you could do the same?):
```
    tensorInputSync.Create();
    if(mainInput.blocking){
        tensorInput.Create(..., &tensorInputSync);
    } else {
        tensorInput.Create(..., &tensorInputSync);
    }
    ...
                if(io.blocking) in->input.Create(..., &tensorInputSync);
                else in->input.Create(..., &tensorInputSync);
    ...
            // Wait till any inputs arrived
            if(tensorInputSync.Wait() == 0){
    ...
```

JakubFara: thanks @asahtik is seems that SemSync works nicely!

That comes with a qustion - would not be better to have also `waitAny` on RVC4?
Than I can do

```
std::shared_ptr<GateControl> Gate::sendMessages(int numMessages) {
    for(int i = 0; i < numMessages; i++) {
        if (MessageQueue::waitAny(inputs))
            if(auto inControl = inputControl.tryGet<GateControl>()) {
                    return inControl;
            }
            if(auto in = input.tryGet()) {
                output.send(in);
            }
    }
    return std::make_shared<GateControl>(false, -1);
}
```
Advantages
- I don't need to create unordered_map on the heap.
- I can firstly process the` inputControl` without getting the `input`

JakubFara: Implemented here:
- [Add waitAny](https://github.com/luxonis/depthai-core/pull/1634/commits/b529fcf2580341f7bb8770d852f575659c2f44f7)
[b529fcf](https://github.com/luxonis/depthai-core/pull/1634/commits/b529fcf2580341f7bb8770d852f575659c2f44f7)

- [Change getAny to waitAny](https://github.com/luxonis/depthai-core/pull/1634/commits/93cd33b8424eb605e6c669931e5e8d93528a26fe)
[93cd33b](https://github.com/luxonis/depthai-core/pull/1634/commits/93cd33b8424eb605e6c669931e5e8d93528a26fe)

I am most satisfied with this solution.

bindings/python/src/pipeline/node/GateBindings.cpp
```diff
@@ -0,0 +1,41 @@
...
+    Callstack* callstack = (Callstack*)pCallstack;
+    auto cb = callstack->top();
+    callstack->pop();
+    cb(m, pCallstack);
+    ///////////////////////////////////////////////////////////////////////
+
+    gate.def_readonly("input", &Gate::input, DOC(dai, node, Gate, input))
+
+        .def_property(
+            "initialConfig",
```
majvan, line 26: It's called initialConfig, but you can change it. So is it only config at init, or what is the property?

JakubFara: Every deviceNode has it's property, which is passed to  the devcie counterpart when the node is initialized, so you can set initial values here. The initialConfig is of type GateControl -> the "initial state" of the gate

```
class GatePropety : Property {
    GateControl initialConfig;
}
```

A bit confusing agree.

src/pipeline/MessageQueue.cpp
```diff
@@ -215,7 +215,42 @@ void MessageQueue::notifyCondVars() {
```
majvan, line 215: I would move these pipeline changes to another commit, out of GateNode

src/pipeline/MessageQueue.cpp
```diff
@@ -215,7 +215,42 @@ void MessageQueue::notifyCondVars() {
...
+    // Store IDs so we can unsubscribe later
+    std::vector<std::pair<MessageQueue&, MessageQueue::CallbackId>> ids;
+
+    // 1. Subscribe to all queues in the vector
+    for(MessageQueue& queue : queues) {
+        ids.push_back({queue, queue.addCondVar(inputsWaitCv)});
+    }
+
+    // 2. Wait until any queue has data or closes
+    std::unique_lock<std::mutex> lock(waitMutex);
```
majvan, line 231: I don't understand the lock / mutex here. Lock / mutex requires some shared 'component' but you don't have any.

moratom: (likely transfered over from `getAny`, which should also have this addressed, cc: @asahtik).
Great catch @majvan!

asahtik: You need a lock instance for the `inputWaitCv->wait()` call (one line below this one)

majvan: But the lock instance cannot be a local one (created on a stack), correct?

asahtik: Ah I think I see your point. I was working under the assumption that notify cannot be called while the predicate is being checked, which is apparently not the case.

include/depthai/pipeline/node/Gate.hpp
JakubFara: [6a345a67d Add docstrings](https://github.com/luxonis/depthai-core/pull/1634/commits/6a345a67d43e7e0978749a07182d45a739333ba4)

MaticTonin: Adressed thanks

PR #1639, opened 2026-01-16
Improve image align

moratom:
## Purpose
Updates ImageAlign on all platforms (RVC2, RVC4, Host) to:
* use ImgTransformations for intrinsics, distortion
* apply undistortion for the input if necessary
* correctly track intrinsics
* handle custom depth sizes 

## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

Tests extended.

### DIFF
--- a/bindings/python/src/pipeline/datatype/ImgFrameBindings.cpp
+++ b/bindings/python/src/pipeline/datatype/ImgFrameBindings.cpp
@@ -175,6 +175,7 @@ void bind_imgframe(pybind11::module& m, void* pCallstack) {
         .def("remapPointFrom", &ImgTransformation::remapPointFrom, py::arg("to"), py::arg("point"), DOC(dai, ImgTransformation, remapPointFrom))
         .def("remapRectTo", &ImgTransformation::remapRectTo, py::arg("to"), py::arg("rect"), DOC(dai, ImgTransformation, remapRectTo))
         .def("remapRectFrom", &ImgTransformation::remapRectFrom, py::arg("to"), py::arg("rect"), DOC(dai, ImgTransformation, remapRectFrom))
+        .def("isAlignedTo", &ImgTransformation::isAlignedTo, py::arg("to"), DOC(dai, ImgTransformation, isAlignedTo))
         .def("isValid", &ImgTransformation::isValid, DOC(dai, ImgTransformation, isValid));
 
     // TODO add RawImgFrame::CameraSettings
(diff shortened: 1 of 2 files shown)


examples/cpp/RVC2/ImageAlign/image_align.cpp
```diff
@@ -50,7 +50,7 @@ int main() {
     sync->setSyncThreshold(std::chrono::milliseconds(static_cast<int>((1 / FPS) * 1000.0 * 0.5)));
 
     align->outputAligned.link(sync->inputs["aligned"]);
-    auto* RGBOutput = camRgb->requestOutput(std::make_pair(640, 400));
+    auto* RGBOutput = camRgb->requestOutput(std::make_pair(640, 400), std::nullopt, dai::ImgResizeMode::CROP, std::nullopt, true);
```
aljazkonec1, line 53: Could we add an overload similar to the new `NN .build()` overloads where it takes in a `dai::ImgFrameCapability` parameter? it would future proof the function and could use the same capability object across multiple camera streams (eg depth - rgb alignment) with ease.

include/depthai/common/ImgTransformations.hpp
```diff
@@ -240,6 +240,12 @@ struct ImgTransformation {
      */
     dai::RotatedRect remapRectFrom(const ImgTransformation& from, dai::RotatedRect rect) const;
 
+    /**
+     * Check if the transformations are aligned
+     * @param to Transformation to compare with
+     */
+    bool isAlignedTo(const ImgTransformation& to) const;
```
aljazkonec1, line 247: I couldn't find pybind in ImgFrameBindings.cpp. Is it somewhere else?

tests/src/ondevice_tests/image_align_node_test.cpp
```diff
@@ -51,23 +62,31 @@ void runImageAlignTest(bool useDepth, bool runOnHost) {
...
 }
 
 TEST_CASE("Test ImageAlign node depth to image alignment on host") {
     bool useDepth = true;
     bool runOnHost = true;
-    runImageAlignTest(useDepth, runOnHost);
+    for(const auto resizeMode : {dai::ImgResizeMode::CROP, dai::ImgResizeMode::LETTERBOX, dai::ImgResizeMode::STRETCH}) {
+        runImageAlignTest(useDepth, runOnHost, resizeMode);
+    }
 }
```
aljazkonec1, line 92: In the future we should add tests if one frame is rotated

examples/cpp/RVC2/ImageAlign/image_align.cpp
moratom: Already exists, it's a bit annoying to use since you can't do a single liner then.

include/depthai/common/ImgTransformations.hpp
moratom: Added

PR #1644, opened 2026-01-21
Fix getAny race condition

asahtik:
Fixes race condition in `getAny` by implementing `BinarySemaphore` (should be replaced with `std::binary_semaphore` when we switch to C++20) and `ManyToOneNotifier`. Also adds `waitAny`.

Might need some extra eyes on this :eyes: 

Here's why I believe this works:

```
func waitAny():
  1. add notifiers
  // If we get a message here the next line will evaluate to true + notifier semaphore will have available set
  2. check if any messages are already present
  // If we get a message here it will be caught in the notifier predicate + notifier semaphore will have available set
  3. check predicate
  // If we get a message here the notifier semaphore will have available set
  4. acquire semaphore
  // From here on the semaphore should ensure that the thread continues when available is set
  5. go to 3. which should now return true
```

```
// This is implemented according to the reference so it should be correct
func Sem::acquire:
  6. lock the available variable
  // If there were any new messages from point 3 to 6, the next line will evaluate to true
  7. check predicate (is available true - in cv wait)
  8. unlock
  9. sleep & wait for notify
```

## Summary by CodeRabbit

* **New Features**
  * Optional timeout support for multi-queue wait operations.
  * New lightweight semaphore and notifier helpers to coordinate wait/notify across queues.
  * Python bindings for multi-queue wait/get APIs.

* **Improvements**
  * More robust thread-safe queue state checks and more reliable multi-queue retrieval/wakeup semantics.

* **API Changes**
  * Listener registration/removal interfaces renamed for clarity.

* **Tests**
  * Added tests covering multi-queue wait/get and timeout scenarios.


include/depthai/utility/ManyToOneNotifier.hpp
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    /**
+     * @brief Wait until notified
+     * @arg pred Predicate to check after each notification - IMPORTANT: must be thread-safe
+     */
+    template <typename Pred>
+    void wait(Pred pred) {
+        if(waiting.exchange(true)) {
+            throw std::runtime_error("ManyToOneNotifier: Multiple threads waiting on the same notifier is not supported");
```
majvan: What is the purpose of the condition that only one thread can wait for a notifier; what is that beneficial for?
In the message queues, you typically want that many threads wait in a notification queue and only one thread gets notified.

asahtik: The purpose of this class is for one main thread to wait on any one of multiple worker threads. This check ensures that someone using this class doesn't get unexpected results if they try to wait from multiple threads.

include/depthai/utility/BinarySemaphore.hpp
```diff
@@ -0,0 +1,65 @@
...
+    BinarySemaphore() = default;
+    explicit BinarySemaphore(bool initiallyAvailable) : available(initiallyAvailable) {}
+
+    // signal / post / V
+    void release() {
+        {
+            std::lock_guard<std::mutex> lk(mtx);
+            available = true;
+        }
+        cv.notify_one();
```
majvan: If we talk about BinarySemaphore (only one waiting), then you can notify conditionally.
If a semaphore was already released and anybody would call "release", do not notify. Take it as an erroneous situation with no action.

include/depthai/utility/ManyToOneNotifier.hpp
majvan: Is that a beneficial behavior?
If you did not use a BinarySemaphore but used a class cv + mutex, then there could be more waiting threads waiting while Worker threads could be notifying them.

asahtik: If I used a mutex I'd have to lock the `queue.push` method and no output would be able to send while another watched queue is being sent to or while checking the predicate. I'd also have to add a lock for each waiting thread which would increase code complexity, which is not necessary since this accomplishes the goal of this PR.

include/depthai/utility/BinarySemaphore.hpp
asahtik: If there's no waiting thread then there's practically no overhead to notifying, which would be the case if `available` was already true. If I conditionally notified, I'd have to do it under the lock which would mean that the waiting thread would wake up and then immediately block and wait for the mutex to be released which would cause two extra context switches.

majvan: Not sure you understood me. Let's make it clear.

This is one scenario I was thinking about:
```
    void release() {
        ASSERT(available == true, "There is a bug in the SW, a thread calls release while the semaphore is not acquired. We will not call notify_one");
        {
            std::lock_guard<std::mutex> lk(mtx);
            available = true;
        }
        cv.notify_one();
    }

```    
 Another one, without ASSERT but with 
 ```
    void release() {
        bool was_available = false;
        {
            std::lock_guard<std::mutex> lk(mtx);
            was_available = available;
            available = true;
        }
        if (!was_available)
            cv.notify_one();  // call notify one only in an expected scenario.
    }
```

File: include/depthai/utility/ManyToOneNotifier.hpp (hunk 1 above)
Comment (majvan) [reply include/depthai/utility/ManyToOneNotifier.hpp]: I assume we have the following scenario - and correct me, if I am wrong.
1. Multiple writers. One writer doesn't wait for a signal, it tries to push to a particular queue. But there may be multiple writers trying to push data to the same queue.
2. Multiple readers on the particular queue. A reader is waiting to get a signal that the queue has any data available - and it reads the data and processes further.

What about this solution then:

1. A worker thread (writer) that pushes data to a queue
1.1. The thread locks the queue with queue's related queueMutex
1.2. Data is pushed into queue
1.3. The queue is unlocked with queue's related queueMutex
1.4. A conditional variable dataReady is called to notify max. one thread

2. A worker thread (reader) that waits for a data from a queue:
2.1 The queue locks the queue with queue's related queueMutex
2.2 Conditional variable dataReady->wait(queueMutex...) is called:
2.2.1 Checks the predicate and if true, return
2.2.2 **Unlocks** the queueMutex
2.2.3 **Waits** for a signal from dataReady (with unlocked queueMutex)
2.2.4 **Locks** the queueMutex and goes to 2.2.1
2.2 The worker thread reads the data from the queue
2.3 The worker thread unlocks the queueMutex

File: include/depthai/utility/BinarySemaphore.hpp (hunk 2 above)
Comment (asahtik) [reply include/depthai/utility/BinarySemaphore.hpp]: Ah I see. `notify_one` has little overhead when there are no waiters but it certainly doesn't hurt to add this. Thanks

File: include/depthai/utility/ManyToOneNotifier.hpp (hunk 1 above)
Comment (asahtik) [reply include/depthai/utility/ManyToOneNotifier.hpp]: If I understand correctly, you're describing the way `queue.get()` works. This is a separate thing. Each message queue already has its own conditional variable and mutex that handles pushes and pops from the queue.

The issue I had for getAny is that there isn't a mechanism built in to only wait until one conditional variable notifies. To do this you need to pass a single condition variable instance to multiple queues and wait until one notifies.

So the way it works now is that each time you want to wait on any of a list of queues, a semaphore is created and added to every queue in the list. When a new message arrives to any of the queues (or when a queue is destroyed) the semaphore is notified, and the waiting thread gets woken up.

File: include/depthai/utility/WaitAnyNotifier.hpp [hunk 3]
```diff
@@ -0,0 +1,73 @@
...
+     * @brief Wait until notified
+     * @arg pred Predicate to check after each notification - IMPORTANT: must be thread-safe
+     */
+    template <typename Pred>
+    void wait(Pred pred) {
+        if(waiting.exchange(true)) {
+            throw std::runtime_error("ManyToOneNotifier: Multiple threads waiting on the same notifier is not supported");
+        }
+        try {
+            while(!pred()) semaphore.acquire();
```
majvan, line 46: Two issues.

[1]
We check pred() and then (maybe) acquire a semaphore
This function has then two expected outcomes then:
- the pred was fulfilled BUT NO semaphore acquired (the case when the pred() returns true the first time)
- the pred was fulfilled AND semaphore acquired (the case when the pred() returns true the next time)
The semaphore acquisition is not causally linked to the predicate becoming true.



[2]
Typically the Pred in the wait (like cv::wait) are used just because the whole pred() is running protected under the mutex. The reason is that the mutex ensures the atomicity of the operation together with cv's waiting cycle.

It's not a case here. If pred() is an unprotected function, it can be preempted by other thread which can access the state of the pred() - and this creates a potential RACE CONDITION.


Example:
// We need to ensure that the queue has data and that camera is on (an abstract example)
bool pred() { 
    if (!camera_on)
        return false;
    if (queue.empty)
        return false;
    return true;
}

Step 0 - initial state:
   camera_on = true; 
   queue.empty = true; 

Step 1 - ThreadA acquires a semaphore and continues:
gets into pred() and computes camera_on = true
-- preemptive context switch --

Step 2 - Thread C runs:
{
   lock(camera_and_queue);
   camera_on = false;  // just set that the camera is off
}
-- preemptive context switch --

Step 3 - ThreadA continues:
continues the pred() and computes queue.empty = false and returns true
ManyToWaitNotifier::wait() returns but the state is:
camera_on = false;  <-- RACE CONDITION: we wanted to wait only till camera is on

The predicate reads shared state (camera_on, queue.empty) without holding the mutex that protects it. This allows concurrent modification by other threads, resulting in undefined behavior and inconsistent observations of the system state.
In contrast, std::condition_variable::wait(lock, pred) guarantees that the predicate is evaluated while holding the mutex, making the check and sleep cycle atomic.



Another outcome of [2] is a possible deadlock:

Step 1: ThreadA - runs:
if (!pred())
    semaphore.acquire(); // waits in the semaphore for the cv

Step 2: ThreadB - is about to release the semaphore:
{
   std::lock(camera_and_queue)
   queue.empty = false; 
   notifyOne(); -> semaphore.release(); // notify there is next some data, release the semaphore
}

Step 3: ThreadA - continues:
 semaphore.acquire(); // finishes the semaphore acquiring
-- preemptive context switch --

Step 4: ThreadC - runs:
{
   lock(camera_and_queue);
   camera_on = false;  // just set that the camera is off
}
-- preemptive context switch --

Step 5 - ThreadA continues:
if (!pred())
   semaphore.acquire(); // DEADLOCK <-- semaphore count is zero and no further release will occur;  ThreadA will get not awaken anymore

The semaphore release is not tied to the predicate remaining true.
A thread may wake, consume the semaphore, and then observe the predicate as false due to a concurrent state change. Since the semaphore signal has already been consumed and no further releases are guaranteed, the waiting thread may block forever.

asahtik: The second issue is documented:
```
Should always be called after the state change that the waiting thread is waiting for
```
(maybe the wording is not clear, should it be 'after any state change'?)

ThreadC should notify after setting camera_on to false.

The first issue is a good point. I'll document this limitation.

majvan: This constraint works for the deadlock, but these rules have to be followed.
And it brings additional unnecessary wakeups even in the case your state is going "against" the wakeup (why would you wakeup threads when you switch off the camera when you want them to work when the camera is switched on?)

Using std::condition_variable is much easier and exactly what you need here without reinventing it. :)

src/pipeline/MessageQueue.cpp
```diff
@@ -203,117 +207,79 @@ void MessageQueue::callCallbacks(std::shared_ptr<ADatatype> message) {
...
-std::unordered_map<std::string, std::shared_ptr<ADatatype>> MessageQueue::getAny(const std::unordered_map<std::string, MessageQueue&>& queues,
-                                                                                 std::optional<std::chrono::milliseconds> timeout) {
-    std::vector<std::pair<MessageQueue&, MessageQueue::CallbackId>> condVarIds;
-    condVarIds.reserve(queues.size());
-    std::unordered_map<std::string, std::shared_ptr<ADatatype>> inputs;
-
-    std::mutex inputsWaitMutex;
-    auto inputsWaitCv = std::make_shared<std::condition_variable>();
+    };
+    auto checkAllClosed = [&]() { return boost::algorithm::all_of(queues, [](const std::reference_wrapper<MessageQueue> q) { return q.get().isClosed(); }); };
```
moratom: no boost please!

src/pipeline/MessageQueue.cpp
```diff
@@ -203,117 +205,81 @@ void MessageQueue::callCallbacks(std::shared_ptr<ADatatype> message) {
...
+    std::vector<std::reference_wrapper<MessageQueue>> queuesVec;
+    queuesVec.reserve(queues.size());
+    for(const auto& kv : queues) {
+        queuesVec.emplace_back(kv.second);
+    }
+    bool gotAny = waitAny(queuesVec, timeout);
+    if(gotAny) {
+        for(const auto& kv : queues) {
+            auto& input = kv.second;
+            if(!input.isClosed() && input.has()) inputs[kv.first] = input.get<ADatatype>();
```
majvan: Shouldn't be this atomically checked and retrieved?
Get an input only when it is not closed and has data. Any modification and readout of the queue state should be mutex-ed.

moratom:
@coderabbit review

include/depthai/pipeline/MessageQueue.hpp
```diff
@@ -83,11 +86,13 @@ class MessageQueue : public std::enable_shared_from_this<MessageQueue> {
         queue = std::move(m.queue);
         name = std::move(m.name);
         callbacks = std::move(m.callbacks);
+        notifiers = std::move(m.notifiers);
         uniqueCallbackId = m.uniqueCallbackId;
         pipelineEventDispatcher = m.pipelineEventDispatcher;
         return *this;
```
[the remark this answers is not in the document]
moratom: Good point on first glance @asahtik

moratom (approved):
After a long review, I think this is good for our usecase.

Short checklist before merging:
* Make sure it's formatted before merging, there seem to be some formatting changes
* Python bindings
* Add docstrings to getAny & waitAny

PR #1657, opened 2026-01-28
Cached snaps and events are now gradually uploaded during runtime, mi…

aljazdu:
…nor fixes

## Purpose
Cached snaps could previously be uploaded only when restarting the application. Now, in case there is some cached data available, they will be gradually uploaded runtime, while making sure that the limits imposed by the Hub are not hit.

## Specification
Updated EventsManager class.

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation
This was tested by sending 30 snaps per second, each containing an Image and Detections pair. A connection was dropped on purpose, and 2000 snaps were cached. Connection was reestablished and the cached snaps were slowly uploaded side by side with the newly sent snaps.

### DIFF
--- a/src/utility/EventsManager.cpp
+++ b/src/utility/EventsManager.cpp
@@ -16,6 +16,41 @@ std::string generateLocalID(int64_t timestamp) {
     oss << timestamp << "_" << std::setw(6) << std::setfill('0') << counter++;
     return oss.str();
 }
+
+const std::unordered_map<std::string, std::string> extensionToMimeTypeMap = {
+    {".html", "text/html"},
+    {".css", "text/css"},
+    {".js", "application/javascript"},
+    {".png", "image/png"},
+    {".jpg", "image/jpeg"},
+    {".jpeg", "image/jpeg"},
+    {".gif", "image/gif"},
+    {".webp", "image/webp"},
+    {".bmp", "image/bmp"},
+    {".tiff", "image/tiff"},
+    {".svg", "image/svg+xml"},
+    {".json", "application/json"},
+    {".txt", "text/plain"},
+    {".annotation", "application/x-protobuf; proto=SnapAnnotation"},
+    {"", "application/octet-stream"},
+};
+
+const std::unordered_map<std::string, std::string> mimeTypeToExtensionMap = {
+    {"text/html", ".html"},
+    {"text/css", ".css"},
+    {"application/javascript", ".js"},
+    {"image/png", ".png"},
+    {"image/jpeg", ".jpg"},
+    {"image/gif", ".gif"},
+    {"image/webp", ".webp"},
+    {"image/bmp", ".bmp"},
+    {"image/tiff", ".tiff"},
+    {"image/svg+xml", ".svg"},
+    {"application/json", ".json"},
+    {"text/plain", ".txt"},
+    {"application/x-protobuf; proto=SnapAnnotation", ".annotation"},
+    {"application/octet-stream", ""},
+};
 }  // namespace
 
 #include "Environment.hpp"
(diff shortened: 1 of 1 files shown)


src/utility/EventsManager.cpp
```diff
@@ -217,7 +220,12 @@ bool FileData::toFile(const std::filesystem::path& inputPath) {
         logger::error("Filename is empty");
         return false;
     }
-    std::string extension = mimeType == "image/jpeg" ? ".jpg" : ".txt";
+    std::string extension;
+    if(mimeType == "image/jpeg") {
+        extension = ".jpg";
+    } else if(mimeType == "application/x-protobuf; proto=SnapAnnotation") {
+        extension = ".annotation";
+    }
```
lnotspotl, line 250: the extension will be an empty string if both the above conditions are false, expected? It was `.txt` before.

include/depthai/utility/EventsManager.hpp
```diff
@@ -193,9 +193,10 @@ class EventsManager {
 
     /**
      * Fetch the configuration limits and quotas for snaps & events
+     * @param retryOnFail Retry fetching on failure
```
lnotspotl: Can we add that this will keep on trying, even multiple times, until successful?

src/utility/EventsManager.cpp
```diff
@@ -893,8 +905,11 @@ void EventsManager::cacheEvents() {
     logger::info("Caching events");
     std::lock_guard<std::mutex> lock(eventBufferMutex);
     for(const auto& eventData : eventBuffer) {
-        std::filesystem::path path(cacheDir);
-        path = path / ("event_" + eventData->event->name() + "_" + std::to_string(eventData->event->created_at()));
+        std::filesystem::path inputPath(cacheDir);
+        std::filesystem::path path = inputPath / ("event_" + eventData->event->name() + "_" + std::to_string(eventData->event->created_at()));
+        for(int i = 1; std::filesystem::exists(path); ++i) {
+            path = inputPath / ("event_" + eventData->event->name() + "_" + std::to_string(eventData->event->created_at()) + "_" + std::to_string(i));
```
lnotspotl: Just a suggestion, noop 
```suggestion
            path = inputPath / (fmt::format("event_{}_{}_{}", eventData->event->name(), eventData->event->created_at(), i);
```

src/utility/EventsManager.cpp
aljazdu: Added missing mime types and their extensions.

include/depthai/utility/EventsManager.hpp
aljazdu: Done.

src/utility/EventsManager.cpp
aljazdu: Agreed, added your suggestion.

src/utility/EventsManager.cpp
```diff
@@ -221,10 +221,24 @@ bool FileData::toFile(const std::filesystem::path& inputPath) {
...
+        {"image/webp", ".webp"},
+        {"image/bmp", ".bmp"},
+        {"image/tiff", ".tiff"},
+        {"image/gif", ".gif"},
+        {"image/svg+xml", ".svg"},
+        {"application/json", ".json"},
+        {"text/plain", ".txt"},
+        {"text/html", ".html"},
+        {"text/css", ".css"},
+        {"application/javascript", ".js"},
```
lnotspotl: What about `application/octet-stream`, set on line `140`, [here](https://github.com/luxonis/depthai-core/blob/cec417c4c7311e25f77d36d5abd51fbb5b33294a/src/utility/EventsManager.cpp#L140)

Can we maybe move the mapping (mimo -> extension and extension -> mimo) outside the two methods that use it, somewhere to the top of this file, such that it's easier to verify that they contain the same data and no combination was forgotten? The mimo type `application/octet-stream` seems to be missing too.

aljazdu: Agreed regarding moving it to the top. Fixed in the update.

Added missing `application/octet-stream` entries in the maps - to which we default in case of unsupported mime types.

aljazdu:
> LGTM, thanks!

Thanks for the review!

PR #1661, opened 2026-01-29
Add SegmentationMask message and SegmentationParser

aljazkonec1:
## Purpose

This PR adds a SegmentationMask message and a new SegmentationParser node to the core library. This will allow for easier and faster segmentation pipelines and allow for updating other nodes (such as ObjectTracker) to use object segmentation instead of a bounding box.

## Specification

**- SegmentationMask message** is a general object that stores segmentation data in an uint8 array with the value 255 being reserved for the background. The message does not differentiate between instance and semantic segmentation masks and it is up to the user to properly handle the differences. For easier use, multiple helper functions and opencv functions are also added.
**- SegmentationParser** is the asociated parser ported over from [depthai-nodes](https://github.com/luxonis/depthai-nodes/blob/main/depthai_nodes/node/parsers/segmentation.py). The parser supports the current two types of segmentation models we have on HubAI. First type are models that return a tenosr of size H x W x C, where the C dimension represents the probability values for class C. For these models, the parser performs an (optionally thresholded) Argmax operation across C to get the segmentation mask. And the second type of models already have the Argmax operation baked into the NN and thus the parser only copies the NN outputs to the segmentation mask message. The parser supports the following HubAI models (*some nnarchives need to be updated as they are currently incorrect):
- luxonis/deeplab-v3-plus:512x288*
- luxonis/deeplab-v3-plus:512x512
- luxonis/deeplab-v3-plus:256x256
- luxonis/deeplab-v3-plus:person-256x256
- luxonis/yolo-p:bdd100k-320x320*
- luxonis/pp-liteseg:1024x512
- luxonis/mediapipe-selfie-segmentation:256x144*
- luxonis/ewasr:512x384*

Bindings, tests and docstrings are also added.

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

Tested the above liste models, added test.

### DIFF
--- a/utilities/device_manager.py
+++ b/utilities/device_manager.py
@@ -197,7 +197,7 @@ class SearchDevice:
         else:
             rows = []
             for info in self.infos:
-                if "X_LINK_GATE" == info.state.name: continue # Skip RVC4 devices
+                if "GATE" in info.state.name: continue # Skip RVC4 devices
                 rows.append([info.getDeviceId(), info.name, deviceStateTxt(info.state)])
             self.window['table'].update(values=rows)
 
@@ -270,20 +270,37 @@ def factoryReset(device: dai.DeviceInfo, type: dai.DeviceBootloader.Type):
     try:
         pr = Progress('Preparing and connecting...')
 
-        blBinary = dai.DeviceBootloader.getEmbeddedBootloaderBinary(type)
-        # Clear 1 MiB for USB BL and 8 MiB for NETWORK BL
-        mib = 1 if type == dai.DeviceBootloader.Type.USB else 8
-        blBinary = blBinary + ([0xFF] * ((mib * 1024 * 1024 + 512) - len(blBinary)))
-        tmpBlFw = tempfile.NamedTemporaryFile(delete=False)
-        tmpBlFw.write(bytes(blBinary))
-
         bl = dai.DeviceBootloader(device, True)
-
         progress = lambda p : pr.update(p)
-        success, msg = bl.flashBootloader(progress, tmpBlFw.name)
+        if type == dai.DeviceBootloader.Type.AUTO:
+            type = bl.getType()
+
+        tmpBlFw = None
+        try:
+            blBinary = dai.DeviceBootloader.getEmbeddedBootloaderBinary(type)
+            # Clear 1 MiB for USB BL and 8 MiB for NETWORK BL
+            mib = 1 if type == dai.DeviceBootloader.Type.USB else 8
+            blBinary = blBinary + ([0xFF] * ((mib * 1024 * 1024 + 512) - len(blBinary)))
+            tmpBlFw = tempfile.NamedTemporaryFile(delete=False)
+            tmpBlFw.write(bytes(blBinary))
+            tmpBlFw.flush()
+            success, msg = bl.flashBootloader(progress, tmpBlFw.name)
+        except Exception as ex:
+            # In some depthai Python builds vector<uint8_t> return conversion is unavailable.
+            if "Unable to convert function return value to a Python type" in str(ex):
+                success, msg = bl.flashBootloader(
+                    memory=dai.DeviceBootloader.Memory.FLASH,
+                    type=type,
+                    progressCallback=progress
+                )
+            else:
+                raise
+        finally:
+            if tmpBlFw is not None:
+                tmpBlFw.close()
+
         msg = "Factory reset was successful." if success else f"Factory reset failed. Error: {msg}"
         pr.finish(msg)
-        tmpBlFw.close()
     except Exception as ex:
         PrintException()
         sg.Popup(f'{ex}')
@@ -864,7 +881,8 @@ class DeviceManager:
             else:
                 for deviceInfo in deviceInfos:
                     deviceTxt = deviceInfo.getDeviceId()
-                    if "X_LINK_GATE" == deviceInfo.state.name: continue # Skip RVC4 devices
+                    print(deviceInfo.state.name)
+                    if "GATE" in deviceInfo.state.name: continue # Skip RVC4 devices
                     listedDevices.append(deviceTxt)
                     self.devices[deviceTxt] = deviceInfo
 


include/depthai/pipeline/datatype/SegmentationMask.hpp
```diff
@@ -0,0 +1,219 @@
...
+
+/**
+ * SegmentationMask message.
+ *
+ * Segmentation mask of an image is stored as a single-channel INT8 array, where each value represents a class or instance index.
+ * The value 255 is treated as background pixels (no class/instance).
+ */
+class SegmentationMask : public Buffer, public ProtoSerializable {
+    // Optimization option: if network is bottleneck, implement RLE compression for the mask data
+   protected:
```
aljazdu: Do we expect sub-classing of `SegmentationMask` in the future or is the only reason for this the hidden implementation accessed using friend? 
In case the only reason is friend access, then this can be private AFAIK

include/depthai/pipeline/datatype/SegmentationMask.hpp
```diff
@@ -0,0 +1,219 @@
...
+    friend struct priv::SegmentationMaskAccess;
+    void setSizeInternal(size_t width, size_t height);
+
+   public:
+    using Buffer::getSequenceNum;
+    using Buffer::getTimestamp;
+    using Buffer::getTimestampDevice;
+    using Buffer::setSequenceNum;
+    using Buffer::setTimestamp;
+    using Buffer::setTimestampDevice;
```
aljazdu: Any particular reason for these `using` declarations? `SegmentationMask` already inherits `public` and `protected` members of `Buffer` (including the members defined here).

include/depthai/pipeline/datatype/SegmentationMask.hpp
```diff
@@ -0,0 +1,219 @@
...
+    std::optional<ImgTransformation> transformation;
+
+    SegmentationMask();
+    SegmentationMask(const std::vector<std::uint8_t>& data, size_t width, size_t height);
```
aljazdu, line 38: Nitpick. Const width and height. Some other places as well.

src/pipeline/datatype/SegmentationMask.cpp
```diff
@@ -0,0 +1,372 @@
...
+    this->width = width;
+    this->height = height;
+    return data->getData();
+}
+
+void SegmentationMask::setMask(dai::ImgFrame& frame) {
+    if(frame.getType() != dai::ImgFrame::Type::GRAY8) {
+        throw std::runtime_error("SegmentationMask: ImgFrame type must be GRAY8");
+    }
+    auto dataSpan = frame.getData();
```
aljazdu: Could existing `setMask()` overload be called here instead? This would make the intermediate vector obsolete and result in less copying.

src/pipeline/datatype/SegmentationMask.cpp
```diff
@@ -0,0 +1,372 @@
...
+}
+
+void SegmentationMask::setMask(span<const std::uint8_t> mask, size_t width, size_t height) {
+    if(mask.size() != width * height) {
+        throw std::runtime_error("SegmentationMask: data size does not match width*height");
+    }
+    data->setSize(mask.size());
+    std::memcpy(data->getData().data(), mask.data(), mask.size());
+    this->width = width;
+    this->height = height;
```
aljazdu, line 57: This might be totally wrong - ignore it in this case - but would setting the timestamps here and in other `setMask()` overloads be a good idea? 
In the case when using the constructor that already uses setMask that would not make much of a difference, but what about if using a constructor without data and only later setting the mask? In this case the timestamps are never set.

include/depthai/pipeline/datatype/SegmentationMask.hpp
```diff
@@ -0,0 +1,219 @@
...
+    /**
+     * Returns all class labels associated with the segmentation mask. If no labels are set, returns an empty vector.
+     */
+    std::vector<std::string> getLabels() const;
+
+    /**
+     * Returns a binary mask where pixels belonging to the specified instance/class index are set to 1, others to 0. If mask data is not set, returns an empty
+     * vector.
+     */
+
```
aljazdu: Nitpick. Remove the empty line.

src/pipeline/datatype/SegmentationMask.cpp
```diff
@@ -0,0 +1,372 @@
...
+    cv::Size size(static_cast<int>(getWidth()), static_cast<int>(getHeight()));
+    int type = CV_8UC1;
+
+    const size_t requiredSize = CV_ELEM_SIZE(type) * static_cast<size_t>(size.area());
+    const size_t actualSize = data->getSize();
+
+    DAI_CHECK_V(actualSize == requiredSize, "Segmentation mask data size does not match the expected size, required {}, actual {}.", requiredSize, actualSize);
+
+    mask = cv::Mat(size, type, data->getData().data());
+    CV_Assert(mask.type() == CV_8UC1);
```
aljazdu: Nitpick, but is this assert necessary? I don't see any way this not being the correct type.

include/depthai/pipeline/datatype/SegmentationMask.hpp
```diff
@@ -0,0 +1,219 @@
...
+    /**
+     * Serialize schema to proto buffer
+     *
+     * @returns serialized schema
+     */
+    ProtoSerializable::SchemaPair serializeSchema() const override;
+#endif
+
+    DEPTHAI_SERIALIZE(SegmentationMask, Buffer::ts, Buffer::tsDevice, Buffer::sequenceNum, transformation, width, height, labels);
```
aljazdu, line 200: Are there any issues that certain `.proto` fields are missing here? Such as `sequenceNum` and `data`.

include/depthai/pipeline/datatype/SegmentationParserConfig.hpp
```diff
@@ -0,0 +1,54 @@
...
+    virtual ~SegmentationParserConfig();
+
+    /**
+     * Add a confidence threshold to the argmax operation over the segmentation tensor.
+     * Pixels with confidence values below this threshold will be assigned the background class (255).
+     * @param threshold Confidence threshold for segmentation parsing
+     * @note Default is -1.0f, which means no thresholding is applied.
+     * @note Only applicable if output classes are not in a single layer (eg. classesInOneLayer = false).
+     */
+    void setConfidenceThreshold(float threshold);
```
aljazdu, line 25: Nitpick for `const`

src/pipeline/utilities/SegmentationParser/SegmentationParserUtils.hpp
```diff
@@ -0,0 +1,84 @@
...
+ * @param channelStart Index of the starting channel to consider (default is 0, set to 1 if background class is part of the model output)
+ * @param logger Logger
+ */
+template <typename T>
+void tensorArgmax(span<std::uint8_t> dstData,
+                  span<const uint8_t> tensorSpan,
+                  dai::TensorInfo& tensorInfo,
+                  dai::SegmentationParserConfig& config,
+                  int channelStart,
+                  std::shared_ptr<spdlog::async_logger>& logger);
```
aljazdu, line 46: This declaration does not match the definition in the `.cpp` file:

```
template <typename T, bool backgroundClass>
void tensorArgmax(span<std::uint8_t> dstData,
                  const span<const uint8_t> tensorSpan,
                  dai::TensorInfo& tensorInfo,
                  dai::SegmentationParserConfig& config,
                  std::shared_ptr<spdlog::async_logger>& logger) {
```

You have 2 options:
1. Since `tensorArgmax()` is currently a helper function inside of `.cpp` file, you can remove the declaration in the `.hpp`. 
2. In case you want to keep this function publicly available to other callers which include the header, then you must match the declaration and definition to avoid potential linking errors.

src/pipeline/datatype/SegmentationMask.cpp
aljazkonec1: Just for my clarification what you mean is: You construct an empty message object at time t and at set its data at time t+1 but the timestamp would still show t ? In that case I agree with your comment. I'll update the overloads, thanks!

src/pipeline/datatype/SegmentationMask.cpp
aljazkonec1: Good catch, added.

src/pipeline/utilities/SegmentationParser/SegmentationParserUtils.hpp
```diff
@@ -0,0 +1,84 @@
...
+ * @param config SegmentationParserConfig containing parameters like confidenceThreshold and stepSize
+ * @param channelStart Index of the starting channel to consider (default is 0, set to 1 if background class is part of the model output)
+ * @param logger Logger
+ */
+void tensorArgmaxFP16(span<std::uint8_t> dst,
+                      span<const uint8_t> tensorSpan,
+                      dai::TensorInfo& tensorInfo,
+                      dai::SegmentationParserConfig& config,
+                      int channelStart,
+                      std::shared_ptr<spdlog::async_logger>& logger);
```
aljazdu, line 62: same as above for `tensorArgmax()`

src/pipeline/datatype/SegmentationMask.cpp
aljazdu: Yes exactly

src/pipeline/utilities/SegmentationParser/SegmentationParserUtils.hpp
aljazkonec1: Thanks for catching it! updated the signatures.

include/depthai/pipeline/datatype/SegmentationMask.hpp
aljazkonec1: Only reason is friend access for RVC2 yes. Will change to private.

include/depthai/pipeline/datatype/SegmentationMask.hpp
aljazkonec1: I followed ImgFrame.hpp implementation where `using` declarations but overlooked that it is because it overrides the functions. Removed.

include/depthai/pipeline/datatype/SegmentationMask.hpp
aljazkonec1: Implemented in the cpp. If in header, clang-tidy warning: 
"Parameter 'width' is const-qualified in the function declaration; const-qualification of parameters only has an effect in function definitions (fix available)clang-tidyreadability-avoid-const-params-in-decls"
cc @moratom should we disable this clang-tidy check?

include/depthai/pipeline/datatype/SegmentationMask.hpp
aljazkonec1: sequenceNum is serialized, data is a raw payload that is handled by the buffer transport. This macro just serializes the metadata. [Adjusted proto file](https://github.com/luxonis/depthai-core/pull/1661/changes/4021b5f94c107e270dbdbf207e2af8b2e6a06fae)

include/depthai/pipeline/datatype/SegmentationMask.hpp
aljazdu: Looked it up, the header const has not effect in this case -> same signature (should be kept in definitions though). I guess it's a preference whether you like to be consistent across declarations and definitions, or as concise as possible - omitting const where it can be omitted without effect.

src/pipeline/utilities/SegmentationParser/SegmentationParserUtils.hpp
aljazkonec1: [Added](https://github.com/luxonis/depthai-core/pull/1661/changes/decdbe4fa530ffe5dcde91eda5c2b278dee34255)

src/pipeline/utilities/SegmentationParser/SegmentationParserUtils.hpp
aljazkonec1: [Added](https://github.com/luxonis/depthai-core/pull/1661/changes/decdbe4fa530ffe5dcde91eda5c2b278dee34255)

include/depthai/pipeline/datatype/SegmentationMask.hpp
aljazkonec1: [Added](https://github.com/luxonis/depthai-core/pull/1661/changes/c5c223d323cf586370304813d8e33fa8fa534193)

include/depthai/pipeline/datatype/SegmentationMask.hpp
aljazkonec1: [Removed](https://github.com/luxonis/depthai-core/pull/1661/changes/c5c223d323cf586370304813d8e33fa8fa534193)

src/pipeline/datatype/SegmentationMask.cpp
aljazkonec1: [Added](https://github.com/luxonis/depthai-core/pull/1661/changes/f17d4014897b9e1b6601929fd3ba3be8e5b70d5d)

src/pipeline/datatype/SegmentationMask.cpp
aljazkonec1: [Added](https://github.com/luxonis/depthai-core/pull/1661/changes/f17d4014897b9e1b6601929fd3ba3be8e5b70d5d)

include/depthai/pipeline/node/SegmentationParser.hpp
```diff
@@ -0,0 +1,129 @@
...
+     *
+     * @note Only applicable if the number of classes is greater than 1 and the output classes are not in a single layer (eg. classesInOneLayer = false).
+     */
+    void setBackgroundClass(bool backgroundClass);
+
+    /**
+     * Gets whether the first class (index 0) is considered the background class.
+     */
+    bool getBackgroundClass() const;
+
```
moratom, line 97: Why is the background class special? Is it not just one of options?

src/pipeline/datatype/SegmentationMask.cpp
```diff
@@ -0,0 +1,387 @@
...
+}
+std::size_t SegmentationMask::getHeight() const {
+    return height;
+}
+
+void SegmentationMask::setMask(const std::vector<std::uint8_t>& mask, size_t width, size_t height) {
+    if(mask.size() != width * height) {
+        throw std::runtime_error("SegmentationMask: data size does not match width*height");
+    }
+    setTimestamp(std::chrono::steady_clock::now());
```
moratom: Do we do the same in ImgFrame?
Could behave in unexpected ways IMO, if one makes the message, then sets the timestamp and later calls `setMask`.

I'm up for removing this in all setters.

include/depthai/pipeline/datatype/SegmentationMask.hpp
```diff
@@ -0,0 +1,212 @@
...
+     * Returns the segmentation mask as an ImgFrame. If mask data is not set, returns an empty frame with only metadata set.
+     */
+    dai::ImgFrame getFrame() const;
+
+    /**
+     * Returns the area (number of pixels) of the specified instance/class index in the segmentation mask.
+     * @param index Instance/Class index
+     * @note If index is not present in the mask, returns std::nullopt.
+     */
+    std::optional<int32_t> getArea(uint8_t index) const;
```
moratom: Nit: Does this ever return 0?
Maybe better to use uint32_t to as area can't be negative.

include/depthai/pipeline/datatype/SegmentationMask.hpp
```diff
@@ -0,0 +1,212 @@
...
+};
+
+// Private access for setting size on RVC2
+namespace priv {
+struct SegmentationMaskAccess {
+    static void setSize(SegmentationMask& mask, size_t width, size_t height) {
+        mask.setSizeInternal(width, height);
+    }
+};
+}  // namespace priv
```
moratom: on RVC2?
Why is this needed?

include/depthai/pipeline/datatype/SegmentationMask.hpp
```diff
@@ -0,0 +1,212 @@
...
+namespace dai {
+
+namespace priv {
+struct SegmentationMaskAccess;
+}  // namespace priv
+
+/**
+ * SegmentationMask message.
+ *
+ * Segmentation mask of an image is stored as a single-channel INT8 array, where each value represents a class or instance index.
```
moratom: int8 or uint8?

moratom: (no such thing as 255 in int8)

include/depthai/pipeline/datatype/SegmentationMask.hpp
```diff
@@ -0,0 +1,212 @@
...
+    std::vector<std::vector<dai::Point2f>> getContour(uint8_t index);
+
+    /**
+     * Returns the bounding box of the specified instance/class index.
+     * @param index class index
+     * @param useContour If true, the rotation of the bounding box is calculated using the opencv functions findContours and minAreaRect, otherwise a fast
+     * calculation of the axis-aligned bounding box is returned.
+     */
+    std::optional<dai::RotatedRect> getBoundingBox(uint8_t index, bool useContour = false);
+
```
moratom, line 181: Would it make more sense to return a vector of bboxes, one for every continous area?

examples/python/Segmentation/semantic_segmentation.py
```diff
@@ -0,0 +1,92 @@
...
+    segFrame = segFrame.astype(np.uint8)
+    segFrame[segFrame < 255] = (segFrame[segFrame < 255] * colorScalingFactor).astype(np.uint8)
+    coloredSegFrame = cv2.applyColorMap(segFrame, cv2.COLORMAP_JET)
+    coloredSegFrame[segFrame == 255] = 0
+    uniqueLabelIndexes = segMask.getUniqueIndices()
+    if len(segMask.getLabels()) > 0:
+        for i, labelIndex in enumerate(uniqueLabelIndexes):
+            labelStr = segMask.getLabels()[labelIndex]
+            color_value = np.array([[int(labelIndex * colorScalingFactor)]], dtype=np.uint8)
+            text_color = cv2.applyColorMap(color_value, cv2.COLORMAP_JET)[0, 0].tolist()
```
moratom: camelCase

examples/python/Segmentation/semantic_segmentation.py
```diff
@@ -0,0 +1,92 @@
...
+    modelName = "luxonis/deeplab-v3-plus:person-256x256"
+
+with dai.Pipeline(device) as pipeline:
+    print("Creating pipeline...")
+
+    cameraNode = pipeline.create(dai.node.Camera).build()
+    neuralNetwork = pipeline.create(dai.node.NeuralNetwork).build(cameraNode, modelName)
+
+    segParser = pipeline.create(dai.node.SegmentationParser).build(neuralNetwork.out, modelName)
+    segParser.initialConfig.setConfidenceThreshold(0.0)
```
moratom: Should this be the default, if we use it in the example? Why/why not?

examples/cpp/Segmentation/semantic_segmentation.cpp
```diff
@@ -0,0 +1,154 @@
...
+
+    auto cameraNode = pipeline.create<dai::node::Camera>();
+    cameraNode->build();
+
+    auto neuralNetwork = pipeline.create<dai::node::NeuralNetwork>();
+    neuralNetwork->build(cameraNode, modelName);
+
+    auto segParser = pipeline.create<dai::node::SegmentationParser>();
+    segParser->build(neuralNetwork->out, modelName);
+    segParser->setBackgroundClass(true);
```
moratom: Is this part of the NNArchive information? Why do we need to set it in the example?

src/pipeline/utilities/SegmentationParser/SegmentationParserUtils.hpp
```diff
@@ -0,0 +1,81 @@
...
+ * @param nnData NNData object containing the tensor data.
+ * @param tensorInfo TensorInfo object describing the tensor layout and data type.
+ * @param config SegmentationParserConfig object containing parsing parameters like confidenceThreshold and stepSize.
+ * @param backgroundClass If true, the first class (index 0) is considered as background and ignored during calculating argmax.
+ * @param logger Logger
+ */
+void computeSegmentationMask(dai::SegmentationMask& outputMask,
+                             dai::NNData& nnData,
+                             dai::TensorInfo& tensorInfo,
+                             dai::SegmentationParserConfig& config,
```
moratom: Should all of these be const?

moratom: It will likely need fixes in making the relevant getters const, but shouldn't be too big of a change.

src/pipeline/utilities/SegmentationParser/SegmentationParserUtils.cpp
```diff
@@ -0,0 +1,426 @@
...
+const T* assumeAlignedPtr(const uint8_t* ptr) {
+#if defined(__GNUC__) || defined(__clang__)
+    return static_cast<const T*>(__builtin_assume_aligned(ptr, alignof(T)));
+#elif defined(_MSC_VER)
+    __assume((reinterpret_cast<uintptr_t>(ptr) & (alignof(T) - 1)) == 0);
+    return reinterpret_cast<const T*>(ptr);
+#else
+    return reinterpret_cast<const T*>(ptr);
+#endif
+}
```
moratom: I don't think we can safely assume the ptr is aligned...
How much slower does the algorithm work with this? We should add a runtime check for this instead.

include/depthai/pipeline/datatype/SegmentationMask.hpp
aljazkonec1: No. We could return 0 if the specified class index has 0 area.

include/depthai/pipeline/datatype/SegmentationMask.hpp
aljazkonec1: In this case, I would change the return to `std::vector<dai::ImgDetection>` to store class info for each bbox. Is that OK?

include/depthai/pipeline/node/SegmentationParser.hpp
aljazkonec1: Some models predict the background class while others dont. In the NNArchives we have a flag where `backgroundClass = true;` means that the class at index 0 is the background. User can decide if they want to combine that class into the background index of 255 or keep it as its own value in the segmask

src/pipeline/utilities/SegmentationParser/SegmentationParserUtils.cpp
aljazkonec1: Could you elaborate? The function is only called when the NNData tensor is vectorizable (data is tightly packed). The check is [here](https://github.com/luxonis/depthai-core/blob/3cd89608fa6aca09ec02b8a1e98178866e912d92/src/pipeline/utilities/SegmentationParser/SegmentationParserUtils.cpp#L238-L244)

include/depthai/pipeline/datatype/SegmentationMask.hpp
aljazkonec1: Needed when converting from lpb type to dai type because the data stream (buffer) isnt copied so cant use the `setMask` functions to set the size and we dont want to expose the width / height parameters as they shouldn't be manipulated without manipulating the segmentation mask data.

src/pipeline/utilities/SegmentationParser/SegmentationParserUtils.cpp
aljazkonec1: Removed the function and kept only memory safe non-vectorized function as it has minimal performance difference in the current implementation.

PR #1673, opened 2026-02-05
Bugfix: gate node queue size

JakubFara:
# Why
- RVC2 did not reflect the setup of the queue (size, blocking), which causes an issue
- Change the default size of the input queue
- Add tests for the FW fix


include/depthai/pipeline/node/Gate.hpp
```diff
@@ -37,7 +37,7 @@ class Gate : public DeviceNodeCRTP<DeviceNode, Gate, GateProperties> {
      * * Default queue size: 4
```
moratom: Update the comment as well

JakubFara: done here: https://github.com/luxonis/depthai-core/pull/1673/commits/f9cc18a5bef76510b9c9539260daa5b931581123

PR #1701, opened 2026-03-04
Object tracker update

aljazdu:
## Purpose
This PR updates the current implementation of the object tracker. The updated object tracker expands the use of 2D bounding boxes with the spatial coordinates (if available) of the detected objects. 
The update also adds velocity and speed measurement to the tracked objects.

## Specification
The PR updates the following:
- Object tracker implementation (with an additional Kalman filter used for 3D spatial coordinates)
- Tracklet data (by adding velocity and speed to the existing list of members)

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation
This was tested with side by side comparison of performance of the updated and current object tracker.

## Summary by CodeRabbit

* **New Features**
  * Spatial association added to object tracking for improved 3D-aware matching, tie-breaking and acceptance logic, yielding more robust tracking.
  * Tracklets now expose motion properties (velocity and speed) for richer motion analysis and reporting.
  * New spatial configuration options: enable/disable spatial association, association weight, distance threshold, and depth-aware scaling.

* **Chores**
  * Updated device configuration/version identifiers.


include/depthai/pipeline/datatype/Tracklets.hpp
```diff
@@ -53,7 +53,15 @@ struct Tracklet {
      * Spatial coordinates of tracklet.
      */
     Point3f spatialCoordinates;
-    DEPTHAI_SERIALIZE(Tracklet, roi, id, label, age, status, srcImgDetection, spatialCoordinates);
+    /**
+     * Estimated 3D velocity of the tracklet in mm/s.
```
asahtik: How come we're using mm/s instead of the more standard m/s?

src/utility/ObjectTrackerImpl.cpp
```diff
@@ -927,7 +1150,9 @@ std::tuple<std::vector<Eigen::Matrix<int, 1, 2>>, std::vector<int>, std::vector<
     Eigen::Matrix<int, 1, 2> tmp;
     for(int i = 0; i < matched_indices.rows(); i++) {
         tmp = (matched_indices.row(i)).cast<int>();
-        if(iou_matrix(tmp(0), tmp(1)) < iou_threshold) {
+        const bool hasSpatial = spatial_available && spatial_pair_valid(tmp(0), tmp(1));
+        const bool pairAccepted = hasSpatial ? spatial_gate_ok(tmp(0), tmp(1)) : (iou_matrix(tmp(0), tmp(1)) >= iou_threshold);
```
asahtik: Maybe we should take into account iou + spatial here instead of just spatial?

asahtik: Or is iou already taken into account in spatial_gate?

include/depthai/pipeline/datatype/Tracklets.hpp
aljazdu: Fixed

src/utility/ObjectTrackerImpl.cpp
```diff
@@ -708,10 +944,11 @@ std::vector<Eigen::RowVectorXf> OCSTracker::State::update(const std::vector<ImgD
     for(size_t i = 0; i < detections.size(); i++) {
         uint32_t index = track_by_class ? detections[i].label : 0;
         dets_map[index].first.push_back(detections[i]);
-        if(i <= spatialData.size())
+        if(i < spatialData.size()) {
             dets_map[index].second.push_back(spatialData[i]);
-        else
+        } else {
             dets_map[index].second.push_back(Point3f(0, 0, 0));
+        }
```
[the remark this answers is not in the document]
aljazdu: not an issue.

src/utility/ObjectTrackerImpl.cpp
```diff
@@ -374,6 +493,31 @@ void OCSTracker::State::KalmanBoxTracker::update(Eigen::Matrix<float, 5, 1>* bbo
...
+        kf_spatial->R = Eigen::MatrixXf::Identity(3, 3);
+        kf_spatial->R(0, 0) = xyStd * xyStd;
+        kf_spatial->R(1, 1) = xyStd * xyStd;
+        kf_spatial->R(2, 2) = zStd * zStd;
+        kf_spatial->update(spatialZ);
+        spatial_observation = spatialMeasurement;
+        spatial_prediction.position.head<3>() = kf_spatial->x.head<3>();
+        spatial_prediction.validPosition = true;
+        spatial_velocity = kf_spatial->x.segment<3>(3);
+        has_spatial_state = true;
```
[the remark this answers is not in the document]
aljazdu: this has been fixed. check again

include/depthai/properties/ObjectTrackerProperties.hpp
```diff
@@ -72,6 +72,23 @@ struct ObjectTrackerProperties : PropertiesSerializable<Properties, ObjectTracke
...
+    bool spatialAssociation = true;
+    /**
+     * Weight for spatial distance in association cost [0,1].
+     */
+    float spatialAssociationWeight = 0.5f;
+    /**
+     * Base 3D gating threshold in millimeters used for spatial association.
+     * When using ObjectTracker::setSpatialDistanceThreshold(), provide meters and it will be converted to this internal unit.
+     */
+    float spatialDistanceThreshold = 1500.0f;
```
asahtik: Is there a reason why this property is in mm? IMO (if possible) this should be in m and any conversions necessary for the OT impl should be done in the node so that this property, `setSpatialDistanceThreshold` and output units are consistent.

src/utility/ObjectTrackerImpl.cpp
asahtik, line 781: If I understand this correctly, this accepts matches based only on iou and ignores spatial. Is this intentional?

asahtik (commented):
Left a couple of comments / questions. Also [this comment by coderabbit](https://github.com/luxonis/depthai-core/pull/1701#pullrequestreview-4012416913) seems to not be handled yet (in the outside diff range comments section)?
Other than that LGTM

src/utility/ObjectTrackerImpl.cpp
aljazdu: Yes in this particular step it is. Spatial data is used only as a soft ranking (a tie breaker) between very closely scored matches. However it is not used as a "gating" mechanism like in the first step.

include/depthai/properties/ObjectTrackerProperties.hpp
aljazdu: Agreed, fix in the update.

asahtik (approved):
Nitpick: I'd move the conversion from m to mm from `ObjectTracker.cpp:113` to the OCSTracker constructor and then remove `Internally converted to millimeters` comment in `ObjectTrackerProperties.hpp:85` since it's not really relevant to the user.
Otherwise LGTM

PR #1715, opened 2026-03-12
Feature: PointCloudNode

pheec:
## Purpose
Adds the PointCloud node

## Specification
- Added `PointCloud` node implementation (`src/pipeline/node/PointCloud.cpp`) with:
  - Point cloud output modes, controlled via `keepPointCloudOrganized()`
  - CPU, multi-threaded CPU, and GPU (Kompute) compute backends copied from RGBD node.
  - Optional coordinate system transformation via `setTargetCoordinateSystem()` (camera socket or housing coordinate system)
  - Configurable length units via `setLengthUnit()`
  - Runtime calibration change detection using the `getEepromId()` pattern (same as `ImageAlign`) — automatically reinitializes when calibration is updated during runtime
- Updated `PointCloudData` to support organized (width×height) and filtered (N×1) layouts with bounding box metadata
- Updated `pointcloud_test.cpp` to reflect the current API (`keepPointCloudOrganized()` replaces the removed `initialConfig->setSparse()`)

- Depends on `StereoDepth` node output for depth input
- `PointCloudConfig::setSparse()` is no longer the control mechanism for output layout, replaced by `keepPointCloudOrganized()` on the node. 

## Deployment Plan
- **The option to get colored pointcloud will be added after the merge of image transformations - extrinsics**

## Testing & Validation
- `pointcloud_test`: covers organized and sparse output modes (10 frames each), validates dimensions, point count, and bounding box correctness
- `pointclouddata_test`: validates `PointCloudData` serialization and accessors
- `point_cloud_test`: unit tests for the compute backend (intrinsics, extrinsics, length units)

## Summary by CodeRabbit

* **New Features**
  * New PointCloud examples (showcase + visualizer) with live 3D rendering, colorized depth preview, and build/run instructions.
  * Configurable length unit, host-run toggle, CPU/CPU-MT/GPU backends, threading/device controls, frame-pool control, and coordinate-system transform options.

* **Breaking Changes**
  * Mode renamed: "sparse" → "organized"; use getOrganized()/isOrganized(). Legacy sparse APIs deprecated and emit warnings; serialization now derives sparsity from height.

* **Tests / Docs**
  * Extensive new tests and PointCloud README added.


bindings/python/src/pipeline/datatype/PointCloudConfigBindings.cpp
```diff
@@ -41,23 +40,7 @@ void bind_pointcloudconfig(pybind11::module& m, void* pCallstack) {
 
     // Message
     config.def(py::init<>())
-        .def("__repr__", &PointCloudConfig::str)
```
MaticTonin, line 44: Are we fine with just removing them or do we wanna add that this will be depricated message.

bindings/python/src/pipeline/datatype/PointCloudConfigBindings.cpp
```diff
@@ -42,22 +41,19 @@ void bind_pointcloudconfig(pybind11::module& m, void* pCallstack) {
     // Message
     config.def(py::init<>())
         .def("__repr__", &PointCloudConfig::str)
-        // .def(py::init<std::shared_ptr<RawPointCloudConfig>>())
-
-        // .def("set", &PointCloudConfig::set, py::arg("config"), DOC(dai, PointCloudConfig, set))
-        // .def("get", &PointCloudConfig::get, DOC(dai, PointCloudConfig, get))
-        .def("getSparse", &PointCloudConfig::getSparse, DOC(dai, PointCloudConfig, getSparse))
```
MaticTonin, line 49: Would be good to add that this command is deprecated so that we do not suddenly break already ecsistsing pipelines

examples/cpp/PointCloud/PointCloudShowcase.cpp
```diff
@@ -0,0 +1,223 @@
+/**
+ * PointCloud Node Showcase
```
MaticTonin, line 2: IMO, rather add a readMe.md with as well some visuals to the example here and keep them minimal than overexplode them (so they can be reused in docs).

examples/python/PointCloud/point_cloud_showcase.py
JakubFara, line 1: Add the python examples to [examples/python/CMakeLists.txt](https://github.com/luxonis/depthai-core/blob/main/examples/python/CMakeLists.txt) .

examples/python/PointCloud/PointCloudShowcase.py
```diff
@@ -0,0 +1,213 @@
...
+  C) **Custom 4×4 matrix** – applied via ``initialConfig`` or the runtime
+     ``inputConfig`` queue.
+     ``pc.initialConfig.setTransformationMatrix([[...], ...])``
+     This can also be combined with (A) or (B); the custom matrix is
+     applied *after* the calibration-based transform.
+"""
+
+import time
+import os
+# os.environ["DEPTHAI_DEVICE_RVC4_FWP"] = "/home/tomas/code/depthai-device-kb/build_docker_arm64_rvc4/RelWithDebInfo/depthai-device-rvc4-fwp.tar.xz"
```
MaticTonin: :)

examples/python/PointCloud/PointCloudShowcase.py
```diff
@@ -0,0 +1,213 @@
...
+    # ------------------------------------------------------------------
+    with dai.Pipeline(device) as pipeline:
+        left = pipeline.create(dai.node.Camera).build(dai.CameraBoardSocket.CAM_B)
+        right = pipeline.create(dai.node.Camera).build(dai.CameraBoardSocket.CAM_C)
+        stereo = pipeline.create(dai.node.StereoDepth)
+        left.requestFullResolutionOutput().link(stereo.left)
+        right.requestFullResolutionOutput().link(stereo.right)
+
+        # ── 1. Sparse point cloud  (METER, multi-threaded, large pool) ────
+        pc_sparse = pipeline.create(dai.node.PointCloud)
```
MaticTonin: camelCase instead of _ for whole example

examples/python/PointCloud/PointCloudVisualizer.py
```diff
@@ -0,0 +1,112 @@
...
+        right.requestOutput((640, 400)).link(stereo.right)
+
+        # ── PointCloud node ───────────────────────────────────────────
+        pc = pipeline.create(dai.node.PointCloud)
+        pc.setRunOnHost(True)
+        pc.setLengthUnit(dai.LengthUnit.METER)
+        stereo.depth.link(pc.inputDepth)
+
+        queue = pc.outputPointCloud.createOutputQueue(maxSize=4, blocking=False)
+        q_depth = stereo.depth.createOutputQueue(maxSize=4, blocking=False)
```
MaticTonin: Same here, camelCase insead of _ whole example

examples/python/PointCloud/point_cloud_visualizer.py
JakubFara, line 1: Use thisFormat instead of this_format

include/depthai/pipeline/datatype/PointCloudConfig.hpp
```diff
@@ -1,64 +1,75 @@
 #pragma once
 
-#include <unordered_map>
+#include <array>
```
MaticTonin, line 3: From changes below, it does not seem like we require this in .hpp, could wrong.

include/depthai/pipeline/node/PointCloud.hpp
```diff
@@ -2,30 +2,109 @@
...
 
-// standard
-#include <fstream>
-
 // shared
 #include <depthai/properties/PointCloudProperties.hpp>
 #include <memory>
 
 #include "depthai/pipeline/datatype/PointCloudConfig.hpp"
+#include "depthai/pipeline/datatype/StereoDepthConfig.hpp"
```
MaticTonin: Required?

include/depthai/pipeline/datatype/PointCloudConfig.hpp
```diff
@@ -1,64 +1,75 @@
...
+     * Enable or disable organized point cloud output.
+     * When true all width*height points are kept; when false only points with z > 0 are emitted.
+     */
+    PointCloudConfig& setOrganized(bool enable);
+
+    /**
+     * Set a 4x4 transformation matrix applied to every output point.
+     * Default is the identity matrix.
      */
     PointCloudConfig& setTransformationMatrix(const std::array<std::array<float, 4>, 4>& transformationMatrix);
```
JakubFara, line 70: Can we replace this with the extrinsics, which will be added to ImgFrame? - Transformation with respect to the lowest index.

src/pipeline/datatype/PointCloudData.cpp
```diff
@@ -141,8 +151,53 @@ PointCloudData& PointCloudData::setMaxZ(float val) {
...
     return *this;
 }
-PointCloudData& PointCloudData::setSparse(bool val) {
-    sparse = val;
+PointCloudData& PointCloudData::setSparse(bool /*val*/) {
+    // Deprecated no-op - organization is now determined by width/height (isOrganized = height > 1)
+    return *this;
+}
+
+PointCloudData& PointCloudData::updateBoundingBox() {
```
JakubFara, line 174: Do we need to store the bounding box in the pointCloudData? Would it not be better to have just a function `BoundingBox getBoundingBox()`?

bindings/python/src/pipeline/datatype/PointCloudDataBindings.cpp
```diff
@@ -169,5 +170,6 @@ void bind_pointclouddata(pybind11::module& m, void* pCallstack) {
         .def("setMaxX", &PointCloudData::setMaxX, DOC(dai, PointCloudData, setMaxX))
         .def("setMaxY", &PointCloudData::setMaxY, DOC(dai, PointCloudData, setMaxY))
         .def("setMaxZ", &PointCloudData::setMaxZ, DOC(dai, PointCloudData, setMaxZ))
-        .def("setInstanceNum", &PointCloudData::setInstanceNum, DOC(dai, PointCloudData, setInstanceNum));
+        .def("setInstanceNum", &PointCloudData::setInstanceNum, DOC(dai, PointCloudData, setInstanceNum))
```
MaticTonin: setInstance defined twice in bindings while getInstance was removed.

bindings/python/src/pipeline/node/PointCloudBindings.cpp
```diff
@@ -37,7 +37,22 @@ void bind_pointcloud(pybind11::module& m, void* pCallstack) {
...
+        .def("useGPU", &PointCloud::useGPU, py::arg("device") = 0, DOC(dai, node, PointCloud, useGPU))
+        .def("setTargetCoordinateSystem", 
+             py::overload_cast<CameraBoardSocket, bool>(&PointCloud::setTargetCoordinateSystem),
+             py::arg("targetCamera"), 
+             py::arg("useSpecTranslation") = false,
+             DOC(dai, node, PointCloud, setTargetCoordinateSystem))
+        .def("setTargetCoordinateSystem",
+             py::overload_cast<HousingCoordinateSystem, bool>(&PointCloud::setTargetCoordinateSystem),
+             py::arg("housingCS"),
+             py::arg("useSpecTranslation") = false,
```
MaticTonin: useSpecHousing coordinate system as false, while in C++ implementation its true, fix consistency as default false

examples/python/PointCloud/PointCloudShowcase.py
```diff
@@ -0,0 +1,213 @@
...
+
+
+def main() -> None:
+    print("PointCloud Node Showcase")
+    print("========================")
+    print("Connecting to device...")
+
+    # device = dai.Device()
+    # print(f"Device: {device.getDeviceName()}  (ID: {device.getDeviceId()})\n")
+    device_ip = "10.11.0.29"
```
MaticTonin: Remove hardcoded device as well

bindings/python/src/pipeline/node/PointCloudBindings.cpp
```diff
@@ -37,7 +37,22 @@ void bind_pointcloud(pybind11::module& m, void* pCallstack) {
...
-        .def("setNumFramesPool", &PointCloud::setNumFramesPool, DOC(dai, node, PointCloud, setNumFramesPool));
+        .def("setNumFramesPool", &PointCloud::setNumFramesPool, py::arg("numFramesPool"), DOC(dai, node, PointCloud, setNumFramesPool))
+        .def("setRunOnHost", &PointCloud::setRunOnHost, py::arg("runOnHost"), DOC(dai, node, PointCloud, setRunOnHost))
+        .def("useCPU", &PointCloud::useCPU, DOC(dai, node, PointCloud, useCPU))
+        .def("useCPUMT", &PointCloud::useCPUMT, py::arg("numThreads") = 2, DOC(dai, node, PointCloud, useCPUMT))
+        .def("useGPU", &PointCloud::useGPU, py::arg("device") = 0, DOC(dai, node, PointCloud, useGPU))
+        .def("setTargetCoordinateSystem", 
+             py::overload_cast<CameraBoardSocket, bool>(&PointCloud::setTargetCoordinateSystem),
+             py::arg("targetCamera"), 
+             py::arg("useSpecTranslation") = false,
```
MaticTonin, line 53: Same here

include/depthai/pipeline/datatype/PointCloudConfig.hpp
```diff
@@ -1,64 +1,75 @@
...
 
     std::array<std::array<float, 4>, 4> transformationMatrix = {{{1, 0, 0, 0}, {0, 1, 0, 0}, {0, 0, 1, 0}, {0, 0, 0, 1}}};
 
+    LengthUnit lengthUnit = LengthUnit::MILLIMETER;
+
    public:
-    /**
-     * Construct PointCloudConfig message.
-     */
     PointCloudConfig() = default;
```
MaticTonin, line 39: Add it back, no required to be removed

include/depthai/pipeline/datatype/PointCloudConfig.hpp
```diff
@@ -1,64 +1,75 @@
...
-     * Set 3x3 transformation matrix for point cloud calculation. Default is an identity matrix.
-     * @param transformationMatrix
+     * Convenience overload: set a 3x3 rotation matrix (translation set to zero).
      */
     PointCloudConfig& setTransformationMatrix(const std::array<std::array<float, 3>, 3>& transformationMatrix);
 
+    /**
+     * Set the length unit for output point coordinates.
+     */
+    PointCloudConfig& setLengthUnit(LengthUnit unit);
```
MaticTonin, line 79: Add as well in docs, what is default unit set in PCLNode.

include/depthai/pipeline/datatype/PointCloudData.hpp
```diff
@@ -91,9 +91,18 @@ class PointCloudData : public Buffer, public ProtoSerializable {
 
     /**
      * Retrieves whether point cloud is sparse
+     * @deprecated Use isOrganized() instead. Sparse means height == 1
      */
+    [[deprecated("isSparse is deprecated, use isOrganized() instead (isSparse == !isOrganized())")]]
```
MaticTonin, line 97: I would not go for == sign, since it is little confusing, rather go with wording

include/depthai/pipeline/node/PointCloud.hpp
```diff
@@ -2,30 +2,109 @@
...
+       public:
+        Impl() = default;
+        
+        void setLogger(std::shared_ptr<spdlog::logger> log);
+        
+        // Compute DENSE point cloud (width * height points, includes invalid z=0 or negative)
+        void computePointCloudDense(const uint8_t* depthData, std::vector<Point3f>& points);
+        
+        // Apply extrinsic transformation to points
+        void applyTransformation(std::vector<Point3f>& points);
```
MaticTonin: Every public method should have its own documentation of docstrings

include/depthai/pipeline/node/PointCloud.hpp
```diff
@@ -2,30 +2,109 @@
...
 /**
  * @brief PointCloud node. Computes point cloud from depth frames.
  */
-class PointCloud : public DeviceNodeCRTP<DeviceNode, PointCloud, PointCloudProperties> {
+class PointCloud : public DeviceNodeCRTP<DeviceNode, PointCloud, PointCloudProperties>, public HostRunnable {
    public:
     constexpr static const char* NAME = "PointCloud";
 
+    class Impl {
+       public:
```
MaticTonin, line 41: Do we want Impl to be a public class?

include/depthai/pipeline/datatype/PointCloudConfig.hpp
```diff
@@ -1,64 +1,75 @@
...
+     * Default is the identity matrix.
      */
     PointCloudConfig& setTransformationMatrix(const std::array<std::array<float, 4>, 4>& transformationMatrix);
 
     /**
-     * Set 3x3 transformation matrix for point cloud calculation. Default is an identity matrix.
-     * @param transformationMatrix
+     * Convenience overload: set a 3x3 rotation matrix (translation set to zero).
      */
     PointCloudConfig& setTransformationMatrix(const std::array<std::array<float, 3>, 3>& transformationMatrix);
```
MaticTonin, line 74: setTranformationMatrix, when its 3x3 is a little missleading, I would then rather seperate to basicly define one setter for Rotation and one for Offset

src/pipeline/node/PointCloud.cpp
```diff
@@ -1,12 +1,284 @@
...
+            break;
+        case ComputeMethod::GPU:
+            computePointCloudDenseGPU(depthData, points);
+            break;
+    }
+}
+
+void PointCloud::Impl::applyTransformation(std::vector<Point3f>& points) {
+    if(!hasExtrinsics) {
+        if(logger) {
```
MaticTonin, line 79: Is there actually an instance where logger does not extsists

src/pipeline/node/PointCloud.cpp
```diff
@@ -1,12 +1,284 @@
...
+            sparsePoints.push_back(p);
+        }
+    }
+
+    return sparsePoints;
+}
+
+void PointCloud::Impl::setLengthUnit(dai::LengthUnit lengthUnit) {
+    // Check if unit actually changed
+    bool unitChanged = (targetLengthUnit != lengthUnit);
```
MaticTonin: Unnecesarry, can be direcly as this in logger

src/pipeline/node/PointCloud.cpp
```diff
@@ -1,12 +1,284 @@
...
+
+    targetLengthUnit = lengthUnit;
```
MaticTonin: WOuld move under 95 so it is scale factor and targetLengthUnit = lengthUnit togeter + added comment, update new defualt values above, so you know why it was done like this

src/pipeline/node/PointCloud.cpp
```diff
@@ -1,12 +1,284 @@
...
+    lengthUnitMultiplier = getLengthUnitMultiplier(lengthUnit);
+
+    // Depth values from sensor are in millimeters (uint16_t raw values)
+    // scaleFactor converts from mm to target unit
+    constexpr float MM_MULTIPLIER = getLengthUnitMultiplier(LengthUnit::MILLIMETER);
```
MaticTonin: Already defined in hpp, can be reused from there

src/pipeline/node/PointCloud.cpp
```diff
@@ -1,12 +1,284 @@
...
+        }
+    }
+}
+
+void PointCloud::Impl::computePointCloudDenseCPU(const uint8_t* depthData, std::vector<Point3f>& points) {
+    calcPointsChunkDense(depthData, points, 0, height);
+}
+
+void PointCloud::Impl::computePointCloudDenseCPUMT(const uint8_t* depthData, std::vector<Point3f>& points) {
+    unsigned int rowsPerThread = height / threadNum;
```
MaticTonin, line 217: if threadNum = 0, will result to crashing, so redefine that its 0 and warning that it was set to 0.

src/pipeline/node/PointCloud.cpp
```diff
@@ -1,12 +1,284 @@
...
+        }
+    }
+
+    if(logger) {
+        logger->debug("Transformed {} valid points (z > 0)", transformedCount);
+    }
+}
+
+void PointCloud::Impl::calcPointsChunkDense(const uint8_t* depthData, std::vector<Point3f>& points, unsigned int startRow, unsigned int endRow) {
+    float scale = scaleFactor;
```
MaticTonin: const float scale

src/pipeline/node/PointCloud.cpp
```diff
@@ -1,12 +1,284 @@
...
+        f.get();
+    }
+}
+
+void PointCloud::Impl::computePointCloudDenseGPU(const uint8_t* depthData, std::vector<Point3f>& points) {
+#ifdef DEPTHAI_ENABLE_KOMPUTE
+    std::vector<float> xyzOut;
+    xyzOut.resize(size * 3);
+
+    float scale = scaleFactor;
```
MaticTonin: const float scale

src/utility/ProtoSerialize.cpp
```diff
@@ -965,8 +967,16 @@ void setProtoMessage(PointCloudData& obj, const google::protobuf::Message* msg,
     obj.setMaxX(pcl->maxx());
     obj.setMaxY(pcl->maxy());
     obj.setMaxZ(pcl->maxz());
-    obj.setSparse(pcl->sparse());
     obj.setColor(pcl->color());
+    
+    // // Read sparse flag for backward compatibility, but don't use it
```
MaticTonin: Required, seems kinda wrongly placed

MaticTonin (changes requested):
Will require some polishing, but overall it looks OK.

src/pipeline/datatype/PointCloudData.cpp
pheec: Based on the conversation with @MaticTonin @CenekAlbl @moratom on Slack, we should not break the backward compatibility more than just deprecating the "sparse" field. So I left the boundingBox unchanged

include/depthai/pipeline/datatype/PointCloudConfig.hpp
pheec: This custom transformation will change its meaning when the imageTransformations-extrinsics are merged. It will define the transformation to the user's coordinate system, in which he wants the pointscloud.

For now this transformation means the transformation from the camera socket (from which was the depth generated) to the user's coordinate system.

include/depthai/pipeline/node/PointCloud.hpp
```diff
@@ -2,30 +2,109 @@
...
+        unsigned int width, height;
+        size_t size;
+        bool intrinsicsSet = false;
+        uint32_t threadNum = 2;
+
+        std::vector<std::vector<float>> extrinsics;
+        bool hasExtrinsics = false;
+
+        std::shared_ptr<spdlog::logger> logger;
+    };
```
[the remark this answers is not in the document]
pheec: I will leave it as it as due to tests implementation

examples/cpp/PointCloud/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+**B) Housing coordinate system** — uses housing calibration stored on the device.
+
+```cpp
+pc->setTargetCoordinateSystem(dai::HousingCoordinateSystem::VESA_A);
+// useSpecTranslation=false  →  use per-unit calibrated values instead of spec
+```
+
+Available targets: `CAM_A`…`CAM_J`, `FRONT_CAM_A`…`FRONT_CAM_J`, `VESA_A`…`VESA_J`, `IMU`.
+`useSpecTranslation` defaults to `true`.
```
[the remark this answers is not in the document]
pheec: Nope, we decided not to add it to the examples. We want the user to keep the default value there

MaticTonin (approved):
Looks fine by me.

bindings/python/src/pipeline/node/PointCloudBindings.cpp
```diff
@@ -31,13 +31,33 @@ void bind_pointcloud(pybind11::module& m, void* pCallstack) {
...
+            "inputColor", [](PointCloud& node) -> Node::Input* { return &node.getColorInput(); }, py::return_value_policy::reference_internal)
+#endif
         .def_readonly(
             "outputPointCloud", &PointCloud::outputPointCloud, DOC(dai, node, PointCloud, outputPointCloud), DOC(dai, node, PointCloud, outputPointCloud))
         .def_readonly(
             "passthroughDepth", &PointCloud::passthroughDepth, DOC(dai, node, PointCloud, passthroughDepth), DOC(dai, node, PointCloud, passthroughDepth))
         .def_readonly("initialConfig", &PointCloud::initialConfig, DOC(dai, node, PointCloud, initialConfig), DOC(dai, node, PointCloud, initialConfig))
-        .def("setNumFramesPool", &PointCloud::setNumFramesPool, DOC(dai, node, PointCloud, setNumFramesPool));
+        .def("setNumFramesPool", &PointCloud::setNumFramesPool, py::arg("numFramesPool"), DOC(dai, node, PointCloud, setNumFramesPool))
+        .def("setRunOnHost", &PointCloud::setRunOnHost, py::arg("runOnHost"), DOC(dai, node, PointCloud, setRunOnHost))
```
MaticTonin, line 46: py::args on a single value defined setter I think is not needed

bindings/python/src/pipeline/datatype/PointCloudDataBindings.cpp
```diff
@@ -169,5 +177,8 @@ void bind_pointclouddata(pybind11::module& m, void* pCallstack) {
         .def("setMaxX", &PointCloudData::setMaxX, DOC(dai, PointCloudData, setMaxX))
         .def("setMaxY", &PointCloudData::setMaxY, DOC(dai, PointCloudData, setMaxY))
         .def("setMaxZ", &PointCloudData::setMaxZ, DOC(dai, PointCloudData, setMaxZ))
-        .def("setInstanceNum", &PointCloudData::setInstanceNum, DOC(dai, PointCloudData, setInstanceNum));
+        .def("getInstanceNum", &PointCloudData::getInstanceNum, DOC(dai, PointCloudData, getInstanceNum))
+        .def("updateBoundingBox", &PointCloudData::updateBoundingBox, DOC(dai, PointCloudData, updateBoundingBox))
+        .def("getTransformation", [](PointCloudData& data) -> ImgTransformation& { return data.getTransformation(); }, py::return_value_policy::reference_internal)
```
MaticTonin: Maybe overload_cast instead of the lambda fuction, cleaner

`        .def("getTransformation", [](PointCloudData& data) -> ImgTransformation& { return data.getTransformation(); }, py::return_value_policy::reference_internal)`

include/depthai/pipeline/datatype/PointCloudData.hpp
```diff
@@ -191,6 +203,30 @@ class PointCloudData : public Buffer, public ProtoSerializable {
...
 
+    /**
+     * Retrieves image transformation data
+     */
+    ImgTransformation& getTransformation();
+
+    /**
+     * Retrieves image transformation data (const)
+     */
+    const ImgTransformation& getTransformation() const;
```
MaticTonin, line 209: DO we need both methods one const and one non const?

src/pipeline/node/PointCloud.cpp
```diff
@@ -16,5 +490,384 @@ void PointCloud::setNumFramesPool(int numFramesPool) {
...
+                          spdlog::level::level_enum level,
+                          const std::string& name,
+                          const std::vector<std::vector<float>>& m) {
+    logger->log(level, "{}:", name);
+    for(int i = 0; i < 4; i++) {
+        logger->log(level, "  [{:8.4f}, {:8.4f}, {:8.4f}, {:8.4f}]", m[i][0], m[i][1], m[i][2], m[i][3]);
+    }
+}
+
+// ── PointCloud helper methods ──
```
MaticTonin: To MatrixOps.cpp

MaticTonin (changes requested):
Small comments

bindings/python/src/pipeline/node/PointCloudBindings.cpp
pheec: My best friend told me not to do that:

The comment is technically correct that the code would compile and work without it, but it's a best practice to keep it.

include/depthai/pipeline/datatype/PointCloudData.hpp
pheec: non-const removed

MaticTonin (approved):

src/pipeline/node/PointCloud.cpp
```diff
@@ -16,5 +492,358 @@ void PointCloud::setNumFramesPool(int numFramesPool) {
...
+    pimpl->logger->info("PointCloud node started (colorMode={})", colorMode);
+
+    uint32_t currentEepromId = getParentPipeline().getEepromId();
+    auto latestConfig = initialConfig;
+
+    while(mainLoop()) {
+        if(outputPointCloud.getQueueConnections().empty() && outputPointCloud.getConnections().empty()) {
+            std::this_thread::sleep_for(std::chrono::milliseconds(100));
+            continue;
+        }
```
MaticTonin, line 717: Is it required to sleep for 100 miliseconds?

src/pipeline/node/PointCloud.cpp
```diff
@@ -16,5 +492,358 @@ void PointCloud::setNumFramesPool(int numFramesPool) {
...
+
+        // Get synced frames from MessageGroup
+        std::shared_ptr<MessageGroup> group;
+        {
+            auto blockEvent = this->inputBlockEvent();
+            group = inSync.get<MessageGroup>();
+        }
+
```
MaticTonin, line 726: if(group == nullptr) continue;

src/pipeline/node/PointCloud.cpp
```diff
@@ -16,5 +492,358 @@ void PointCloud::setNumFramesPool(int numFramesPool) {
...
+        // Check for runtime config update
+        auto newConfig = inputConfig.tryGet<PointCloudConfig>();
+        if(newConfig) {
+            latestConfig = newConfig;
+            initialized = false;
+        }
+
+        // Sync organized mode from config
+        keepOrganized = latestConfig->getOrganized();
```
MaticTonin: I would not do it as header defined value, rather go for
const bool organized = latestConfig->getOrganized()

and

processColorized(depthFrame, colorFrame, pc, organized)

so that the state is one loop ownership and could not be overwritten.

src/pipeline/node/PointCloud.cpp
```diff
@@ -16,5 +492,358 @@ void PointCloud::setNumFramesPool(int numFramesPool) {
...
+        }
+
+        if(hasTransformationChanged(*depthFrame)) initialized = false;
+
+        if(!initialized) {
+            initialize(depthFrame, latestConfig);
+        }
+
+        auto width = depthFrame->getWidth();
+        auto height = depthFrame->getHeight();
```
MaticTonin: const since they will not be reassigned.

src/pipeline/node/PointCloud.cpp
```diff
@@ -16,5 +492,358 @@ void PointCloud::setNumFramesPool(int numFramesPool) {
...
+    pimpl->logger->info("PointCloud::initialize() completed");
+}
+
+//------------------------------------------------------------------
+// Processing helpers (depth-only and colorized paths)
+//------------------------------------------------------------------
+
+void PointCloud::processDepthOnly(std::shared_ptr<ImgFrame> depthFrame, std::shared_ptr<PointCloudData> pc) {
+    auto width = depthFrame->getWidth();
+    auto height = depthFrame->getHeight();
```
MaticTonin: const auto width, height

src/pipeline/node/PointCloud.cpp
```diff
@@ -1,13 +1,489 @@
...
+
+    for(uint32_t t = 0; t < threadNum; ++t) {
+        unsigned int startRow = t * rowsPerThread;
+        unsigned int endRow = (t == threadNum - 1) ? height : (startRow + rowsPerThread);
+        futures.emplace_back(std::async(std::launch::async, processRows, startRow, endRow));
+    }
+
+    for(auto& f : futures) {
+        f.get();
+    }
```
MaticTonin: Mybe a comment that this waits for completion of all async tasks.

src/pipeline/node/PointCloud.cpp
```diff
@@ -1,13 +1,489 @@
...
+    switch(computeMethod) {
+        case ComputeMethod::CPU:
+            transformPointsCPU(points);
+            break;
+        case ComputeMethod::CPU_MT:
+            transformPointsCPU(points);
+            break;
+        case ComputeMethod::GPU:
+            transformPointsCPU(points);
+            break;
```
MaticTonin: No need for the switch if all are returned as same method.

src/pipeline/node/PointCloud.cpp
```diff
@@ -1,13 +1,489 @@
...
+    }
```
MaticTonin, line 99: No need for a swicth if all are computed by same method.

src/pipeline/node/PointCloud.cpp
```diff
@@ -1,13 +1,489 @@
...
+            p.y = y;
+            p.z = z;
+            transformedCount++;
+        }
+    }
+
+    if(logger) {
+        logger->debug("Transformed {} valid points (z > 0)", transformedCount);
+    }
+}
```
MaticTonin: Same function, same method, different inputs, rather make a template function imo:
```

template <typename PointT>
void PointCloud::Impl::transformPointsCPU(std::vector<PointT>& points) {
    // Both points and extrinsics translations are in the same unit (target unit)
    // No conversion needed - just apply the transformation directly

    if(logger) {
        logger->debug("Applying transformation to {} points", points.size());
    }

    size_t transformedCount = 0;
    for(auto& p : points) {
        if(p.z > 0.0f) {
            // Standard 4x4 transformation: R*p + t
            float x = extrinsics[0][0] * p.x + extrinsics[0][1] * p.y + extrinsics[0][2] * p.z + extrinsics[0][3];
            float y = extrinsics[1][0] * p.x + extrinsics[1][1] * p.y + extrinsics[1][2] * p.z + extrinsics[1][3];
            float z = extrinsics[2][0] * p.x + extrinsics[2][1] * p.y + extrinsics[2][2] * p.z + extrinsics[2][3];

            p.x = x;
            p.y = y;
            p.z = z;
            transformedCount++;
        }
    }

    if(logger) {
        logger->debug("Transformed {} valid points (z > 0)", transformedCount);
    }
}
```

src/pipeline/node/PointCloud.cpp
```diff
@@ -1,13 +1,489 @@
...
+            transformPointsCPU(points);
+            break;
+        case ComputeMethod::CPU_MT:
+            transformPointsCPU(points);
+            break;
+        case ComputeMethod::GPU:
+            transformPointsCPU(points);
+            break;
+    }
+}
```
MaticTonin: Same here, rather a template function:
```
template <typename PointT>
void PointCloud::Impl::applyTransformation(std::vector<PointT>& points) {
    if(!hasExtrinsics) {
        if(logger) {
            logger->debug("No extrinsics set, skipping transformation");
        }
        return;
    }

    if(logger) {
        logger->debug("Applying coordinate system transformation");
    }

    switch(computeMethod) {
        case ComputeMethod::CPU:
            transformPointsCPU(points);
            break;
        case ComputeMethod::CPU_MT:
            transformPointsCPU(points);
            break;
        case ComputeMethod::GPU:
            transformPointsCPU(points);
            break;
    }
}
```

src/pipeline/node/PointCloud.cpp
```diff
@@ -1,13 +1,489 @@
...
+    sparsePoints.reserve(densePoints.size() / 2);
+
+    for(const auto& p : densePoints) {
+        if(p.z > 0.0f) {
+            sparsePoints.push_back(p);
+        }
+    }
+
+    return sparsePoints;
+}
```
MaticTonin: Same template function probably best to use template, less prone to errror since its Impl internal method.

src/pipeline/node/PointCloud.cpp
```diff
@@ -1,13 +1,489 @@
...
+            break;
+        case ComputeMethod::GPU:
+            transformPointsCPU(points);
+            break;
+    }
+}
+
+std::vector<Point3f> PointCloud::Impl::filterValidPoints(const std::vector<Point3f>& densePoints) {
+    std::vector<Point3f> sparsePoints;
+    sparsePoints.reserve(densePoints.size() / 2);
```
MaticTonin, line 110: Comment why specificaly / 2 is reserved, probably does not matter, but just for readibility in future.

src/pipeline/node/PointCloud.cpp
```diff
@@ -16,5 +492,358 @@ void PointCloud::setNumFramesPool(int numFramesPool) {
...
+        pc->setWidth(static_cast<unsigned int>(points.size()));
+        pc->setHeight(1);
+    }
+
+    pc->setPoints(std::move(points));
+}
+
+void PointCloud::processColorized(std::shared_ptr<ImgFrame> depthFrame, std::shared_ptr<ImgFrame> colorFrame, std::shared_ptr<PointCloudData> pc) {
+    auto width = depthFrame->getWidth();
+    auto height = depthFrame->getHeight();
```
MaticTonin: const auto width, height

src/pipeline/node/PointCloud.cpp
```diff
@@ -16,5 +492,358 @@ void PointCloud::setNumFramesPool(int numFramesPool) {
...
+        pimpl->logger->warn("PointCloud: color frame type ({}) is not RGB888i -- skipping colorization",
+                            static_cast<int>(colorFrame->getType()));
+        colorValid = false;
+    }
+
+    if(!colorValid) {
+        // Fall back to depth-only when color validation fails
+        processDepthOnly(depthFrame, pc);
+        return;
+    }
```
MaticTonin: If you check if color is valid at line 631 and immediately return only processDepthOnly, you avoid all the additional tracking of if color is valid or not:

```
`if(colorFrame->getWidth() != width || colorFrame->getHeight() != height) {
    pimpl->logger->warn("PointCloud: color frame size ({}x{}) does not match depth ({}x{}) -- skipping colorization",
                        colorFrame->getWidth(), colorFrame->getHeight(), width, height);
    processDepthOnly(depthFrame, pc);
    return;
}

if(colorFrame->getType() != ImgFrame::Type::RGB888i) {
    pimpl->logger->warn("PointCloud: color frame type ({}) is not RGB888i -- skipping colorization",
                        static_cast<int>(colorFrame->getType()));
    processDepthOnly(depthFrame, pc);
    return;
}`

```
and then you dont require to have colorValid argument.

src/pipeline/datatype/PointCloudData.cpp
```diff
@@ -141,8 +165,56 @@ PointCloudData& PointCloudData::setMaxZ(float val) {
...
+        for(std::size_t i = 0; i < count; ++i) {
+            update(pts[i].x, pts[i].y, pts[i].z);
+        }
+    } else {
+        const auto count = rawSize / sizeof(Point3f);
+        const auto* pts = reinterpret_cast<const Point3f*>(rawData);
+        for(std::size_t i = 0; i < count; ++i) {
+            update(pts[i].x, pts[i].y, pts[i].z);
+        }
+    }
```
MaticTonin: Nitpick:
Mybe a template, but here is simple duplication.

src/pipeline/node/PointCloud.cpp
```diff
@@ -16,5 +492,358 @@ void PointCloud::setNumFramesPool(int numFramesPool) {
...
+        if(latestEepromId > currentEepromId) {
+            pimpl->logger->debug("Calibration data changed (ID: {} -> {}), reinitializing...", currentEepromId, latestEepromId);
+            initialized = false;
+            currentEepromId = latestEepromId;
+        }
+
+        if(hasTransformationChanged(*depthFrame)) initialized = false;
+
+        if(!initialized) {
+            initialize(depthFrame, latestConfig);
```
MaticTonin: since you are either way dereferencing in initialize, you dont need to pass the shared pointer, but rather a refrence to exsisting object.

`void initialize(const ImgFrame& depthFrame, const PointCloudConfig& config)`

initialize(*depthFrame, *latestConfig);

src/pipeline/node/PointCloud.cpp
```diff
@@ -16,5 +492,358 @@ void PointCloud::setNumFramesPool(int numFramesPool) {
...
+        auto newConfig = inputConfig.tryGet<PointCloudConfig>();
+        if(newConfig) {
+            latestConfig = newConfig;
+            initialized = false;
+        }
+
+        // Sync organized mode from config
+        keepOrganized = latestConfig->getOrganized();
+
+        uint32_t latestEepromId = getParentPipeline().getEepromId();
```
MaticTonin: const unit32_t latestEepromId

src/pipeline/node/PointCloud.cpp
```diff
@@ -16,5 +492,358 @@ void PointCloud::setNumFramesPool(int numFramesPool) {
...
+// ── PointCloud helper methods ──
+
+void PointCloud::setIntrinsicsFromFrame(const ImgFrame& frame) {
+    auto width = frame.getWidth();
+    auto height = frame.getHeight();
+    auto intrinsics = frame.transformation.getIntrinsicMatrix();
+    float fx = intrinsics[0][0];
+    float fy = intrinsics[1][1];
+    float cx = intrinsics[0][2];
+    float cy = intrinsics[1][2];
```
MaticTonin: Nitpick: const for clarity

src/pipeline/node/PointCloud.cpp
```diff
@@ -16,5 +492,358 @@ void PointCloud::setNumFramesPool(int numFramesPool) {
...
+            } else {
+                pimpl->logger->info("No coordinate system transformation applied (identity)");
+            }
+            break;
+        }
+    }
+
+    // Apply the final transform
+    if(T_final) {
+        logMatrix4x4(pimpl->logger, spdlog::level::info, "T_final", *T_final);
```
MaticTonin: Delete, since in setExtrinsics you are already logging it.

src/pipeline/node/PointCloud.cpp
MaticTonin, line 1: Nitpick: Generally I would change order of functions, so you have:
- apply, get methods inside Impl
- CPU GPU implementation inside Impl
- intrinsics Extrinsics setters inside Impl

examples/python/PointCloud/point_cloud.py
MaticTonin, line 1: point_cloud.py

examples/python/PointCloud/point_cloud_showcase.py
MaticTonin, line 1: pointcloud_showcase.py

examples/python/PointCloud/point_cloud_visualizer.py
MaticTonin, line 1: pointcloud_visualizer.py

MaticTonin (changes requested):
Overall comments, when resolved, can be merged.

src/pipeline/node/PointCloud.cpp
pheec: it avoids expensive point cloud computation when nobody is consuming the output

src/pipeline/node/PointCloud.cpp
pheec: Not a legit comment for these lines — adding a comment about async task completion here would be misleading since no async work happens at this point.

src/pipeline/node/PointCloud.cpp
pheec: True, but I would still keep it

bindings/python/src/pipeline/datatype/PointCloudConfigBindings.cpp
```diff
@@ -19,6 +18,11 @@ void bind_pointcloudconfig(pybind11::module& m, void* pCallstack) {
     // py::class_<RawPointCloudConfig, RawBuffer, std::shared_ptr<RawPointCloudConfig>> rawConfig(m, "RawPointCloudConfig", DOC(dai, RawPointCloudConfig));
     py::class_<PointCloudConfig, Py<PointCloudConfig>, Buffer, std::shared_ptr<PointCloudConfig>> config(m, "PointCloudConfig", DOC(dai, PointCloudConfig));
 
+    py::enum_<PointCloudConfig::CoordinateSystemType>(config, "CoordinateSystemType")
+        .value("NONE", PointCloudConfig::CoordinateSystemType::NONE)
```
JakubFara: Whay it NONE coordinate system? The name seems to me confusing

examples/cpp/PointCloud/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
...
+
+## function: dai_add_example(example_name example_src enable_test use_pcl)
+## function: dai_set_example_test_labels(example_name ...)
+
+dai_add_example(PointCloud PointCloud.cpp OFF OFF)
+dai_add_example(PointCloudShowcase PointCloudShowcase.cpp OFF OFF)
+
+if(DEPTHAI_ENABLE_REMOTE_CONNECTION)
+    dai_add_example(PointCloudVisualizer PointCloudVisualizer.cpp OFF OFF)
+endif()
```
JakubFara, line 12: Allow tests? OFF -> ON

examples/cpp/PointCloud/PointCloudShowcase.cpp
```diff
@@ -0,0 +1,196 @@
...
+
+        pipeline.start();
+
+        // Wait for auto-exposure to settle and stereo depth to stabilize
+        std::cout << "Waiting for auto-exposure to settle...\n";
+        std::this_thread::sleep_for(std::chrono::seconds(1));
+
+        // Drain stale frames that arrived during warm-up
+        while(qSparse->tryGet<dai::PointCloudData>()) {
+        }
```
JakubFara: can you use `tryGetAll`?

JakubFara: Consider storing the queues to std::vector< { queue, "info" ... } > and then execute all the cases in a for loop?

examples/python/CMakeLists.txt
```diff
@@ -376,3 +376,10 @@ dai_set_example_test_labels(neural_assisted_stereo ondevice rvc4 ci)
 # Ips output
 add_python_example(camera_isp Camera/camera_isp.py)
 dai_set_example_test_labels(camera_isp rvc2_all rvc4 ci)
+
+## PointCloud
+add_python_example(pointcloud_showcase PointCloud/point_cloud_showcase.py)
+dai_set_example_test_labels(pointcloud_showcase ondevice rvc4 ci)
+
+add_python_example(pointcloud_visualizer PointCloud/point_cloud_visualizer.py)
+dai_set_example_test_labels(pointcloud_visualizer ondevice rvc4 ci nowindows)
```
JakubFara, line 388: add also `point_cloud.py` example

include/depthai/pipeline/datatype/PointCloudConfig.hpp
```diff
@@ -1,64 +1,125 @@
...
+    // true  = keep all width*height points (organized)
+    bool organized = false;
 
     std::array<std::array<float, 4>, 4> transformationMatrix = {{{1, 0, 0, 0}, {0, 1, 0, 0}, {0, 0, 1, 0}, {0, 0, 0, 1}}};
 
+    LengthUnit lengthUnit = LengthUnit::MILLIMETER;
+
+   public:
+    enum class CoordinateSystemType : uint8_t {
+        NONE,           ///< No transformation (camera coordinates)
```
JakubFara: `CoodrinateSystemType::NONE` seems weird, can you change it to something like DEFAULT?

include/depthai/pipeline/node/PointCloud.hpp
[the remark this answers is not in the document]
JakubFara: I would rather consider Imp being in "private" (in src) .hpp. Then you can include private header in the tests but not keep it in the public.

Or you can create PointCloudTest class and set it as a friend.

Or, at least, make it protected.

include/depthai/pipeline/node/PointCloud.hpp
```diff
@@ -37,11 +131,35 @@ class PointCloud : public DeviceNodeCRTP<DeviceNode, PointCloud, PointCloudPrope
...
+     * Get the optional color input for colorized point clouds.
+     * Lazily creates the Sync entry so that depth-only mode works
+     * without Sync waiting for a color frame that never arrives.
+     *
+     * Link an aligned color image (RGB888i, same dimensions as depth)
+     * to this input to enable colored point cloud output.
      */
-    Input inputDepth{*this, {"inputDepth", DEFAULT_GROUP, false, 4, {{{DatatypeEnum::ImgFrame, false}}}, true}};
+    Input& getColorInput();
+#endif
```
JakubFara, line 162: Only `Input& inputDepth` should not be compiled on RVC4.

src/pipeline/node/PointCloud.cpp
MaticTonin: Ok

include/depthai/pipeline/node/PointCloud.hpp
```diff
@@ -53,11 +171,80 @@ class PointCloud : public DeviceNodeCRTP<DeviceNode, PointCloud, PointCloudPrope
...
+    /**
+     * Use single-threaded CPU for processing
+     */
+    void useCPU();
+
+    /**
+     * Use multi-threaded CPU for processing
+     */
+    void useCPUMT(uint32_t numThreads = 2);
+
```
JakubFara, line 196: Is there a difference between `useCPU()` and `useCPUMT(1)` ?

src/pipeline/node/PointCloud.cpp
MaticTonin: Ok

MaticTonin (approved):
LGTM.

include/depthai/pipeline/node/PointCloud.hpp
```diff
@@ -53,11 +171,80 @@ class PointCloud : public DeviceNodeCRTP<DeviceNode, PointCloud, PointCloudPrope
...
+     * Point cloud will be transformed to this housing coordinate system
+     * @param housingCS Target housing coordinate system
+     * @param useSpecTranslation Whether to use spec translation (default: true)
+     */
+    void setTargetCoordinateSystem(HousingCoordinateSystem housingCS, bool useSpecTranslation = true);
+
+    bool runOnHost() const override;
+
+#ifndef DEPTHAI_INTERNAL_DEVICE_BUILD_RVC4
+    void buildInternal() override;
```
JakubFara, line 221: Can this be private and set Pipeline as a friend?

include/depthai/pipeline/node/PointCloud.hpp
```diff
@@ -53,11 +171,80 @@ class PointCloud : public DeviceNodeCRTP<DeviceNode, PointCloud, PointCloudPrope
...
+
+   private:
+    Pimpl<Impl> pimplPointCloud;
+
+    /// Private input receiving synced MessageGroup from Sync subnode
+    Input inSync{*this, {"inSync", DEFAULT_GROUP, false, 0, {{DatatypeEnum::MessageGroup, true}}}};
+
+    void run() override;
+    void initialize(const ImgFrame& depthFrame, const PointCloudConfig& config);
+    bool hasTransformationChanged(const ImgFrame& frame);
```
JakubFara, line 232: make this function const

include/depthai/pipeline/node/PointCloud.hpp
```diff
@@ -37,11 +131,35 @@ class PointCloud : public DeviceNodeCRTP<DeviceNode, PointCloud, PointCloudPrope
...
+    /**
+     * Get the optional color input for colorized point clouds.
+     * Lazily creates the Sync entry so that depth-only mode works
+     * without Sync waiting for a color frame that never arrives.
+     *
+     * Link an aligned color image (RGB888i, same dimensions as depth)
+     * to this input to enable colored point cloud output.
      */
-    Input inputDepth{*this, {"inputDepth", DEFAULT_GROUP, false, 4, {{{DatatypeEnum::ImgFrame, false}}}, true}};
+    Input& getColorInput();
```
JakubFara, line 161: make this function const

src/pipeline/datatype/PointCloudConfig.cpp
```diff
@@ -9,30 +9,67 @@ void PointCloudConfig::serialize(std::vector<std::uint8_t>& metadata, DatatypeEn
     datatype = DatatypeEnum::PointCloudConfig;
 }
 
-bool PointCloudConfig::getSparse() const {
-    return sparse;
+bool PointCloudConfig::getOrganized() const {
+    return organized;
 }
```
JakubFara, line 14: maybe `isOrganised` is a better name

src/pipeline/datatype/PointCloudConfig.cpp
```diff
@@ -9,30 +9,67 @@ void PointCloudConfig::serialize(std::vector<std::uint8_t>& metadata, DatatypeEn
...
+    return targetCameraSocket;
+}
+
+HousingCoordinateSystem PointCloudConfig::getTargetHousingCS() const {
+    return targetHousingCS;
+}
+
+bool PointCloudConfig::getUseSpecTranslation() const {
+    return useSpecTranslation;
+}
```
JakubFara, line 73: `isSpec`Translation ?

src/pipeline/datatype/PointCloudData.cpp
```diff
@@ -141,8 +165,56 @@ PointCloudData& PointCloudData::setMaxZ(float val) {
...
+PointCloudData& PointCloudData::updateBoundingBox() {
+    const auto* rawData = data->getData().data();
+    const auto rawSize = data->getData().size();
+
+    bool foundValid = false;
+    float mnX = 0.f, mnY = 0.f, mnZ = 0.f;
+    float mxX = 0.f, mxY = 0.f, mxZ = 0.f;
+
+    auto update = [&](float x, float y, float z) {
+        if(!foundValid) {
```
JakubFara, line 183: Can you set `mnX/Y/Z = infinity` `mxX/Y/Z = - infinity` and remove `foundValid` and if infinity is still there afterwards then return 0?

src/pipeline/datatype/PointCloudData.cpp
```diff
@@ -141,8 +165,56 @@ PointCloudData& PointCloudData::setMaxZ(float val) {
...
+        }
+    };
+
+    auto updatePoints = [&]<typename PointT>() {
+        const auto count = rawSize / sizeof(PointT);
+        const auto* pts = reinterpret_cast<const PointT*>(rawData);
+        for(std::size_t i = 0; i < count; ++i) {
+            update(pts[i].x, pts[i].y, pts[i].z);
+        }
+    };
```
JakubFara, line 204: `updatePoints` is not a good name, is should be something like `findMaxMin' / `findBBox`

include/depthai/pipeline/node/PointCloud.hpp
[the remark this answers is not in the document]
JakubFara: I can not see any test which uses this class

src/pipeline/node/PointCloud.cpp
```diff
@@ -16,5 +452,355 @@ void PointCloud::setNumFramesPool(int numFramesPool) {
...
+            auto configMat = matrix::toVecMatrix4x4(config.getTransformationMatrix());
+
+            if(!matrix::isIdentity4x4(configMat)) {
+                pimpl->logger->info("Applying custom transform composed with frame extrinsics");
+                T_final = matrix::matMul(configMat, T_frame_to_ref);
+            } else if(!matrix::isIdentity4x4(T_frame_to_ref)) {
+                pimpl->logger->info("Applying frame extrinsics (T_frame_to_ref)");
+                T_final = T_frame_to_ref;
+            } else {
+                pimpl->logger->info("No coordinate system transformation applied (identity)");
```
JakubFara: info -> debug

src/pipeline/node/PointCloud.cpp
```diff
@@ -16,5 +452,355 @@ void PointCloud::setNumFramesPool(int numFramesPool) {
...
+    CalibrationHandler calibHandler;
+    if(device) {
+        calibHandler = device->getCalibration();
+    } else {
+        calibHandler = getParentPipeline().getCalibrationData();
+    }
+
+    // Compute the target transform based on coordSystemType
+    // The final transform is: T_ref→target * T_frame→ref
+    std::optional<std::vector<std::vector<float>>> T_final;
```
JakubFara: set `T_final = std::nullopt;` 


https://en.cppreference.com/w/cpp/utility/optional.html : 
The object is initialized with/assigned from a value of type [std::nullopt_t](https://en.cppreference.com/w/cpp/utility/optional/nullopt_t.html) **or** an optional object that **does not contain a value**.

src/pipeline/node/PointCloud.cpp
```diff
@@ -16,5 +452,355 @@ void PointCloud::setNumFramesPool(int numFramesPool) {
...
+
+        // Get synced frames from MessageGroup
+        std::shared_ptr<MessageGroup> group;
+        {
+            auto blockEvent = this->inputBlockEvent();
+            group = inSync.get<MessageGroup>();
+        }
+        if(!group) continue;
+
+        auto depthFrame = std::dynamic_pointer_cast<ImgFrame>(group->group.at(depthInputName));
```
JakubFara: use this `auto depthFrame =  group->get<dai::ImgFrame>(name);`

src/pipeline/node/PointCloud.cpp
```diff
@@ -16,5 +452,355 @@ void PointCloud::setNumFramesPool(int numFramesPool) {
...
+        if(!depthFrame) {
+            pimpl->logger->warn("PointCloud: failed to extract depth frame from MessageGroup -- skipping");
+            continue;
+        }
+
+        std::shared_ptr<ImgFrame> colorFrame;
```
JakubFara: use this `auto colorFrame =  group->get<dai::ImgFrame>(name);`

src/pipeline/node/PointCloud.cpp
```diff
@@ -16,5 +452,355 @@ void PointCloud::setNumFramesPool(int numFramesPool) {
...
+            continue;
+        }
+
+        // Create PointCloudData
+        auto pc = std::make_shared<PointCloudData>();
+        pc->setTimestamp(depthFrame->getTimestamp());
+        pc->setTimestampDevice(depthFrame->getTimestampDevice());
+        pc->setSequenceNum(depthFrame->getSequenceNum());
+        pc->setInstanceNum(depthFrame->getInstanceNum());
+        pc->setTransformation(depthFrame->getTransformation());
```
JakubFara: The transformation is different if you transform the pointCloud?

JakubFara (commented):
LGTM, I have suggested some minor changes

examples/python/CMakeLists.txt
JakubFara: why only RVC4?

bindings/python/src/pipeline/datatype/PointCloudConfigBindings.cpp
pheec: When set to None, the pointcloud node applies transformation matrix set by the user in the config, which is by default identity

include/depthai/pipeline/node/PointCloud.hpp
JakubFara: `void useCPU(uint32_t numThreads = 1);` is more readable for me

include/depthai/pipeline/datatype/PointCloudConfig.hpp
pheec: Renamed ✅

src/pipeline/node/PointCloud.cpp
pheec: Thank you so much! This is actually huge.

PR #1736, opened 2026-03-27
HIL Testing Fixes

aljazdu:
## Purpose
This PR contains fixes for recently observed regressions in our HIL testing runs.

## Specification

The list of issues fixed in this PR:
- XLinkConnection::getFirstDevice() did not respect env. variables DEPTHAI_PROTOCOL & DEPTHAI_PLATFORM and defaulted to X_LINK_ANY_PROTOCOL & X_LINK_ANY_PLATFORM - which resulted in USB devices trying to be used when POE was meant to be used in a test
- No available devices found issue, which is the actual underlying issue causing the above to fail
- IMU tests failing (fixed by bumping RVC2 FW commit)
- Examples ran as tests fail in case a previous example disconnects the host from the device ungracefully

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

## Summary by CodeRabbit

* **Chores**
  * Updated device-side binary configuration to a newer commit.
  * Increased test execution timeout for more reliable runs.

* **Bug Fixes / Reliability**
  * Improved device connection selection to use default protocol/platform for better matching.
  * Enhanced test wrapper to pin target devices and perform time‑bounded crash-dump retrieval on failures.

* **Examples**
  * Added signal-based shutdown and orderly pipeline stop to multiple examples.
  * Made the edge detector example more robust with exception handling and clean shutdown.


moratom (approved):

tests/src/helpers/test_wrapper.cpp
```diff
@@ -92,15 +115,59 @@ int main(int argc, char* argv[]) {
...
+            } catch(const std::exception& ex) {
+                std::cout << "[wrapper] Post-failure crash dump sweep failed: " << ex.what() << std::endl;
+            }
+        }
+
         // 2 signifies that proc was killed by a timeout
-        if(retcode == 2) {
+        if(retcode == 2 && !crashDumpFound) {
             retcode = 0;
         }
```
moratom, line 172: If crashDumpFound I would fail the test even if the process itself returned 0

aljazdu: Currently crash dump sweep is performed only on processes which return != 0. 
```
        if(retcode == 2 && !crashDumpFound) {
            retcode = 0;
        }
```
this only makes sure that the processes which return retcode 2 are treated as successful - but only in case when there is no crash dump collected (this is a an additional safety as the test will fail due to Regex expression handling of crash dump logging)

aljazdu: Would you suggest trying to sweep for crash dumps for every case - even when processes return 0 - instead? Or should it be left as is?

moratom: Oh, did that cover all the test cases that were making crashdumps?
If yes, we can keep it as is and merge.

PR #1741, opened 2026-04-03
Fix DEPTHAI_VCPKG_INTERNAL_ONLY option

Amronos:
## Purpose
The `DEPTHAI_VCPKG_INTERNAL_ONLY` option had behaviour opposite to its description. I have corrected this behaviour and changed the default value to `OFF`.

## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

## Summary by CodeRabbit

* **Chores**
  * Adjusted default build configuration: changed how internal-only package handling and automatic selection of external interface libraries are determined, affecting which external libraries are used by default during builds.

### DIFF
--- a/cmake/depthaiOptions.cmake
+++ b/cmake/depthaiOptions.cmake
@@ -42,11 +42,11 @@ option(DEPTHAI_INSTALL "Enable install target for depthai-core targets" ON)
 
 # ---------- Dependency Management -------------
 option(DEPTHAI_BOOTSTRAP_VCPKG "Automatically bootstrap VCPKG" ON)
-option(DEPTHAI_VCPKG_INTERNAL_ONLY "Use VCPKG internally, but not for interface libraries" ON)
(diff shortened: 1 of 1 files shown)


moratom:
Thanks for the PR @Amronos !

I think the behavior was good, but the semantics are a bit confusing.

When `DEPTHAI_VCPKG_INTERNAL_ONLY` is set to `ON`  we only use VCPKG for all the private libraries, but not for the ones present in the headers.

For the public ones we either expect them to be installed on the system (for the larger ones, like PCL and OpenCV), or we use fetchcontent so we can easily install their headers ourselves (nlohman-json, xtensor, libnop).

So USE_EXTERNAL_INTERFACE_LIBS_DEFAULT=OFF when DEPTHAI_VCPKG_INTERNAL_ONLY=ON is set because we later bring our own nlohman-json, xtensor, etc compared to letting VCPKG install them for us and making them "external".

Amronos:
Oh, I think I get it now, thanks!

PR #1752, opened 2026-04-10
Fix four points transform

asahtik:
## Purpose

Fixes four points transform.

## Specification

Normally when using four points transform (in opencv) the output size is always provided. In depthai the output size can be automatically determined, however this can backfire for odd perspective / four points transformations because the calculated output size is very large. This PR clamps the automatically determined output size to 4000x3000. If a larger output is needed, the output size needs to be manually set.

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

## Summary by CodeRabbit

* **New Features**
  * Added a homography utility to compute 3x3 perspective transforms from four-point correspondences.
  * Improved four-point perspective handling with an automatic fallback when native optimizations aren't available.

* **Tests**
  * Added on-host and on-device tests validating homography math and four-point ImageManip behavior.

* **Chores**
  * Updated device-side configuration/version references.

* **Bug Fixes**
  * Ensured deterministic initialization for replay timing.


src/utility/ImageManipImpl.cpp
```diff
@@ -16,6 +16,9 @@
     #define _RESTRICT __restrict__
 #endif
 
+constexpr size_t MAX_AUTO_WIDTH = 4000;
+constexpr size_t MAX_AUTO_HEIGHT = 3000;
```
moratom, line 20: What's the idea behind these?

asahtik: The way ImageManip automatically sets the output size is by calculating the bounding box of the warped frame. With 4pt transform it's easy to warp the frame in such a way that the bounding box is very large which can crash the device. This limits the auto bounding box size to a reasonable value.

moratom (approved):

src/utility/ImageManipImpl.cpp
```diff
@@ -664,29 +670,29 @@ void dai::impl::getTransformImpl(const ManipOp& op,
                                o.dst[i].x *= width;
                                o.dst[i].y *= height;
```
aljazkonec1, line 671: width and height is calculated from getOuterRect() based on the input imageCorners. The assumption here is that imageCorners are non-normalized, does this assumption always hold?

aljazkonec1 (approved):
Thanks! Left one comment, otherwise looks good!

PR #1763, opened 2026-04-21
Remove implicit node instantiation in Python

aljazkonec1:
## Purpose

This PR removes the following deprecated feature:
``` 
with dai.Pipeline as pipeline:
    myNode = Node()
```
The only exception is (Threaded)HostNodes as they might require implicit pipeline in their init() methods as in the case of ParsingNeuralNetwork:
```
class ParsingNeuralNetwork(dai.node.ThreadedHostNode):
    def __init__(self, *args, **kwargs) -> None:
        super().__init__(*args, **kwargs) # Calls ThreadedHostNode.__init__(), which adds this node to the pipeline making getParentPipeline() available
        self._pipeline = self.getParentPipeline()
        self._nn = self._pipeline.create(dai.node.NeuralNetwork)
        self._parsers: Dict[int, BaseParser] = {}
```

## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

Tested by running a couple of oak-examples.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced error handling in node creation with more informative error messages and improved resource cleanup.

* **API Changes**
  * Simplified constructors for ColorCamera, DetectionNetwork, StereoDepth, and VideoEncoder nodes. Users now initialize these nodes without arguments and configure them explicitly using dedicated methods.

### DIFF
--- a/bindings/python/src/pipeline/PipelineBindings.cpp
+++ b/bindings/python/src/pipeline/PipelineBindings.cpp
@@ -43,6 +43,7 @@
 
 // depthai/
 #include <memory>
+#include <stdexcept>
 
 #include "depthai/properties/GlobalProperties.hpp"
 #include "depthai/utility/RecordReplay.hpp"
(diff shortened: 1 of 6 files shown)


aljazkonec1:
The ImplicitPipeline mechanism was not removed due to change in behavior in certain depthai-nodes where the [pipeline is needed at initializaiton](https://github.com/luxonis/depthai-nodes/blob/658240850afaacff3933e257f6f3611016521baa/depthai_nodes/node/parsing_neural_network.py#L13-L48). If we fully remove the mechanism we need to either compensate it with another mechanism that gets the pipeline or update depthai-nodes and oak-examples (+ there is a posibility this would break user nodes).

MaticTonin:
@aljazkonec1 based on comment above, should we just close this PR?

PR #1782, opened 2026-05-07
Buxfix/align runtime calibration change

JakubFara:
## Summary

  Fix runtime calibration handling in ImageAlign on branch bugfix/align_runtime_calibration_change.

  ## What changed

  - ImageAlign now detects runtime calibration changes and refreshes its cached calibration state.
  - When calibration changes during streaming, ImageAlign re-reads the latest inputAlignTo frame transform before
    rebuilding alignment maps.
  - Reset residual shift state during recalibration so stale horizontal offsets are not carried into the new alignment.
  - Added on-device regression coverage for runtime calibration behavior in image_align_node_test.cpp.

  ## Why

  Before this fix, calling device.setCalibration(...) during streaming could leave ImageAlign using stale alignment
  state. In practice this caused:

  - incorrect aligned output after runtime calibration updates
  - stale side-shift artifacts
  - mismatches between refreshed calibration data and cached align target geometry

  This change makes the align node switch coherently to the new calibration state without requiring a pipeline restart.

  ## Testing

  -  tests/src/ondevice_tests/image_align_node_test.cpp

## Summary by CodeRabbit

* **Bug Fixes**
  * Image alignment now robustly refreshes alignment data when calibration changes at runtime, resetting relevant state and pausing processing until an updated align-to frame is available so transforms and intrinsics remain consistent.

* **Tests**
  * Reworked and expanded tests to cover runtime calibration updates, verifying aligned depth outputs and intrinsics/transform refresh behavior after device calibration changes.

[Review Change Stack](https://app.coderabbit.ai/change-stack/luxonis/depthai-core/pull/1782)

### DIFF
--- a/src/pipeline/node/ImageAlign.cpp
+++ b/src/pipeline/node/ImageAlign.cpp
@@ -353,10 +353,42 @@ void ImageAlign::run() {
     auto latestConfig = initialConfig;
 
     int previousShiftFactor = 0;
+    bool refreshAlignToTransform = false;
 
     ImgTransformation inputAlignToTransform;
     ImgFrame inputAlignToImgFrame;
     uint32_t currentEepromId = getParentPipeline().getEepromId();
+
+    auto updateAlignToData = [&](const std::shared_ptr<ImgFrame>& inputAlignToImg) {
+        inputAlignToImgFrame = *inputAlignToImg;
+
+        auto alignTransform = inputAlignToImg->transformation;
+        const auto alignToDistortion = alignTransform.getDistortionCoefficients();
+        const bool hasDistortion = std::any_of(alignToDistortion.begin(), alignToDistortion.end(), [](float value) { return std::abs(value) > 0.0f; });
+        if(hasDistortion) {
+            logger->warn(
+                "The input connected to inputAlignTo is distorted. The aligned image will still be undistorted, meaning it won't be perfectly "
+                "aligned.");
+        }
+
+        alignTo = static_cast<CameraBoardSocket>(inputAlignToImg->getInstanceNum());
+        if(alignWidth == 0 || alignHeight == 0) {
+            alignWidth = inputAlignToImg->getWidth();
+            alignHeight = inputAlignToImg->getHeight();
+        }
+
+        auto alignTransformForIntrinsics = alignTransform;
+        auto [alignTransformWidth, alignTransformHeight] = alignTransformForIntrinsics.getSize();
+        if(static_cast<int>(alignTransformWidth) != alignWidth || static_cast<int>(alignTransformHeight) != alignHeight) {
+            float scaleX = static_cast<float>(alignWidth) / static_cast<float>(alignTransformWidth);
+            float scaleY = static_cast<float>(alignHeight) / static_cast<float>(alignTransformHeight);
+            alignTransformForIntrinsics.addScale(scaleX, scaleY);
+            alignTransformForIntrinsics.setSize(alignWidth, alignHeight);
+        }
+
+        alignSourceIntrinsics = alignTransformForIntrinsics.getIntrinsicMatrix();
+        inputAlignToTransform = alignTransformForIntrinsics;
+    };
 
     while(mainLoop()) {
         std::shared_ptr<ImgFrame> inputImg = nullptr;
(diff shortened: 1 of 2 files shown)


src/pipeline/node/ImageAlign.cpp
```diff
@@ -451,10 +455,22 @@ void ImageAlign::run() {
         if(latestEepromId > currentEepromId) {
             logger->debug("EEPROM data changed (ID: {} -> {}), reconfiguring ...", currentEepromId, latestEepromId);
             calibrationSet = false;
+            previousShiftFactor = 0;
```
MaticTonin, line 458: Any reason for this to be now 0?

MaticTonin (changes requested):
Only one question otherwise seems fine.

src/pipeline/node/ImageAlign.cpp
JakubFara: `previousShiftFactor` only tracks the shift already applied to map_x_1. When EEPROM changes we rebuild map_x_1 from fresh calibration, so that mesh is back in an unshifted state. Resetting `previousShiftFactor` to 0 ensures the next frame reapplies the full `constantShiftFactor` to the new mesh. If we keep the old value, the delta can become 0 and we skip shifting the rebuilt mesh.

JakubFara: L486
```
        int currentShiftFactor = constantShiftFactor - previousShiftFactor;

        if(currentShiftFactor != 0) {
            shiftMesh(map_x_1, -currentShiftFactor);
        }
```

MaticTonin:
Spoke to @aljazkonec1
He said, he already rewrote some parts of the ImageAlign node in the depthai-core and that he in his part completely changed that everything is image tranformation related.
And that we can merge this PR as it is, not add image tranformations, to be then all included in one PR, not half in one and half in other.
cc: @moratom for visibility.

src/pipeline/node/ImageAlign.cpp
MaticTonin: Oh, didnt see that, thanks

src/pipeline/node/ImageAlign.cpp
```diff
@@ -451,10 +455,22 @@ void ImageAlign::run() {
...
+            refreshAlignToTransform = true;
```
aljazkonec1, line 459: Do we need refreshAlignToTransform? To me it looks like the inverse of calibrationSet and they are inherently linked (when refreshAlignToTransform == True then setCalibration has to be false to properly update the maps). The updateAlignToData lambda has a capture by reference for setCalibration bool and could have the same logic as extractCalibrationData where `if (calibrationSet) return;`.

src/pipeline/node/ImageAlign.cpp
```diff
@@ -451,10 +455,22 @@ void ImageAlign::run() {
...
             calibHandler = pipeline.getCalibrationData();
             currentEepromId = latestEepromId;
         }
 
+        if(refreshAlignToTransform) {
+            if(auto latestAlignToImg = inputAlignTo.tryGet<ImgFrame>()) {
+                updateAlignToData(latestAlignToImg);
```
aljazkonec1, line 466: This is going to be unpredictable behaviour. Lets say that the eeprom ID got updated but an ImgFrame with the new ImgTransformation hasn`t gotten to the inputAlignTo queue (eg. Camera hasnt generated a new ImgFrame yet), then latestAlignToImg will be a message with the old ImgTransformation. So the update wont actually update the ImgTransformation.

I'm guessing this happens more often if you dont have camera node connected directly to the ImageAling node but have some other nodes in between that increase the time between between new inputAlignTo messages. IMO best solution is to store the previous ImgTransformation and only update when that changes. But I have that made in another branch where its also fallsback to previous impl in case no ImgTransformations available.

src/pipeline/node/ImageAlign.cpp
```diff
@@ -451,10 +455,22 @@ void ImageAlign::run() {
...
             logger->debug("EEPROM data changed (ID: {} -> {}), reconfiguring ...", currentEepromId, latestEepromId);
             calibrationSet = false;
+            previousShiftFactor = 0;
+            refreshAlignToTransform = true;
             calibHandler = pipeline.getCalibrationData();
             currentEepromId = latestEepromId;
         }
 
+        if(refreshAlignToTransform) {
+            if(auto latestAlignToImg = inputAlignTo.tryGet<ImgFrame>()) {
```
aljazkonec1, line 465: I would preffer `get<ImgFrame>()` instead of tryGet because this line will wait for new message to arrive instead of hoping that on the next iteration of this while loop there will be a message available.

aljazkonec1 (changes requested):
Thanks! Left some questions that need to be resolved.

MaticTonin:
@aljazkonec1 should we even work on this, or should we just close this PR and thing rather to genericAlignNode?

aljazkonec1:
My PR will be moved to a new "Align" node and all changes to ImageAlign will be reverted from my side. This PR can stay open for now and we can decide after Align node is merged to dai.

PR #1800, opened 2026-05-15
Added pre-merging of main eeprom calibration with attached cba eeprom calibrations

aljazdu:
## Purpose
This pull request updates the logic of reading calibration data from EEPROMs by merging the main board's EEPROM data with the EEPROM data of attached CBAs.

## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

## AI Usage
Assisted-by: Codex 5.4.

## Summary by CodeRabbit

* **New Features**
  * Automatic runtime merging of connected camera-board adapter calibrations into the device calibration before pipeline start.

* **Improvements**
  * Safer, thread-aware calibration handling that transparently prefers newer per-socket EEPROM data and persists merged results when applicable.

* **Tests**
  * Added on-host end-to-end tests validating calibration merge behavior and restoration across pipeline start.

[Review Change Stack](https://app.coderabbit.ai/change-stack/luxonis/depthai-core/pull/1800?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


bindings/python/src/DeviceBindings.cpp
MaticTonin: Rebase in develop, this already exists there.

src/device/DeviceBase.cpp
```diff
@@ -1820,6 +1827,53 @@ CalibrationHandler DeviceBase::getCalibration() {
...
+                if(const auto it = cbaData.cameraData.find(cameraSocket); it != cbaData.cameraData.end()) {
+                    eepromData.cameraData.emplace(cameraSocket, it->second);
+                    merged = true;
+                } else {
+                    pimpl->logger.warn("Skipping CBA calibration merge for {}: camera data not present for the specified CameraBoardSocket", cameraSocket);
+                }
+            } catch(const dai::EepromError& ex) {
+                pimpl->logger.warn("No readable CBA calibration on {}: {}", cameraSocket, ex.what());
+            }
+        }
```
MaticTonin: Is there any better way to go for try catch error?

src/device/DeviceBase.cpp
```diff
@@ -1820,6 +1827,53 @@ CalibrationHandler DeviceBase::getCalibration() {
...
+            if(cameraSocket == CameraBoardSocket::AUTO || eepromData.cameraData.find(cameraSocket) != eepromData.cameraData.end()) continue;
+
+            try {
+                const auto cbaData = readCalibration2(cameraSocket).getEepromData();
+                if(cbaData.batchTime < eepromData.batchTime) {
+                    pimpl->logger.info("Skipping CBA calibration merge for {}: Newer user calibration data found on device", cameraSocket);
+                    continue;
+                }
+                if(const auto it = cbaData.cameraData.find(cameraSocket); it != cbaData.cameraData.end()) {
+                    eepromData.cameraData.emplace(cameraSocket, it->second);
```
MaticTonin: What happens here with Extriniscs in cameraData? Since CBA eeprom has empty Extrinsics data, does it overwrite the cameraData calibrated Extrinsics on FFC EEPROM with the cameraData from CBA, which is empty?

tests/CMakeLists.txt
```diff
@@ -411,6 +411,11 @@ endif()
 dai_add_test(calibration_handler_test src/onhost_tests/calibration_handler_test.cpp)
 dai_set_test_labels(calibration_handler_test onhost ci)
 
+# Calibration CBA merge test
+dai_add_test(calibration_cba_merge_test src/onhost_tests/calibration_cba_merge_test.cpp)
+dai_set_test_labels(calibration_cba_merge_test onhost)
+set_property(TEST calibration_cba_merge_test APPEND PROPERTY ENVIRONMENT "DEPTHAI_ALLOW_FACTORY_FLASHING=868632271")
```
MaticTonin, line 425: @moratom is it ok for this to be exposed as property env? Good that its not for now set as ci, would enable CI as soon as we get the FFC in CI /CD

tests/src/onhost_tests/calibration_cba_merge_test.cpp
```diff
@@ -0,0 +1,216 @@
...
+    }
+
+    // Go over the compatible CBAs, store current flashed calibrations, followed by flashing with dummy data
+    CalibrationData currentEepromData(device);
+    for(const auto& cbaSocket : cbaSockets) {
+        currentEepromData.capture(cbaSocket);
+
+        device->flashEepromClear(cbaSocket);
+        device->flashCalibration(dai::CalibrationHandler::fromJson(loadCamDataCalibrationJson(cbaSocket)), cbaSocket);
+    }
```
MaticTonin, line 163: I would defenitly add here that check that as well some random Extrinsics are loaded.

MaticTonin (changes requested):
Overall nice, some nitpicks.

bindings/python/src/DeviceBindings.cpp
aljazdu: Done

src/device/DeviceBase.cpp
aljazdu: Currently `cameraData` is not overwritten in case the FFC already has a cameraData calibration for that socket. So regarding extrinsics, it will never be overwritten with empty extrinsics, because the overwrite would not happen in case calibration data would have been present already (for that socket).

So the exact behavior is:

- Existing FFC cameraData entry: not overwritten.

- Missing FFC cameraData entry: whole CameraInfo copied from CBA, including empty extrinsics if that is what CBA has.

- There is no partial merge where FFC intrinsics/extrinsics are preserved while only some CBA fields are filled in.

tests/CMakeLists.txt
aljazdu: In my opinion I would leave it here, considering we need this permission for this specific test in order for it to function at all. It only affects this one test and does not leak in others. 
In terms of  “secret exposure” perspective, this env var is not really secret anyway. The magic values are already hardcoded in our code and can be seen by the users going over `DeviceBase.cpp`

src/device/DeviceBase.cpp
aljazdu: Updated.

src/device/DeviceBase.cpp
```diff
@@ -1829,6 +1836,56 @@ CalibrationHandler DeviceBase::getCalibration() {
...
+            std::tie(success, errorMsg) = pimpl->rpcCall("setCalibration", eepromData).as<std::tuple<bool, std::string>>();
+            if(!success) {
+                throw std::runtime_error(errorMsg);
+            }
+        }
+        runtimeCalibrationMerged = true;
+    } catch(const std::exception& ex) {
+        pimpl->logger.warn("Failed to pre-merge CBA calibration before pipeline start: {}", ex.what());
+    }
+}
```
moratom: I think merging will have to happen on device, not on host, otherwise how will StereoDepth get the merged calibration, when no one on host will call "getCalibration"?

tests/src/onhost_tests/calibration_cba_merge_test.cpp
aljazdu: Updated the pre-defined calibration values (including the `Extrinsics`) with non default values. These are are tested using `compareCameraData` method in order to determine that the FFC calibration values (including `Extrinsics`) are not overwritten in case of existing CameraData on FFC.

src/device/DeviceBase.cpp
aljazdu: Would it make sense to move from the use of `readCalibration/2()` to `getCalibration()` for uses such as `getAvailableStereoPairs()`, which is the root cause of the issue, as now there is an option of a merge runtime calibration - which should be accessed with `getCalibration()` instead (in case of no merging the runtime calibration is of course not updated).

moratom: The root problem is there is no gurantee for `getAvailableStereoPairs` to be called either.

When StereoDepth node starts on device, it would now start with the non merged calibration if I understand correctly, while we'd want StereoDepth to have the right calibration from the beginning (so it needs to be pre-merged on device).

In theory we could still have it done on host, just make sure it's exectued before a pipeline start always, but I think it's better if this merging is done by device side, less room for error.

examples/cpp/VSLAM/rerun_node.hpp
```diff
@@ -67,7 +67,8 @@ class RerunNode : public dai::NodeCRTP<dai::node::ThreadedHostNode, RerunNode> {
                 auto image = imgFrame->getCvFrame();
                 cv::cvtColor(image, image, cv::COLOR_BGR2RGB);
                 rec.log("world/camera/image/rgb",
-                        rerun::Image(reinterpret_cast<const uint8_t*>(image.data), {imgFrame->getWidth(), imgFrame->getHeight()}, rerun::datatypes::ColorModel::RGB));
+                        rerun::Image(
+                            reinterpret_cast<const uint8_t*>(image.data), {imgFrame->getWidth(), imgFrame->getHeight()}, rerun::datatypes::ColorModel::RGB));
```
aljazkonec1, line 71: Does this pass clang format and style check? If yes, you can ignore this comment

aljazkonec1:
Thanks!
 I am testing out the impl and want to confirm the following behavior is expected (cc @MaticTonin and calibration team to confirm). Setup: RVC4 devkit with 1 new CBA and one old CBA and I flashed the board calibration with `{}` calibration using: `device.flashCalibration(calib)`. 
1) Running `examples/python/Calibration/calibration_dump_sockets.py` I get the new CBA calibration and a warning for missing EEPROM for the old CBA
2) Running `examples/python/Calibration/calibration_dump.py` I dont get any camerData in user calibration
3) Chaning to function `getCalibration` instead of `readCalibration2` in `calibration_dump.py`. I get the new CBA in "cameraData"
5) I then connected a second new CBA, `calibration_dump_sockets.py` shows all three CBAs, running `calibration_dump.py` (with getCalibration() funciton) shows both new CBAs in "cameraData" 
6) I then flashed one CBA with some dummy intrinsics using
 ```
calib = dai.CBACalibrationHandler.fromJson(jsonFile)
device.flashCBACalibration(calib, dai.CameraBoardSocket.CAM_F)
```
And the flash was successful and survived device cycle while leaving factory flash as-is.
7) I also repeated the same steps but flashed CAM_A, B and C to device Eeprom and that calibration took priority over the CBAs
8) I ran same sequence on RVC2 FFC 4P with the same results. 

@MaticTonin can you or your team test out behavior from your side? Other than that you have the green light from my side!

File: tests/src/onhost_tests/calibration_cba_merge_test.cpp [hunk 7]
```diff
@@ -0,0 +1,256 @@
...
+}
+
+static bool compareCameraData(const dai::CameraInfo& first, const dai::CameraInfo& second) {
+    return (first.width == second.width) && (first.height == second.height) && (first.lensPosition == second.lensPosition)
+           && (first.intrinsicMatrix == second.intrinsicMatrix) && (first.distortionCoeff == second.distortionCoeff)
+           && (first.extrinsics.isEqualExtrinsics(second.extrinsics)) && (first.specHfovDeg == second.specHfovDeg) && (first.cameraType == second.cameraType);
+}
+
+TEST_CASE("CBA calibrations are merged with the main board's calibration") {
+    std::shared_ptr<dai::Device> device = std::make_shared<dai::Device>();
```
JakubFara, line 142: Should these tests be ion `ondevice_tests` since it uses device?

tests/src/onhost_tests/calibration_cba_merge_test.cpp
```diff
@@ -0,0 +1,256 @@
...
+        REQUIRE(compareCameraData(mainCameraData, cbaCameraData));
+    }
+
+    // Starting the pipeline should keep the already prepared runtime calibration intact
+    dai::Pipeline pipeline(device);
+    auto logger = pipeline.create<dai::node::SystemLogger>();
+    auto outputQueue = logger->out.createOutputQueue(1, false);
+    pipeline.start();
+    dai::CalibrationHandler mergedCalibration = device->getCalibration();
+    for(const auto& cbaSocket : cbaSockets) {
```
JakubFara, line 188: We should also check that the non-CBA sockets are untouched

tests/src/onhost_tests/calibration_cba_merge_test.cpp
```diff
@@ -0,0 +1,256 @@
...
+        REQUIRE(mergedCalibration.hasCameraCalibration(cbaSocket));
+
+        const auto mainCameraData = mergedCalibration.getEepromData().cameraData.at(cbaSocket);
+        const auto cbaCameraData = device->readCBACalibration2(cbaSocket).getEepromData().cameraData.at(dai::CameraBoardSocket::CBA);
```
JakubFara, line 192: I think we should avoid using `readCBACalibration` function in this case. If we flash to CBA and to device eeprom the same but wrong values calibration we don't notice that. You know that there should be `loadCamDataCalibrationJson()`

tests/src/onhost_tests/calibration_cba_merge_test.cpp
JakubFara, line 1: We are testing only` flashCalibration` what about `setCalibration`?

JakubFara: cases like this:

- CAM_A -> don't have CBA
- CAM_B -> have CBA
- CAM_C -> have  CBA

I set calibration with extrinsics A->C.  Will CAM_B have correct CBA calibration?

aljazdu: The scope of the test is cba merging, as such we are testing only merging. `flashcalibration()` is "tested" only as a side product of flashing the dummy test data - which guarantees the test repeatability.

JakubFara:
Can we also test that the extrinsics after CBA merging are always empty `std::vector`?

tests/src/onhost_tests/calibration_cba_merge_test.cpp
JakubFara: If the cba marging in both cases is the same I agree.

aljazkonec1:
Just remembered: lets switch `readCalibration2` in calibration_dump.py to `getCalibration()` instead

aljazkonec1 (approved):

MaticTonin (approved):
LGTM!

PR #1807, opened 2026-05-19
Telemetry

jakgra:
## Purpose
Add anonymous usage telemetry so we can improve the library

## Specification
Added telemetry events from depthai-core and device FW's

no breaking changes or risks

## Deployment Plan
None / not applicable

## Testing & Validation
added 2 tests + manually tested on both rvc2 and rvc4 with 1 and 2 devices. Also tested inside oak app. Tested both python and C++ usage.

## AI Usage
Assisted-by: Codex:gpt-5.5

## Release Notes

* **New Features**
  * Added optional anonymized telemetry collection; disable via `DEPTHAI_TELEMETRY=0` environment variable
  * Added `getOSVersion()` API to retrieve operating system version information
  * Added built-in node detection capability for pipeline nodes

* **Documentation**
  * Updated README with telemetry configuration details and new environment variables (`DEPTHAI_TELEMETRY`, `DEPTHAI_TELEMETRY_URL`, `DEPTHAI_TELEMETRY_API_KEY`)

* **Tests**
  * Added comprehensive telemetry test suite covering redirection and disable scenarios

[Review Change Stack](https://app.coderabbit.ai/change-stack/luxonis/depthai-core/pull/1807?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


bindings/python/src/pipeline/PipelineBindings.cpp
```diff
@@ -201,7 +201,7 @@ void PipelineBindings::bind(pybind11::module& m, void* pCallstack) {
                  d.wait();
              })
         //.def(py::init<const Pipeline&>())
-        .def("getDefaultDevice", &Pipeline::getDefaultDevice, DOC(dai, Pipeline, getDefaultDevice))
+        .def("getDefaultDevice", static_cast<std::shared_ptr<Device> (Pipeline::*)()>(&Pipeline::getDefaultDevice), DOC(dai, Pipeline, getDefaultDevice))
```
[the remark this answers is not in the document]
moratom: An issue from before, but makes sense to fix now.

How come we had to change the binding code here?

src/xlink/XLinkStream.cpp
```diff
@@ -49,9 +49,13 @@ XLinkStream& XLinkStream::operator=(XLinkStream&& other) {
...
-    // If streamId != invalid (eg. wasn't moved to another XLinkStream)
+    close();
+}
+
+void XLinkStream::close() {
     if(streamId != INVALID_STREAM_ID) {
         XLinkCloseStream(streamId);
+        streamId = INVALID_STREAM_ID;
     }
 }
```
moratom, line 56: Needed changes? Can we add a test that showcases how this is better>

src/utility/Telemetry.hpp
```diff
@@ -0,0 +1,52 @@
...
+
+std::string getTemporaryTelemetryHostId();
+std::string getTemporaryTelemetryDeviceId(const std::string& mxid);
+std::string getTelemetrySessionId();
+std::string getTelemetryHostOS();
+std::string getTelemetryHostOSVersion();
+bool isTelemetryEnabled();
+void setTelemetryUsesPython(bool value);
+void emitDepthaiTelemetryLoadEvent();
+
```
moratom: Any reason we don't make them part of the class & bind variables to a singleton instead of having global variables?

src/utility/Telemetry.cpp
```diff
@@ -0,0 +1,860 @@
...
+std::string getTelemetryHostOSImpl() {
+    if(!readEnv("OAKAGENT_APP_ID").empty() || !readEnv("OAKAGENT_APP_IDENTIFIER").empty()) {
+        return "oakapp";
+    }
+
+    const auto platformName = dai::platform::getOSPlatform();
+    if(platformName == "Windows") return "windows";
+    if(platformName == "Linux") return "linux";
+    if(platformName == "MacOS") return "mac";
+    return lowercase(platformName);
```
moratom, line 192: Nit: the first two cases would be handled by the last case, IMO last case is fine w/o the ifs.

src/device/DeviceBase.cpp
```diff
@@ -486,11 +540,143 @@ DeviceBase::DeviceBase(Config config, const DeviceInfo& devInfo) : deviceInfo(de
...
+    telemetryCreatedAt = std::chrono::steady_clock::now();
+    try {
+        tmpDeviceId = pimpl->rpcCallChecked<std::string>("getAnonymousTelemetryId");
+    } catch(const std::exception& ex) {
+        pimpl->logger.debug("Failed to get anonymous telemetry id from device: {}", ex.what());
+    }
+    if(tmpDeviceId.empty()) {
+        tmpDeviceId = utility::getTemporaryTelemetryDeviceId(deviceInfo.getDeviceId());
+    }
+    telemetryLifecycleStarted = true;
```
moratom, line 650: Is one expected for RVC2 and the other for RVC4?
If yes, I think going explicitly rather than a generic "catch" would be better

src/device/DeviceBase.cpp
```diff
@@ -486,11 +540,143 @@ DeviceBase::DeviceBase(Config config, const DeviceInfo& devInfo) : deviceInfo(de
...
+                std::lock_guard<std::mutex> lock(telemetryEventStreamMtx);
+                if(telemetryEventStream == stream) {
+                    telemetryEventStream.reset();
+                }
+            }
+            if(telemetryEventRunning) {
+                pimpl->logger.debug("Telemetry event thread exception caught: {}", ex.what());
+            }
+        }
+    }
```
moratom, line 617: Is this the expected flow on exit? Can we catch a more specific exception if yes?
If this fails for any reason on startup, it won't be caught.

src/pipeline/Pipeline.cpp
```diff
@@ -92,6 +95,70 @@ std::optional<PipelineAutoCalibrationMode> parseAutoCalibrationMode(std::string_
...
+        const auto propertyBytes = properties.get<std::vector<std::uint8_t>>();
+        if(propertyBytes.empty()) {
+            return nlohmann::json::object();
+        }
+        auto parsedProperties = nlohmann::json::parse(propertyBytes.begin(), propertyBytes.end());
+        if(parsedProperties.is_object()) {
+            return parsedProperties;
+        }
+    } catch(const std::exception&) {
+    }
```
moratom: I would avoid this if possible (at the very least log and/or ideally catch a more specific exception)

src/pipeline/Pipeline.cpp
```diff
@@ -92,6 +95,70 @@ std::optional<PipelineAutoCalibrationMode> parseAutoCalibrationMode(std::string_
...
+            std::find(utility::depthaiNodesNames.begin(), utility::depthaiNodesNames.end(), node.second.name) != utility::depthaiNodesNames.end();
+        if(!node.second.builtInNode && !isDepthaiNodesNode) {
+            node.second.name = "CUSTOM";
+        }
+    }
+    return schema;
+}
+
+bool redactTelemetryNodeProperties(const std::string& nodeName) {
+    return nodeName == "CUSTOM" || nodeName == "DetectionParser" || nodeName == "SegmentationParser";
```
moratom, line 110: Why do we want to redact DetectionParser and SegmentationParser?

src/modelzoo/Zoo.cpp
```diff
@@ -632,6 +719,9 @@ fs::path getModelFromZoo(
 
     if(internetIsAvailable) {
         responseJson = zooManager.fetchModelDownloadLinks();
+        if(utility::isTelemetryEnabled()) {
+            modelLoadedTelemetryProperties = zooManager.getModelLoadedTelemetryProperties(responseJson);
+        }
     }
```
moratom, line 725: Will we be able to match up with the node that it was used on? I think this would be good to see what parsers are used, etc.

moratom (commented):
Thanks, looking forward to this!

Left some comments & questions.

src/pipeline/Pipeline.cpp
jakgra: The properties leak nnarchive / zoo model stuff

src/utility/Telemetry.cpp
jakgra: ignored the nit

src/device/DeviceBase.cpp
jakgra: rvc2 doesn't throw but returns empty string, and rvc5 might also return empty string... . So the catch is only for error cases and the tmpDeviceId.empty() is meant for rvc2. So the above code makes sense at least to me... . We could change to warning() but I do not dare...

src/modelzoo/Zoo.cpp
jakgra: after sync, we decided it is out of scope for now.

src/pipeline/Pipeline.cpp
jakgra: fixed

src/device/DeviceBase.cpp
jakgra: made more explicit

src/utility/Telemetry.hpp
jakgra: made nicer

src/xlink/XLinkStream.cpp
jakgra: not needed. reverted to old version for safety's sake.

moratom (approved):

PR #1574, opened 2025-12-04
Fix: Build core without OpenCV support

aljazkonec1:
## Purpose

Build without opencv support was failing due to unhandled opencv calls when opencv support was disabled. 

## Specification

None / not applicable

## Dependencies & Potential Impact

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

Tested with
```
$ cmake -S . -B build -D'BUILD_SHARED_LIBS=ON' -D DEPTHAI_OPENCV_SUPPORT=OFF
$ cmake --build build --parallel 20
```

PR #1818, opened 2026-05-27
Enabling builds without OpenCV

gkienernf:
## Purpose
Enabling builds without OpenCV (`DEPTHAI_OPENCV_SUPPORT=OFF`)

## Specification
Building depthai-core with `-DDEPTHAI_OPENCV_SUPPORT=OFF` currently fails to
compile and link. (This was previously fixed in #1574 but has since regressed —
the OpenCV-off configuration isn't covered by CI.) Four issues are addressed:

- **`src/utility/RecordReplay.cpp`** — the no-OpenCV `getVideoSize()` fallback
  returned `std::tuple<size_t, size_t>`, but the declaration in
  `RecordReplayImpl.hpp` is `std::tuple<size_t, size_t, double>`, so it failed
  as an overload differing only by return type. Fallback signature corrected to
  match the declaration.
- **`src/utility/PipelineImplHelper.cpp`** — the `#else` (non-`DEPTHAI_MERGED_TARGET`)
  branch referenced a bare `recordConfig` instead of `pipeline->recordConfig`
  (undeclared identifier). Qualified correctly.
- **`include/depthai/pipeline/node/ToF.hpp`** — the `ToF` node group
  unconditionally embeds a `Subnode<ImageFilters>` (ImageFilters is an
  OpenCV-only node) and exposes `imageFilters->output` as its filtered `depth`
  output, pulling ImageFilters' vtable/symbols into the always-compiled `ToF`.
  The `ImageFilters` subnode, its `build`/link wiring, and the `imageFilters*`
  accessors are now guarded behind `DEPTHAI_HAVE_OPENCV_SUPPORT`. When OpenCV is
  unavailable, `ToF::depth` falls back to the raw `tofBase` depth output.
- **`src/pipeline/node/Rectification.cpp`** — `Rectification::getCalibrationData()`
  (which uses no OpenCV) was defined only inside the OpenCV-enabled block,
  leaving the always-emitted `Rectification` vtable with an undefined reference.
  Moved it out so it is always defined.

No impact on the default (OpenCV-enabled) build — all changes are inside
`#ifdef DEPTHAI_HAVE_OPENCV_SUPPORT` guards or are a no-op relocation of a
non-OpenCV method. No public API change in the OpenCV-on configuration.

Behavioral note for the OpenCV-off configuration only: with OpenCV disabled,
`ToF::depth` provides unfiltered (raw) depth instead of ImageFilters-filtered
depth, and the `imageFiltersInputConfig` / `imageFiltersNode` accessors are not
present — consistent with `Rectification::run()` already throwing without
OpenCV. No ABI break, since the OpenCV-off configuration did not previously
compile at all.

## Deployment Plan
Not applicable — source-only build fix, no runtime/rollout component. Rollback
is reverting the commit.

## Testing & Validation
Built `depthai-core` on macOS arm64 (Apple Silicon) with OpenCV **not installed
at all** and `-DDEPTHAI_OPENCV_SUPPORT=OFF`:
configures (OpenCV not found/disabled), compiles all translation units, and
links `libdepthai-core` with **zero OpenCV linkage** (verified via `otool -L`).

End-to-end validated against real hardware: built a downstream consumer
(RGB + StereoDepth + VideoEncoder pipeline) against the OpenCV-off
`libdepthai-core` and ran it on an OAK-D Lite — device boots, depth aligns to
RGB (1280x720), MJPEG-encoded pipeline runs, frames stream — all with no OpenCV
present.

## AI Usage
Assisted-by: Claude:claude-opus-4 clang-tidy

## Release Notes

* **New Features**
  * Video size retrieval now provides enhanced information.

* **Bug Fixes**
  * Fixed record and replay state management in holistic configurations.
  * Improved calibration data availability across different build scenarios.
  * Enhanced device compatibility when optional image processing features are unavailable.

[Review Change Stack](https://app.coderabbit.ai/change-stack/luxonis/depthai-core/pull/1818?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/include/depthai/pipeline/node/ToF.hpp
+++ b/include/depthai/pipeline/node/ToF.hpp
@@ -82,14 +82,24 @@ class ToF : public DeviceNodeGroup {
     ToF(const std::shared_ptr<Device>& device)
         : DeviceNodeGroup(device),
           rawDepth{tofBase->depth},
+#ifdef DEPTHAI_HAVE_OPENCV_SUPPORT
           depth{imageFilters->output},
+#else
+          depth{tofBase->depth},
+#endif
           amplitude{tofBase->amplitude},
           intensity{tofBase->intensity},
           phase{tofBase->phase},
           tofBaseInputConfig{tofBase->inputConfig},
+#ifdef DEPTHAI_HAVE_OPENCV_SUPPORT
           imageFiltersInputConfig{imageFilters->inputConfig},
-          tofBaseNode{*tofBase},
-          imageFiltersNode{*imageFilters} {}
+#endif
+          tofBaseNode{*tofBase}
+#ifdef DEPTHAI_HAVE_OPENCV_SUPPORT
+          ,
+          imageFiltersNode{*imageFilters}
+#endif
+    {}
 
     ~ToF() override;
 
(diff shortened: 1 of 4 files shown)


moratom:
Thanks for the PR @gkienernf !

Can you re-target to develop?
The code should already compile though ToF changes might still be needed.

gkienernf:
Thanks!! Sorry, this got filtered out in my dozens of GitHub notifs, so didn't catch it! @moratom

PR #1845, opened 2026-06-11
Fix timed out Camera FPS config test

aljazdu:
## Purpose
Camera FPS config test often timed out for RVC2 POE devices. This fix adds early stopping when enough frames are collected.

## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

## AI Usage

Assisted-by: /

## Summary by CodeRabbit

* **Tests**
  * Added a new on-device camera FPS/config validation test (`camera_fps_config_part_2_test`) to verify camera output queue/frame pool limits across platforms.
  * Updated platform-specific expectations for RVC4 queue size overrides and ensured behavior adapts when `DEPTHAI_AUTOCALIBRATION=OFF`.
* **Bug Fixes**
  * Improved crash-dump state handling after reconnects by correctly preparing the “handled” tracking based on whether a pending crash dump exists.

### DIFF
--- a/src/device/DeviceBase.cpp
+++ b/src/device/DeviceBase.cpp
@@ -1526,7 +1526,11 @@ void DeviceBase::monitorCallback(std::chrono::milliseconds watchdogTimeout, Prev
             if(reconnectionCallback) reconnectionCallback(ReconnectionStatus::RECONNECTED);
             pimpl->logger.warn("Reconnection successful\n");
             if(isCrashDumpCollectionEnabled()) {
-                crashed = hasCrashDump();
+                const bool hasPendingCrashDump = hasCrashDump();
+                crashed = hasPendingCrashDump;
+                if(!hasPendingCrashDump) {
+                    crashDumpHandled.store(false);
+                }
             }
         }
     } catch(const std::exception& ex) {
--- a/tests/src/ondevice_tests/pipeline/node/camera_fps_config_test.cpp
+++ b/tests/src/ondevice_tests/pipeline/node/camera_fps_config_test.cpp
@@ -175,11 +175,18 @@ TEST_CASE("Camera pool sizes") {
         }
         // Check stream still works after script node unblocks the Camera node
         while(std::chrono::duration<double>(std::chrono::steady_clock::now() - startTime).count() < timeToBlock + 30) {
+            bool gotEnoughFrames = true;
             for(int idx = 0; idx < outQueues.size(); ++idx) {
                 auto frame = outQueues[idx]->tryGet();
                 if(frame) {
                     ++outQueuesCounter[idx];
                 }
+                if(outQueuesCounter[idx] <= queueSize + 200) {
+                    gotEnoughFrames = false;
+                }
+            }
+            if(gotEnoughFrames) {
+                break;
             }
         }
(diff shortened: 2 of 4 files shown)


tests/src/ondevice_tests/pipeline/node/camera_fps_config_test.cpp
```diff
@@ -175,11 +175,18 @@ TEST_CASE("Camera pool sizes") {
...
                 if(frame) {
                     ++outQueuesCounter[idx];
                 }
+                if(outQueuesCounter[idx] <= queueSize + 200) {
+                    gotEnoughFrames = false;
+                }
+            }
+            if(gotEnoughFrames) {
+                break;
             }
```
moratom: I think this might defeat the purpuse of the test CC: @jakgra 

I would rather reduce the time to block for 30s if total runtime is a problem

moratom (dismissed):

tests/src/ondevice_tests/pipeline/node/camera_fps_config_test.cpp
aljazdu: The current test implementation asserts only that 200 frames were successfully sent, not asserting sustained streaming for the full 30s window (while those 30s of streaming do happen). The early break preserves that assertion and removes dead wait on the success path, not influencing the test checks in any way.

> I would rather reduce the time to block for 30s if total runtime is a problem

Can go with this, but would have to update the current assertion as well to actually achieve the mentioned purpose of the test.

src/device/DeviceBase.cpp
```diff
@@ -1526,7 +1526,11 @@ void DeviceBase::monitorCallback(std::chrono::milliseconds watchdogTimeout, Prev
             if(reconnectionCallback) reconnectionCallback(ReconnectionStatus::RECONNECTED);
             pimpl->logger.warn("Reconnection successful\n");
             if(isCrashDumpCollectionEnabled()) {
-                crashed = hasCrashDump();
+                const bool hasPendingCrashDump = hasCrashDump();
+                crashed = hasPendingCrashDump;
+                if(!hasPendingCrashDump) {
+                    crashDumpHandled.store(false);
+                }
```
aljazkonec1, line 1533: Was this also causing issues on RVC4?

aljazdu: It was not, as it seems. 
This is probably connected to the change in crash dump detection/collection in RVC2

aljazkonec1 (commented):
w.r.t. the 200 frame limit: the code does not reflect the spirit of the test. Instead of validating running for 30 seconds, it just checks for >200 recieved frames so I agree with what @aljazdu proposed to adapt.

IMO the entire test would need to be rewritten but lets make that a separate PR just because of higher prio items. 
Before merge, just test out splitting the test into two files, one with the first TEST_CASE and half of the queue sizes form second TEST_CASE and another test file with the other half of the queue sizes. This should stop the timeout from hitting without the 200 frame limit.

aljazkonec1 (approved):

aljazkonec1:
Can merge once POE and Crashdump tests pass

PR #1854, opened 2026-06-18
Fix PointCloud node: use calibrated translation from ImgTransformations extrinsics instead of specTranslation

pheec:
## Purpose
When the PointCloud node uses `HousingCoordinateSystem` as target, the `useSpecTranslation` flag is `true` by default. This flag was passed to `getTransformationMatrix()` on the depth frame's extrinsics, which reads `specTranslation`. However, `specTranslation` is not populated by StereoDepth (it's `(0, 0, 0)`), resulting in the entire baseline translation being dropped and the housing transform applied only rotation with zero translation.

## Specification
Always use the calibrated `translation` (not `specTranslation`) when reading the frame's extrinsics (`getTransformationMatrix(false, unit)`). The `useSpecTranslation` flag is still passed to calibration handler lookups (`getCameraExtrinsics`, `getHousingCalibration`) that define the target coordinate system, where spec translations are properly populated from the calibration database.

Affects point cloud output when `setTargetCoordinateSystem(HousingCoordinateSystem)` is used. Previously the translation component was silently dropped; now it is correctly applied. No breaking API changes.

## Deployment Plan
None / not applicable

## Testing & Validation
- Verified with MRE script using Kabsch algorithm to measure the actually applied transform: housing target now shows ~3.75cm translation matching `getHousingCalibration()`, previously showed ~0.
- All 3 PointCloud tests pass (`pointclouddata_test`, `point_cloud_test`, `pointcloud_test`).
- `point_cloud_showcase.py` runs successfully on OAK-D-W.

## AI Usage
Assisted-by: GitHub Copilot:claude-opus-4.6

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved point cloud coordinate transformations to consistently use calibrated frame translations.
  * Updated camera-socket target transforms to be derived from calibrated camera extrinsics consistently.
* **Breaking Changes**
  * `setTargetCoordinateSystem(...)` now takes only the coordinate-system selector (camera socket or housing); the translation toggle behavior has been removed.
  * Removed `PointCloudConfig.getUseSpecTranslation()` and stopped serializing the translation setting.
* **Python Updates**
  * Updated Python bindings to the new signatures; legacy boolean overloads now emit `DeprecationWarning` and ignore the flag.
* **Tests**
  * Added coverage to verify measured vs spec translation behavior in extrinsics.

### DIFF
--- a/bindings/python/src/pipeline/datatype/PointCloudConfigBindings.cpp
+++ b/bindings/python/src/pipeline/datatype/PointCloudConfigBindings.cpp
@@ -60,19 +60,29 @@ void bind_pointcloudconfig(pybind11::module& m, void* pCallstack) {
         .def("getLengthUnit", &PointCloudConfig::getLengthUnit, DOC(dai, PointCloudConfig, getLengthUnit))
         .def("setLengthUnit", &PointCloudConfig::setLengthUnit, DOC(dai, PointCloudConfig, setLengthUnit))
         .def("setTargetCoordinateSystem",
-             py::overload_cast<CameraBoardSocket, bool>(&PointCloudConfig::setTargetCoordinateSystem),
+             py::overload_cast<CameraBoardSocket>(&PointCloudConfig::setTargetCoordinateSystem),
              py::arg("targetCamera"),
-             py::arg("useSpecTranslation") = false,
              DOC(dai, PointCloudConfig, setTargetCoordinateSystem))
         .def("setTargetCoordinateSystem",
-             py::overload_cast<HousingCoordinateSystem, bool>(&PointCloudConfig::setTargetCoordinateSystem),
+             py::overload_cast<HousingCoordinateSystem>(&PointCloudConfig::setTargetCoordinateSystem),
              py::arg("housingCS"),
-             py::arg("useSpecTranslation") = true,
              DOC(dai, PointCloudConfig, setTargetCoordinateSystem, 2))
         .def("getCoordinateSystemType", &PointCloudConfig::getCoordinateSystemType, DOC(dai, PointCloudConfig, getCoordinateSystemType))
         .def("getTargetCameraSocket", &PointCloudConfig::getTargetCameraSocket, DOC(dai, PointCloudConfig, getTargetCameraSocket))
         .def("getTargetHousingCS", &PointCloudConfig::getTargetHousingCS, DOC(dai, PointCloudConfig, getTargetHousingCS))
         .def("getUseSpecTranslation", &PointCloudConfig::getUseSpecTranslation, DOC(dai, PointCloudConfig, getUseSpecTranslation))
+        .def(
+            "setTargetCoordinateSystem",
+            py::overload_cast<CameraBoardSocket, bool>(&PointCloudConfig::setTargetCoordinateSystem),
+            py::arg("targetCamera"),
+            py::arg("useSpecTranslation"),
+            "**Deprecated:** Use setTargetCoordinateSystem(targetCamera) instead.")
+        .def(
+            "setTargetCoordinateSystem",
+            py::overload_cast<HousingCoordinateSystem, bool>(&PointCloudConfig::setTargetCoordinateSystem),
+            py::arg("housingCS"),
+            py::arg("useSpecTranslation"),
+            "**Deprecated:** Use setTargetCoordinateSystem(housingCS) instead.")
         // Deprecated
         .def(
             "getSparse",
--- a/bindings/python/src/pipeline/datatype/PointCloudConfigBindings.cpp
+++ b/bindings/python/src/pipeline/datatype/PointCloudConfigBindings.cpp
@@ -71,18 +71,16 @@ void bind_pointcloudconfig(pybind11::module& m, void* pCallstack) {
         .def("getTargetCameraSocket", &PointCloudConfig::getTargetCameraSocket, DOC(dai, PointCloudConfig, getTargetCameraSocket))
         .def("getTargetHousingCS", &PointCloudConfig::getTargetHousingCS, DOC(dai, PointCloudConfig, getTargetHousingCS))
         .def("getUseSpecTranslation", &PointCloudConfig::getUseSpecTranslation, DOC(dai, PointCloudConfig, getUseSpecTranslation))
-        .def(
-            "setTargetCoordinateSystem",
-            py::overload_cast<CameraBoardSocket, bool>(&PointCloudConfig::setTargetCoordinateSystem),
-            py::arg("targetCamera"),
-            py::arg("useSpecTranslation"),
-            "**Deprecated:** Use setTargetCoordinateSystem(targetCamera) instead.")
-        .def(
-            "setTargetCoordinateSystem",
-            py::overload_cast<HousingCoordinateSystem, bool>(&PointCloudConfig::setTargetCoordinateSystem),
-            py::arg("housingCS"),
-            py::arg("useSpecTranslation"),
-            "**Deprecated:** Use setTargetCoordinateSystem(housingCS) instead.")
+        .def("setTargetCoordinateSystem",
+             py::overload_cast<CameraBoardSocket, bool>(&PointCloudConfig::setTargetCoordinateSystem),
+             py::arg("targetCamera"),
+             py::arg("useSpecTranslation"),
+             "**Deprecated:** Use setTargetCoordinateSystem(targetCamera) instead.")
+        .def("setTargetCoordinateSystem",
+             py::overload_cast<HousingCoordinateSystem, bool>(&PointCloudConfig::setTargetCoordinateSystem),
+             py::arg("housingCS"),
+             py::arg("useSpecTranslation"),
+             "**Deprecated:** Use setTargetCoordinateSystem(housingCS) instead.")
         // Deprecated
         .def(
             "getSparse",
--- a/bindings/python/src/pipeline/node/PointCloudBindings.cpp
+++ b/bindings/python/src/pipeline/node/PointCloudBindings.cpp
@@ -48,15 +48,25 @@ void bind_pointcloud(pybind11::module& m, void* pCallstack) {
         .def("useCPUMT", &PointCloud::useCPUMT, py::arg("numThreads") = 2, DOC(dai, node, PointCloud, useCPUMT))
         .def("useGPU", &PointCloud::useGPU, py::arg("device") = 0, DOC(dai, node, PointCloud, useGPU))
         .def("setTargetCoordinateSystem",
-             py::overload_cast<CameraBoardSocket, bool>(&PointCloud::setTargetCoordinateSystem),
+             py::overload_cast<CameraBoardSocket>(&PointCloud::setTargetCoordinateSystem),
              py::arg("targetCamera"),
-             py::arg("useSpecTranslation") = false,
              DOC(dai, node, PointCloud, setTargetCoordinateSystem))
         .def("setTargetCoordinateSystem",
-             py::overload_cast<HousingCoordinateSystem, bool>(&PointCloud::setTargetCoordinateSystem),
+             py::overload_cast<HousingCoordinateSystem>(&PointCloud::setTargetCoordinateSystem),
              py::arg("housingCS"),
-             py::arg("useSpecTranslation") = true,
-             DOC(dai, node, PointCloud, setTargetCoordinateSystem, 2));
+             DOC(dai, node, PointCloud, setTargetCoordinateSystem, 2))
+        .def(
+            "setTargetCoordinateSystem",
+            py::overload_cast<CameraBoardSocket, bool>(&PointCloud::setTargetCoordinateSystem),
+            py::arg("targetCamera"),
+            py::arg("useSpecTranslation"),
+            "**Deprecated:** Use setTargetCoordinateSystem(targetCamera) instead.")
+        .def(
+            "setTargetCoordinateSystem",
+            py::overload_cast<HousingCoordinateSystem, bool>(&PointCloud::setTargetCoordinateSystem),
+            py::arg("housingCS"),
+            py::arg("useSpecTranslation"),
+            "**Deprecated:** Use setTargetCoordinateSystem(housingCS) instead.");
 
     // ALIAS
     daiNodeModule.attr("PointCloud").attr("Properties") = properties;
--- a/bindings/python/src/pipeline/node/PointCloudBindings.cpp
+++ b/bindings/python/src/pipeline/node/PointCloudBindings.cpp
@@ -55,18 +55,16 @@ void bind_pointcloud(pybind11::module& m, void* pCallstack) {
              py::overload_cast<HousingCoordinateSystem>(&PointCloud::setTargetCoordinateSystem),
              py::arg("housingCS"),
              DOC(dai, node, PointCloud, setTargetCoordinateSystem, 2))
-        .def(
-            "setTargetCoordinateSystem",
-            py::overload_cast<CameraBoardSocket, bool>(&PointCloud::setTargetCoordinateSystem),
-            py::arg("targetCamera"),
-            py::arg("useSpecTranslation"),
-            "**Deprecated:** Use setTargetCoordinateSystem(targetCamera) instead.")
-        .def(
-            "setTargetCoordinateSystem",
-            py::overload_cast<HousingCoordinateSystem, bool>(&PointCloud::setTargetCoordinateSystem),
-            py::arg("housingCS"),
-            py::arg("useSpecTranslation"),
-            "**Deprecated:** Use setTargetCoordinateSystem(housingCS) instead.");
+        .def("setTargetCoordinateSystem",
+             py::overload_cast<CameraBoardSocket, bool>(&PointCloud::setTargetCoordinateSystem),
+             py::arg("targetCamera"),
+             py::arg("useSpecTranslation"),
+             "**Deprecated:** Use setTargetCoordinateSystem(targetCamera) instead.")
+        .def("setTargetCoordinateSystem",
+             py::overload_cast<HousingCoordinateSystem, bool>(&PointCloud::setTargetCoordinateSystem),
+             py::arg("housingCS"),
+             py::arg("useSpecTranslation"),
+             "**Deprecated:** Use setTargetCoordinateSystem(housingCS) instead.");
 
     // ALIAS
     daiNodeModule.attr("PointCloud").attr("Properties") = properties;
--- a/include/depthai/pipeline/datatype/PointCloudConfig.hpp
+++ b/include/depthai/pipeline/datatype/PointCloudConfig.hpp
@@ -9,9 +9,6 @@
 #include "depthai/pipeline/datatype/Buffer.hpp"
 
 namespace dai {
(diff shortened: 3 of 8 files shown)


aljazkonec1:
Thanks, 
Any particular reason the default is to useSpectTranslation? Dont we usually preffer calibration translation as its more accurate?

If so, I would rather see that PointCloudConfig.useSpectTranslation is default to false + a `setUseSpecTranslation` function is exposed on the config.

Could you also please add a test to cover it as you mentioned that it was not caught by a test

MaticTonin (approved):
Looks good to me, please as well prepare other PR and link it to this PR in comments so issue with ImgTRanformation is resolved

src/pipeline/node/PointCloud.cpp
```diff
@@ -534,7 +534,10 @@ void PointCloud::setCoordinateTransformation(const ImgFrame& depthFrame, const P
             "PointCloud: depth frame extrinsics toCameraSocket is AUTO. "
             "Ensure the depth frame has valid extrinsics with a specific camera socket set.");
     }
-    auto T_frame_to_ref = matrix::toVecMatrix4x4(frameExtrinsics.getTransformationMatrix(useSpecTranslation, unit));
+    // Always use the calibrated translation (not specTranslation) from the frame's extrinsics,
+    // because specTranslation may not be populated. useSpecTranslation only applies to
+    // calibration handler lookups that define the target coordinate system.
+    auto T_frame_to_ref = matrix::toVecMatrix4x4(frameExtrinsics.getTransformationMatrix(false, unit));
```
JakubFara: Can you add a simple test testing the new behavior? ... set the spec tvec to nonsense

JakubFara (changes requested):
LGTM; only change the tests such that they are independent on spec tvec

pheec:
@aljazkonec1 
> Any particular reason the default is to useSpectTranslation? Dont we usually preffer calibration translation as its more accurate?

The housing transformations have uncalibrated translations based on the CAD drawings of the devices. Those translations are stored in a database created at compile time.

@aljazkonec1
> If so, I would rather see that PointCloudConfig.useSpectTranslation is default to false + a setUseSpecTranslation function is exposed on the config.

I agree. I suggest we set the default to false in PointCloudConfig, but always call CalibrationHandler.getHousingCalibration with useSpecTranslation=true regardless of the config setting. The reason is that the translation field for housing transformations is not populated during factory calibration, and it does not seem like it will be any time soon.

After these changes, the PointCloud node would use useSpecTranslation as follows:
CalibrationHandler.getCameraExtrinsics → uses useSpecTranslation from config
depthFrameExtrinsics.getTransformationMatrix → uses useSpecTranslation from config
CalibrationHandler.getHousingCalibration → always true (housing only has spec translations)

@aljazkonec1 @JakubFara 
> Could you also please add a test to cover it as you mentioned that it was not caught by a test

> change the tests such that they are independent on spec tvec

Sure

CC: @MaticTonin

aljazkonec1:
> I agree. I suggest we set the default to false in PointCloudConfig, but always call CalibrationHandler.getHousingCalibration with useSpecTranslation=true regardless of the config setting. The reason is that the translation field for housing transformations is not populated during factory calibration, and it does not seem like it will be any time soon.

Yes lets then set false and getHousingCalibration should request useSpecTranslation=true, even if its set to (0, 0, 0) irregardles of if its correct or not.

JakubFara:
What about removing `specTranslation` field from PointCloud config. This seems to me rather confusing then useful.

pheec:
After discussion with @JakubFara, we suggest completely removing useSpecTranslation from PointCloudConfig.

The node would use the following hardcoded settings:
CalibrationHandler::getCameraExtrinsics → useSpecTranslation = false
depthFrameExtrinsics.getTransformationMatrix → useSpecTranslation = false
CalibrationHandler::getHousingCalibration → useSpecTranslation = true

The reason for that is that its meaning is ambiguous, as its used in three different functions.

Thoughts? @aljazkonec1 
cc: @MaticTonin

MaticTonin:
Sounds good by me, its always better to then remove it and bring it back when proper implementation si done.

bindings/python/src/pipeline/datatype/PointCloudConfigBindings.cpp
```diff
@@ -60,19 +60,43 @@ void bind_pointcloudconfig(pybind11::module& m, void* pCallstack) {
...
+                if(PyErr_WarnEx(PyExc_DeprecationWarning,
+                                "useSpecTranslation parameter is deprecated and ignored. Use setTargetCoordinateSystem(housingCS) instead.",
+                                1)
+                   < 0)
+                    throw py::error_already_set();
+                return cfg.setTargetCoordinateSystem(housingCS);
+            },
+            py::arg("housingCS"),
+            py::arg("useSpecTranslation"),
+            "**Deprecated:** useSpecTranslation parameter is ignored.")
```
JakubFara: Now the C++ and python api is different in the python it is deprecated and in c++ removed. Is there a standard way how to deprecate function in Depthai @aljazkonec1 ?

JakubFara (commented):
I think that the deprecation should be done differently:

bindings/python/src/pipeline/datatype/PointCloudConfigBindings.cpp
aljazkonec1: The standard way is to first add a deprecation **notice** so users have time to adapt their code before API change. We then typically remove funcitons after 1 or 2 releases (depending on severity and popularity). 

That being said, what is the reasoning for removing specTranslation param?  
For `CalibrationHandler::getCameraExtrinsics` and `depthFrameExtrinsics.getTransformationMatrix` it feels natural to expose specTranslation no? Why does CalibrationHandler function need to be changed at all? 

For `CalibrationHandler::getHousingCalibration` its natural to not specify specTranslation given the naming.

Asking because I do not want to break API unless necessarry.

JakubFara: I understand that noticing the user first is necessary, but my question was about the unified form—(use `log::warn("Method XY is deprecated, use Z instead.")` + add `@deprecated` to the doc string?).

aljazkonec1: Yes that is correct. But in https://github.com/luxonis/depthai-core/pull/1854/commits/5e1e88de9adaae1985c1752d2aedd2bac6b6411f the actual API was changed and useSpecTranslation was removed.

JakubFara: We can add also the version where we will deprecate it :  `log::warn("Method XY is deprecated and will be removed in Depthai X.Y . Use Z instead.")`. It is not useful only for the user but also for us because we can track which functions we should remove in each release.

aljazkonec1: Yes exactly. So there should be two of each function like (and c++ versions):
```.def("setTargetCoordinateSystem",
             py::overload_cast<CameraBoardSocket, bool>(&PointCloudConfig::setTargetCoordinateSystem),
             py::arg("targetCamera"),
             py::arg("useSpecTranslation") = false,
             DOC(dai, PointCloudConfig, setTargetCoordinateSystem))
```

```
.def("setTargetCoordinateSystem",
             py::overload_cast<CameraBoardSocket>(&PointCloudConfig::setTargetCoordinateSystem),
             py::arg("targetCamera"),
             DOC(dai, PointCloudConfig, setTargetCoordinateSystem))
```
The first funciton should notify in docstring and in the funciton itself (via logger::warn) that it will be removed in a future update.

include/depthai/pipeline/datatype/PointCloudConfig.hpp
```diff
@@ -81,16 +80,14 @@ class PointCloudConfig : public Buffer {
     /**
      * Set target coordinate system to another camera socket.
      * @param targetCamera Target camera socket
-     * @param useSpecTranslation Use spec translation instead of calibration (default: false)
      */
-    PointCloudConfig& setTargetCoordinateSystem(CameraBoardSocket targetCamera, bool useSpecTranslation = false);
+    PointCloudConfig& setTargetCoordinateSystem(CameraBoardSocket targetCamera);
```
MaticTonin, line 85: This one to be left as it is, since user can actually set it and it does not affect the HousingSpecTranslation path at all.

include/depthai/pipeline/datatype/PointCloudConfig.hpp
```diff
@@ -81,16 +80,14 @@ class PointCloudConfig : public Buffer {
...
 
     /**
      * Set target coordinate system to housing coordinate system.
      * @param housingCS Target housing coordinate system
-     * @param useSpecTranslation Whether to use spec translation (default: true)
      */
-    PointCloudConfig& setTargetCoordinateSystem(HousingCoordinateSystem housingCS, bool useSpecTranslation = true);
+    PointCloudConfig& setTargetCoordinateSystem(HousingCoordinateSystem housingCS);
```
MaticTonin, line 92: To depricate this one, you create another function with what you wanna depricate
```
    PointCloudConfig& setTargetCoordinateSystem(HousingCoordinateSystem housingCS, bool useSpecTranslation = true);
    PointCloudConfig& setTargetCoordinateSystem(HousingCoordinateSystem housingCS);
```

while you in the function you wanna depricate, add the warning, that user should rather use the one, which is not depricated.

bindings/python/src/pipeline/node/PointCloudBindings.cpp
```diff
@@ -48,15 +48,40 @@ void bind_pointcloud(pybind11::module& m, void* pCallstack) {
...
+                if(PyErr_WarnEx(PyExc_DeprecationWarning,
+                                "useSpecTranslation parameter is deprecated and ignored. Use setTargetCoordinateSystem(housingCS) instead.",
+                                1)
+                   < 0)
+                    throw py::error_already_set();
+                node.setTargetCoordinateSystem(housingCS);
+            },
+            py::arg("housingCS"),
+            py::arg("useSpecTranslation"),
+            "**Deprecated:** useSpecTranslation parameter is ignored.");
```
MaticTonin: So this then gets removed and the deprecation warning its added to that part.

include/depthai/pipeline/node/PointCloud.hpp
```diff
@@ -203,17 +203,15 @@ class PointCloud : public DeviceNodeCRTP<DeviceNode, PointCloud, PointCloudPrope
     /**
      * Set target coordinate system to transform point cloud
      * @param targetCamera Target camera socket
-     * @param useSpecTranslation Use spec translation instead of calibration (default: false)
      */
-    void setTargetCoordinateSystem(CameraBoardSocket targetCamera, bool useSpecTranslation = false);
+    void setTargetCoordinateSystem(CameraBoardSocket targetCamera);
```
MaticTonin, line 207: We keep as it is, dont depricate, since its OK for user to useSpecTranslation betweeen camera links

include/depthai/pipeline/node/PointCloud.hpp
```diff
@@ -203,17 +203,15 @@ class PointCloud : public DeviceNodeCRTP<DeviceNode, PointCloud, PointCloudPrope
...
 
     /**
      * Set target coordinate system to housing coordinate system
      * Point cloud will be transformed to this housing coordinate system
      * @param housingCS Target housing coordinate system
-     * @param useSpecTranslation Whether to use spec translation (default: true)
      */
-    void setTargetCoordinateSystem(HousingCoordinateSystem housingCS, bool useSpecTranslation = true);
```
MaticTonin, line 216: We depricate as below

MaticTonin (changes requested):
Quick reply based on discussions.

pheec:
Just to make it clear. Is this the desired state after the proposed changes? @aljazkonec1 @JakubFara @MaticTonin 

Storage:
Removed from PointCloudConfig — the field no longer exists. The DEPTHAI_SERIALIZE macro no longer includes it.

Where it can be set (deprecated API):
The two-argument overloads still exist but are deprecated and ignore the bool parameter:
- PointCloudConfig::setTargetCoordinateSystem(CameraBoardSocket, bool) — [[deprecated]]
- PointCloudConfig::setTargetCoordinateSystem(HousingCoordinateSystem, bool) — [[deprecated]]
- PointCloud::setTargetCoordinateSystem(CameraBoardSocket, bool) — [[deprecated]]
- PointCloud::setTargetCoordinateSystem(HousingCoordinateSystem, bool) — [[deprecated]]
- Python bindings (both Config and Node) — lambda shims emit DeprecationWarning and delegate to single-arg version

Where useSpecTranslation is actually used (hardcoded in the node):
In PointCloud.cpp setCoordinateTransformation():
- frameExtrinsics.getTransformationMatrix(false, unit) — frame's specTranslation is (0,0,0), not populated by StereoDepth
- calibHandler.getCameraExtrinsics(ref, target, false, unit) — use calibrated (measured) translation
- calibHandler.getHousingCalibration(ref, housing, true, unit) — housing only has spec translations (from CAD database)

New API (preferred):
config.setTargetCoordinateSystem(dai.CameraBoardSocket.CAM_A)
config.setTargetCoordinateSystem(dai.HousingCoordinateSystem.CAM_A)
Single argument — no useSpecTranslation parameter. The node decides internally which translation source to use per path.

-------------------------------------------------------------------------------------------------------------------------

I talked to @MaticTonin and he would like to keep the useSpecTranslation for example here:
void setTargetCoordinateSystem(CameraBoardSocket targetCamera, bool useSpecTranslation = false);

but thats not clear, as it would just change the useSpecTranslation in multiple different functions:
- CalibrationHandler::getCameraExtrinsics → useSpecTranslation = false
- depthFrameExtrinsics.getTransformationMatrix → useSpecTranslation = false

So I would go with the proposed solution described above, but with the correct way of deprecating.
Thoughts?

JakubFara:
@MaticTonin What it reason for keeping it in `void setTargetCoordinateSystem()`. Is there a use-case where it is useful?

aljazkonec1:
> The two-argument overloads still exist but are deprecated and ignore the bool parameter

Deprecation means that the exact same behavior is kept and ONLY a notice / warning is added. The current implementation is an actual behavior change due to ignoring the bool parameter. 

> I talked to @MaticTonin and he would like to keep the useSpecTranslation for example here:
> void setTargetCoordinateSystem(CameraBoardSocket targetCamera, bool useSpecTranslation = false);
> 
> but thats not clear, as it would just change the useSpecTranslation in multiple different functions:
> 
> CalibrationHandler::getCameraExtrinsics → useSpecTranslation = false
> depthFrameExtrinsics.getTransformationMatrix → useSpecTranslation = false

Isnt this the current version of setTargetCoordinateSystem? Is it intended that no one will ever use housing calibration? What if someone wants to set some external housing calibration from wherever they might integrate our cameras?

include/depthai/pipeline/datatype/PointCloudConfig.hpp
```diff
@@ -9,6 +9,9 @@
 #include "depthai/pipeline/datatype/Buffer.hpp"
 
 namespace dai {
+namespace node {
+class PointCloud;
+}
```
MaticTonin: I dont see its needed.

include/depthai/pipeline/datatype/PointCloudConfig.hpp
```diff
@@ -30,10 +33,12 @@ class PointCloudConfig : public Buffer {
     };
 
    private:
+    friend class node::PointCloud;
+
```
MaticTonin: Same here, does not seem needed.

MaticTonin (changes requested):
Way cleaner than before with good deprecation notice.

**Required to run::**

`cmake --build build --target clangformat`

src/pipeline/node/PointCloud.cpp
```diff
@@ -479,12 +480,28 @@ void PointCloud::useGPU(uint32_t device) {
...
+    initialConfig->setTargetCoordinateSystem(housingCS);
+}
+
 void PointCloud::setTargetCoordinateSystem(CameraBoardSocket targetCamera, bool useSpecTranslation) {
-    initialConfig->setTargetCoordinateSystem(targetCamera, useSpecTranslation);
+    spdlog::warn("PointCloud::setTargetCoordinateSystem(CameraBoardSocket, bool) is deprecated. "
+                 "Use PointCloud::setTargetCoordinateSystem(CameraBoardSocket) instead.");
+    initialConfig->coordSystemType = PointCloudConfig::CoordinateSystemType::CAMERA_SOCKET;
+    initialConfig->targetCameraSocket = targetCamera;
+    initialConfig->useSpecTranslation = useSpecTranslation;
```
MaticTonin: Why doesnt this method just call then the 

```
void PointCloud::setTargetCoordinateSystem(CameraBoardSocket targetCamera) {
    initialConfig->setTargetCoordinateSystem(targetCamera);
}
```
and ignore spec translation completely?

include/depthai/pipeline/datatype/PointCloudConfig.hpp
pheec: The reason the friend was introduced is that the deprecated bool overloads in PointCloud.cpp no longer delegate to PointCloudConfig::setTargetCoordinateSystem(...) and instead write PointCloudConfig private members directly:

include/depthai/pipeline/datatype/PointCloudConfig.hpp
pheec: same as above

src/pipeline/node/PointCloud.cpp
pheec: > Aljaz: Deprecation means that the exact same behavior is kept 

It is done this way to preserve the old behavior of the deprecated overload.
PointCloud::setTargetCoordinateSystem(CameraBoardSocket, bool) historically allowed the caller to explicitly control useSpecTranslation. If it now just called the new overload and ignored the bool, then the deprecated API would silently change behavior.

With the current implementation:
- the deprecated overload still honors the passed useSpecTranslation value exactly as before
- the new overload without bool does not overwrite useSpecTranslation in the config at all

include/depthai/pipeline/datatype/PointCloudConfig.hpp
MaticTonin: AH i see, you are correct yes, Then we can leave it as it is.

MaticTonin (approved):
LGTM now

include/depthai/pipeline/datatype/PointCloudConfig.hpp
aljazkonec1: PointCloudConfig exposes public setters for the bool overloads. So in PointCloud.cpp setters you can just set values like:
`void PointCloud::setTargetCoordinateSystem(HousingCoordinateSystem housingCS, bool useSpecTranslation) {
initialConfig.setTargetCoordinateSystem(housingCS, useSpecTranslation);
}`
no need to add friend class.

aljazkonec1 (changes requested):
Looks good now. No need to add friend class.

JakubFara (approved):
LGTM

aljazkonec1 (approved):

PR #1862, opened 2026-06-26
Bugfix/DCL use getIntrinsics

MaticTonin:
There was an issue that DCL take into the account the SourceIntrinsicMatrix, which is resolution of the frame, before being altered by the rescaling.
SInce RVC2 devices support the 640x400 downscale on sensor level, getIntrinsics = getSourceIntrinsics(), while for the RVC4 devices, that is not true.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated camera calibration handling so the correct intrinsic matrices are used for downstream calibration processing.
  * This should improve consistency in camera calibration results for affected devices.


JakubFara (approved):
LGTM

PR #1866, opened 2026-06-30
Remove move constructor and delegate to copy.

thehiddenwaffle:
This does not improve anything for the C++ side of things but improves usability for other languages via generated FFI bindings by allowing them to hold bare pointers to a pipeline object without the possibility of C++ move semantics invalidating that memory. 

It's probably worth noting that this currently this only happens in a single place and that usage is unable to affect user space, but for the purpose of future guarantees and static assertions made downstream this would be useful.

## Summary by CodeRabbit

* **API Improvements**
  * Pipeline objects now have clearly defined copy and move behavior, making it easier and more consistent to store, pass around, assign, and return pipeline instances in your application code.


aljazkonec1:
Thanks, looks good to me. Will let tests pass and then we can merge :)

aljazkonec1 (dismissed):

PR #1893, opened 2026-07-13
Add save and load methods to messages

asahtik:
## Purpose

Adds save and load methods to `ProtoSerializable` messages. This enables the user to save the message (metadata + data) to disk and then load it at a later time. Also adds aliases `ReplayMessage` and `RecordMessage` that better describe the functionality of `ReplayMetadataOnly` and `RecordMetadataOnly` nodes.

## Specification

Added deserialization to previously unsupported message types (this also enables their usage in Record & Replay).

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

## AI Usage

Assisted-by: AGENT_NAME:MODEL_VERSION [TOOL1] [TOOL2]

Submitted code was reviewed by a human: YES/NO

The author is taking the responsibility for the contribution: YES/NO

## Summary by CodeRabbit

* **New Features**
  * Added Python support for saving and loading protobuf-serialized datatypes.
  * Enabled protobuf deserialization for frames, detections, point clouds, IMU, RGB-D, segmentation, annotations, and encoded frames.
  * Added normalized-coordinate metadata for points, sizes, and rectangles.
  * Exposed `RecordMessage` and `ReplayMessage` node attributes.

* **Bug Fixes**
  * Corrected metadata-only serialization and RGB-D replay alignment.
  * Expanded replay support for additional datatypes.

* **Tests**
  * Added serialization round-trip and replay coverage.


JakubFara:
I've been thinking about the file suffixes. If we used extensions like `.ImgFrame.dai`, `.PointCloud.dai` ..., we could automatically determine how to load the messages and perform additional validation checks.

asahtik:
@JakubFara the datatype is saved in the file and validated when read. IMO this is the better option since renaming the file doesn't corrupt it.

JakubFara:
Another small suggestion: It would be nicer to do

```
left = dai.ImgFrame().load("left.dai")
```

or even

```
left = dai.ImgFrame("left.dai")

```
than

```
left = dai.ImgFrame()
left.load("left.dai")
```

aljazkonec1 (changes requested):
So the reason for the new ADatatypeInterface was so that you could call getDatatype() and to then set it in the file header?

All the serializable messages override `serializeSchema()` which has `schemaName = descriptor()->full_name();` (eg dai.proto.img_frame.ImgFrame) and thats what is used for mcap. We could also use it here to skip the reinterpret cast.

One additional thing: using enumerator as the file header is unstable because it can change when adding new message types, we could replace it with a protobuf message like:

```
syntax = "proto3";
package dai.proto.message_file;
message MessageFile {
    string type_name     = 1;  // descriptor full_name of the payload
    bool   metadata_only = 2;
    bytes  payload       = 3;  // output of serializeProto()
}
```

Then save and load functions would look like:

```
void ProtoSerializable::save(const std::filesystem::path& path, bool metadataOnly) const {
    proto::message_file::MessageFile file;
    file.set_type_name(serializeSchema().schemaName);
    file.set_metadata_only(metadataOnly);
    const auto bytes = serializeProto(metadataOnly);
    file.set_payload(bytes.data(), bytes.size());
    ...
}

void ProtoSerializable::load(const std::filesystem::path& path) {
    std::ifstream in(resolveDataPath(path), std::ios::binary);
    proto::message_file::MessageFile file;
    if(!in || !file.ParseFromIstream(&in)) throw std::runtime_error("Failed to read " + path.string());
    const auto expected = serializeSchema().schemaName;
    if(file.type_name() != expected)
        throw std::runtime_error("File " + path.string() + " contains " + file.type_name() + ", expected " + expected);
    deserializeProto({file.payload().begin(), file.payload().end()}, file.metadata_only());
}
```

We can then get rid of the ADatatypeInterface + you `load` would know if it is metadata_only + mcap, foxglove and the save function would rely on the same full_name() descriptor + we can add a generic `load` buffer function that reads a file and at runtime figures out the datatype

include/depthai/pipeline/datatype/ADatatype.hpp
```diff
@@ -4,12 +4,10 @@
 
 #include "depthai/pipeline/datatype/DatatypeEnum.hpp"
 #include "depthai/utility/Memory.hpp"
-#include "depthai/utility/Serialization.hpp"
 #include "depthai/utility/VectorMemory.hpp"
 
 namespace dai {
 
-/// Abstract message
```
aljazkonec1, line 12: Why delete?

src/utility/ProtoSerialize.cpp
```diff
@@ -128,6 +195,19 @@ ImgTransformation deserializeImgTransformation(const proto::common::ImgTransform
...
+        }
+        for(auto i = 0; i < 3; ++i) {
+            if(matrix.arrays(i).values_size() < 3) {
+                throw std::runtime_error(std::string(matrixName) + " must contain at least three columns per row");
+            }
+        }
+    };
+    validateMatrix(imgTransformation.transformationmatrix(), "Transformation matrix");
+    validateMatrix(imgTransformation.sourceintrinsicmatrix(), "Source intrinsic matrix");
+
```
aljazkonec1: What about old recordings that have empty matrices? Is this relevant to the save / load method at all?

src/utility/ProtoSerialize.cpp
```diff
@@ -20,6 +20,29 @@
...
+            protoMsg protoMessage;                                                           \
+            const auto* descriptor = protoMessage.GetDescriptor();                           \
+            if(descriptor == nullptr) {                                                      \
+                throw std::runtime_error("Failed to get protobuf descriptor");               \
+            }                                                                                \
+            ProtoSerializable::SchemaPair returnPair;                                        \
+            returnPair.schemaName = descriptor->full_name();                                 \
+            returnPair.schema = serializeFdSet(descriptor);                                  \
+            return returnPair;                                                               \
+        }                                                                                    \
```
aljazkonec1, line 39: `getProtoSchema` runs `serializeFdSet` on every call, and save /load calls it just to read `schemaName`. We could cache returnPair as a function-local static const SchemaPair, so it is built once on first use and only copied afterwards

src/utility/ProtoSerializable.cpp
```diff
@@ -1,7 +1,74 @@
...
+        message.set_proto_message(serializedMessage.data(), serializedMessage.size());
+    }
+    writeMsgBinaryFile(resolveDataPath(path), message);
+}
+
+void ProtoSerializable::load(const std::filesystem::path& path) {
+    const auto message = readMsgBinaryFile(resolveDataPath(path));
+    const auto expectedSchemaName = serializeSchema().schemaName;
+    if(message.schema_name() != expectedSchemaName) {
+        throw std::runtime_error("Schema mismatch when reading file: " + path.string());
```
aljazkonec1: I would also print the expectedSchemaName

aljazkonec1 (changes requested):
Thanks! left a couple more comments. + the failing protobuf consistency and build failure when protobuf is disabled

ci/check_protobuf_consistency.sh
aljazkonec1, line 1: Is the only fix to update the protobuf consistency? why is that?

asahtik: I could either remove the new macro and manually declare everything or I could update the script to check for the macro instead of declarations. This way reduces LOC and repetition in ProtoSerialize and makes this script simpler and more robust.

aljazkonec1 (approved):
Looks good I found maybe a couple of nits:
- using `save("")` writes .dai
- using `save(".")` writes ..dai
- using `save("dir/") writes dir/.dai

Could you expand the Protoserialize.cpp to reject empty or "." filenames?
After that you can merge the PR. Thanks for all the hard work on this!

PR #1904, opened 2026-07-21
Remove calibrationHandler from Rectifiacation

JakubFara:
# Purpose
Remove `CalibrationHandler` from the `Rectification` node and replace it with image transformations.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved stereo image rectification by deriving alignment from the input frames’ spatial transformation data rather than device calibration/socket mapping.
  * Rectification geometry now re-initializes automatically when either input frame’s transformation changes, ensuring updated alignment during dynamic setups.
  * Removed the rectification node’s public calibration-data accessor, improving compatibility with workflows that provide geometry via frame transformations.

### DIFF
--- a/include/depthai/pipeline/node/Rectification.hpp
+++ b/include/depthai/pipeline/node/Rectification.hpp
@@ -66,8 +66,6 @@ class Rectification : public DeviceNodeCRTP<DeviceNode, Rectification, Rectifica
 
     void run() override;
 
-    virtual CalibrationHandler getCalibrationData() const;
-
    private:
     bool runOnHostVar = false;
 };
--- a/src/pipeline/node/Rectification.cpp
+++ b/src/pipeline/node/Rectification.cpp
@@ -32,14 +32,6 @@ bool Rectification::runOnHost() const {
     return runOnHostVar;
 }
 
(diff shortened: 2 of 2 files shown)


aljazkonec1 (commented):
Thanks! I would also add an update mechanism if imgTransformations changes. Can be something simple like:
`if((! previousInput1->transformation.isEqualTransformation(input1->transformation) || (! previousInput2->transformation.isEqualTransformation(input2->transformation))) initialized = false;`

MaticTonin (commented):
Overall looks good, woudl just as @aljazkonec1 suggested, remove the check for calibHandler and add imgTransformation check.

aljazkonec1 (approved):

JakubFara:
The reinitialization is beyond of the original purpose of this PR, but I agree that it can be involved:
[Rectification fix reinitialization](https://github.com/luxonis/depthai-core/pull/1904/commits/cbb250eae11fa4f18d08db18a24c6a0ba7aea766)

MaticTonin (approved):
Looks good to me, the input difference to update new imgTransformation is adressed.

PR #1926, opened 2026-08-05
Feature/hil testing overhaul

aljazdu:
## Purpose

Adds the following testing structure:
- Short PR tests
- Specific nightly tests
- Weekly full run

## Specification

- PR tests are triggered on PR approval, however they can be triggered manually by adding the existing "testable" label
- Weekly tests are run on a schedule - specifically on the develop branch. They can be triggered manually in the workflow dispatch, where additional settings can be specified: Run sanitizer tests (true/false), Run OS tests (true/false), Run replay tests (true/false)
- Nightly tests are run on a schedule - specifically on the develop branch. 

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

## AI Usage

Assisted-by: Codex

## Summary by CodeRabbit

* **New Features**
  * Added pull request, nightly, weekly, and scheduled hardware-in-the-loop testing workflows.
  * Added configurable weekly test groups for operating systems, replay, and sanitizers.
  * Added automated test summaries, artifact uploads, and Slack notifications.
  * Added an option to skip standard test suites when running specialized tests.

* **Changes**
  * Pull request validation now runs only when explicitly labeled.
  * Removed the legacy test workflow.
  * Added required result reporting for pull request hardware tests.
  * Added concurrency controls to cancel superseded test runs.


.github/workflows/hil_testing_schedule_develop.yml
```diff
@@ -0,0 +1,45 @@
...
+
+jobs:
+  dispatch:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Trigger develop HIL testing
+        if: github.event.schedule == '0 3 * * 6'
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: gh workflow run hil_testing_main.yml --repo "$GITHUB_REPOSITORY" --ref develop
```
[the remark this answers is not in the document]
aljazdu: Is this still an issue?

.github/workflows/hil_testing_schedule_develop.yml
```diff
@@ -0,0 +1,45 @@
+name: DepthAI Core HIL Testing Develop Schedule
+
+on:
+  schedule:
```
danilo-pejovic, line 4: Why do we need a separate workflow to trigger nightly runs? I think just adding this schedule directly into nightly.yml/weekly file should work well? Nightly/weekly can also be combined into a single file that runs with different OS/platform requirements based on schedule.

.github/workflows/hil_testing_pull_request.yml
```diff
@@ -0,0 +1,179 @@
...
+    types: [labeled]
+
+# Cancel older runs for the same PR and trigger.
+concurrency:
+  group: depthai-core-pr-testing-${{ github.event.pull_request.number }}-${{ github.event.review.state || github.event.label.name }}
+  cancel-in-progress: true
+
+jobs:
+  # Build the docker container
+  build_docker_container:
```
danilo-pejovic: This seems like code duplication as these same functions will also be in test_child? Can this pipeline work similarly to test.workflow.yml previously where it would just call relevant jobs from test_child.yml and then we move relevant reporting there? 

Nothing critical but having such behaviour in reusable scripts makes it much easier to quickly set up testing to replicate specific issues we find during testing.

danilo-pejovic (commented):
Mostly lgtm, barring those nitpicks. You will probably have to create nightly/weekly workflows in main before running them else github can be bit hard to work with as it cant find them to run them.

aljazdu:
> Mostly lgtm, barring those nitpicks. You will probably have to create nightly/weekly workflows in main before running them else github can be bit hard to work with as it cant find them to run them.

Thanks @danilo-pejovic , will go over the suggestions. Yes they have to be merged in main (as our default branch), to be picked by the scheduler and run.

.github/workflows/hil_testing_schedule_develop.yml
aljazdu: The goal is to test develop, while GitHub scheduled workflows always run on the default branch (main) and schedule does not support specifying a different ref. The --ref develop option is only available when dispatching another workflow. We could make each workflow self-dispatch to develop, but that would duplicate the dispatch logic and still create two runs. Keeping one scheduler on main is the cleaner way to trigger both workflows against develop.

.github/workflows/hil_testing_pull_request.yml
aljazdu: Agree with this. Will update the workflows.

aljazkonec1 (approved):
Thanks! Looks good to me.

PR #1929, opened 2026-08-06
Reallocate pool in ImageManip if running low on frames

asahtik:
## Purpose

This is a fix for HFR streams that use the ImageManip node. The output frame pool ran out of frames which limited the output FPS.

## Specification

On RVC4 the node preallocates `numFramesPool` frames with size `outputFrameSize` as it did before. If the pool runs out of frames it reallocates to max 30 frames of the output size. The number of frames is also limited by `maxPoolSize`. When reallocating a warning is shown with the new pool size.

RVC2 is unchanged apart from bumping core with the new `ImageManipProperties` serialization.

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

## Summary by CodeRabbit

* **New Features**
  * Added a configurable maximum output pool size for ImageManip.
  * Exposed the pool-size setting and setter through Python bindings.
  * Added automatic pool-size handling for RVC4 devices.
  * Added a default pool-size configuration for easier setup.

* **Bug Fixes**
  * Improved ImageManip output buffer allocation to use the calculated output size, helping reduce unnecessary memory usage.


aljazkonec1 (approved):
Thanks! I tested it out and works as expected

PR #1930, opened 2026-08-06
Fix user bootloader support check

### LINKED ISSUE #1102: `isUserBootloader()` fails to check conditions correctly
depthai-core v2.27.0 `isUserBootloader()` fails to check conditions correctly.

I think the code needs to check what is flashed...not what is currently running since it could be running from an esoteric/embedded blob. Should also check that it is a NETWORK sensor not usb, so calling `isUserBootloaderSupported()` fits well.

https://github.com/luxonis/depthai-core/blob/729e478108f40d117ef7876b1360596f35645ddf/src/device/DeviceBootloader.cpp#L729-L733

I think that block should instead be

```c++
bool DeviceBootloader::isUserBootloader() const {
    // Check if bootloader version is adequate
    if(!isUserBootloaderSupported()) {
        return false;
    }
```

Daniiiil1:
## Summary
- reuse `isUserBootloaderSupported()` before sending the user-bootloader query
- preserve the bootloader-version guard while also enforcing the NETWORK-only requirement
- avoid sending an unsupported request to USB bootloaders with a sufficiently new version

## Validation
- configured and built `depthai-core` successfully on macOS with optional OpenCV, dynamic calibration, events manager, and MP4 support disabled
- verified `src/device/DeviceBootloader.cpp` compiles and the final `libdepthai-core.a` links
- `clang-format --dry-run --Werror src/device/DeviceBootloader.cpp`
- `git diff --check`

Hardware validation was not run because an OAK device is not available locally.

Closes #1102

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved bootloader capability checks to ensure requests are sent only when the bootloader type and version are supported.

### DIFF
--- a/src/device/DeviceBootloader.cpp
+++ b/src/device/DeviceBootloader.cpp
@@ -703,8 +703,7 @@ bool DeviceBootloader::isUserBootloaderSupported() {
 }
 
 bool DeviceBootloader::isUserBootloader() {
-    // Check if bootloader version is adequate
-    if(getVersion() < Version(Request::IsUserBootloader::VERSION)) {
+    if(!isUserBootloaderSupported()) {
         return false;
     }
 


aljazkonec1:
Thanks, but not relevant as issue is on DepthAI v2

PR #1935, opened 2026-08-11
enforced transform validation in stereo_depth_remap.cppand reduced lo…

ViacheslavTelelyukhin:
## Purpose
stereo_depth_remap.cpp now throws if an imgFrame fails validateTransformations
camera_multiple_outputs.py now only logs every 60 frames instead of every frame

The author is taking the responsibility for the contribution: Probably

## Summary by CodeRabbit

* **Bug Fixes**
  * Invalid frame transformations now stop processing with a clear runtime error instead of continuing silently.
  * Reduced camera metadata log frequency to improve console readability and runtime performance.


aljazkonec1:
Thanks!
You need to update the C++ example as well `examples/cpp/Camera/camera_multiple_outputs.cpp`
Also need to reduce verbosity on py_camera_multiple_outputs

aljazkonec1 (approved):

PR #1939, opened 2026-08-13
Bugfix/detection parser test stability

ViacheslavTelelyukhin:
## Purpose
detection_parser_test.cpp would behave differently depending on enviroment image/video codecs.

## Specification
Forced the VideoPlayer (inside replay node) to throw when it can't open a video. Otherwise replay node would simply block forever.
Switched detection_parser_test/"DetectionParser segmentation mask test" to using neural network data instead of an image as test data.

Added 2 DEPTHAI_FETCH_ARTIFACTS to tests/CMakeLists which stops them from configuring successfully

## Deployment Plan
Need to upload fetch ertefacts before merging

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved video playback startup errors by providing a clear message when a video cannot be opened.

* **Tests**
  * Updated segmentation validation to use representative neural-network data and metadata directly.
  * Improved test reliability by removing unnecessary image loading and preprocessing steps.

### DIFF
--- a/src/opencv/RecordReplay.cpp
+++ b/src/opencv/RecordReplay.cpp
@@ -256,8 +256,9 @@ void VideoPlayer::init(const std::string& filePath) {
         throw std::runtime_error("VideoPlayer file does not exist: " + filePath);
     }
     cvReader = std::make_unique<cv::VideoCapture>();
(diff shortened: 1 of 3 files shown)


aljazkonec1 (approved):
Thanks! you can merge once the affected test passes

PR #1942, opened 2026-08-14
Add device ID to image transformations

pheec:
Adds the source device ID to ImgTransformation.
The device ID is stored directly on the transformation instead of in Extrinsics, because extrinsics are part of per-camera EEPROM calibration data. Storing device-level information there would duplicate it for every camera.
This PR also:
Serializes and deserializes the device ID.
Exposes it through the C++ and Python APIs.
Propagates it through rectification.
Prevents remapping and alignment between transformations with different known device IDs.
Treats an empty device ID as unknown for compatibility with older replay datasets.
Adds tests for serialization, equality, alignment, remapping, and replay compatibility.

## Summary by CodeRabbit

- **New Features**
  - Extrinsics now track an optional destination device ID.
  - Added coordinate-system compatibility checks for transformations.
  - Python bindings expose destination device IDs and compatibility validation.
  - Transformation metadata preserves device information during serialization.

- **Bug Fixes**
  - Image remapping now rejects incompatible device or camera-coordinate transformations instead of processing invalid data.
  - Alignment checks correctly report incompatible coordinate systems.
  - Improved extrinsics comparisons and transformation validation.


include/depthai/common/ImgTransformations.hpp
```diff
@@ -24,6 +26,7 @@ struct ImgTransformation {
     CameraModel distortionModel = CameraModel::Perspective;
     std::vector<float> distortionCoefficients;
     Extrinsics extrinsics = {};
+    std::string deviceId;
```
aljazkonec1: In the context of multi device pipelines, how would deviceId be used? Looking at the current code, it throws any time the deviceId is different. Is there a reason you decided against deviceId meaning the target device? Is there a reason why we need source device ID at all?

src/pipeline/datatype/ImgTransformations.cpp
```diff
@@ -17,8 +17,25 @@
...
+}
+
+void validateSameKnownDevice(const ImgTransformation& lhs, const ImgTransformation& rhs, const std::string& operation) {
+    if(differentKnownDeviceIds(lhs, rhs)) {
+        throw std::runtime_error(
+            fmt::format("Cannot {} ImgTransformations from different devices: '{}' and '{}'.", operation, lhs.getDeviceId(), rhs.getDeviceId()));
+    }
+}
+
+}  // namespace
```
aljazkonec1: lets remove the codex generated anonymous namespace :)

src/pipeline/datatype/ImgTransformations.cpp
```diff
@@ -17,8 +17,25 @@
 #include "pipeline/utilities/Alignment/AlignmentUtilities.hpp"
 namespace dai {
 
+namespace {
+
 constexpr float ROUND_UP_EPS = 1e-3f;
 
+bool differentKnownDeviceIds(const ImgTransformation& lhs, const ImgTransformation& rhs) {
+    const auto& lhsDeviceId = lhs.getDeviceId();
```
aljazkonec1: lhs and rhs arent general parameter names. From my understanding we will be using deviceId for multi-device cases so this naming really does not fit here

src/pipeline/datatype/ImgTransformations.cpp
```diff
@@ -50,6 +67,7 @@ inline bool RRinRR(const dai::RotatedRect& in, const dai::RotatedRect& out) {
 }
 
 dai::Point2f interSourceFrameTransform(dai::Point2f sourcePt, const ImgTransformation& from, const ImgTransformation& to) {
+    validateSameKnownDevice(from, to, "remap between");
```
aljazkonec1: I would expand the function to also validate the toCameraSockets like it is done in the [below lines](https://github.com/luxonis/depthai-core/blob/9266b7658bdae7d11b607bdf7b23aedd8b417cec/src/pipeline/datatype/ImgTransformations.cpp#L78-L81). Also if deviceId means target device to where it is pointing to, then we should make it a member funciton of Extrnisics

aljazkonec1 (changes requested):
Thanks, I'm struggling to understand the usefulness of having source device ID stored? This PR is then not meant for multi device support no?

include/depthai/common/Extrinsics.hpp
```diff
@@ -140,6 +147,14 @@ struct Extrinsics {
      */
     bool isEqualExtrinsics(const Extrinsics& other, float epsilon = 1e-6f) const;
 
+    /**
+     * Check whether these extrinsics can be expressed relative to the same target coordinate system as another Extrinsics object.
+     * Unknown device IDs and AUTO camera sockets are treated as compatible for backwards compatibility.
+     * @param other The other Extrinsics object to compare with
+     * @return true if no known part of the target coordinate system differs, false otherwise
+     */
+    bool hasCompatibleCoordinateSystem(const Extrinsics& other) const;
```
aljazkonec1, line 156: Nit: other functions have the parameter named as `to` instead of `other`

tests/CMakeLists.txt
```diff
@@ -452,6 +452,7 @@ dai_set_test_labels(naming_test onhost ci)
 
 # Image transformations test
 dai_add_test(image_transformations_test src/onhost_tests/image_transformations_test.cpp)
+target_include_directories(image_transformations_test PRIVATE ${CMAKE_SOURCE_DIR}/include/depthai)
```
aljazkonec1, line 455: Why is this header search neede? all includes should contain the `depthai/` prefix

aljazkonec1 (commented):
Thanks, left some comments but looks much better!

src/pipeline/datatype/ImgTransformations.cpp
```diff
@@ -50,6 +50,9 @@ inline bool RRinRR(const dai::RotatedRect& in, const dai::RotatedRect& out) {
 }
 
 dai::Point2f interSourceFrameTransform(dai::Point2f sourcePt, const ImgTransformation& from, const ImgTransformation& to) {
```
aljazkonec1, line 52: Nit: Extrinsics are copied 4 times in this function. Copy extrinsics once at the top and reuse them.

bindings/python/src/pipeline/CommonBindings.cpp
```diff
@@ -573,6 +574,10 @@ void CommonBindings::bind(pybind11::module& m, void* pCallstack) {
              py::arg("unit") = LengthUnit::CENTIMETER,
              DOC(dai, Extrinsics, getTranslationVector))
         .def("isEqualExtrinsics", &Extrinsics::isEqualExtrinsics, py::arg("other"), py::arg("epsilon") = 1e-6f, DOC(dai, Extrinsics, isEqualExtrinsics))
+        .def("hasCompatibleCoordinateSystem",
+             &Extrinsics::hasCompatibleCoordinateSystem,
+             py::arg("other"),
```
aljazkonec1: c++ arg name is "to"

aljazkonec1 (approved):
Thanks! You can merge once you bump FW and the tests pass

fallenmi (changes requested):
This exact head is not backward-compatible with the existing binary `Extrinsics` schema. `DEPTHAI_SERIALIZE_OPTIONAL` only makes the JSON fields optional; its libnop side still expands to fixed-arity `NOP_STRUCTURE`. Appending `toDeviceId` therefore changes the encoded member count.

I reproduced this with an exact-base/exact-head C++17 oracle. A serializer compiled from base `733d46a50c2994e94f41cb865a64f00da7ead587` produced a 40-byte `Extrinsics` payload and exited successfully. A deserializer compiled from exact head `fc4053d3de0293b4635ff418ddebb4f271cdd319` throws `Invalid Member Count` on that payload. As a control, the exact-head serializer produces 42 bytes and the exact-head reader deserializes those bytes. Stored binary data written with the base schema therefore cannot be read at this head.

Please add a versioned/backward-compatible libnop decoding path and a regression fixture produced by the old five-member schema. A same-head round trip does not cover this compatibility boundary.

The final commit also changes `hasCompatibleCoordinateSystem()` from unknown-as-wildcard to direct device-ID equality. Compiling the exact upstream method shows `known <-> empty` was `true/true` at the prior PR head `db22dce6a`, but is `false/false` at the current head. That contradicts the public header contract and makes the PR's own unknown-device and replay assertions false. Please restore the empty-ID compatibility behavior, or explicitly change the contract and tests if rejecting legacy metadata is intentional; either way, the old binary decode failure above still needs a migration path.

AI disclosure: I used OpenAI Codex to inspect the diff, build and run the exact-commit compatibility oracles, and draft this review. I verified the source paths, commit IDs, outputs, and conclusion.

PR #1945, opened 2026-08-17
standardizing logs

ViacheslavTelelyukhin:
## Purpose
One of three pull requests across depthai-core, depthai-rvc2 and depthai-device-kb aiming to make logs of RVC2 and RVC4 devices as close as possible.

## Deployment Plan
All 3 pull requests should be merged at once whenever they are merged.

## AI Usage
Searched for log inconsitencies: Codex:5.6 Luna - high

## Summary by CodeRabbit

- **New Features**
  - Added consistent per-stage processing timing information across pipeline nodes, covering input, processing, and output phases.
  - Added clearer startup diagnostics showing whether processing runs on the host or device.

- **Bug Fixes**
  - Improved invalid detection-input errors by identifying accepted input types and confirming when tracking is skipped.
  - Corrected the “amplitude” warning message.

- **Improvements**
  - Standardized timing and execution-location reporting for more consistent troubleshooting and performance analysis.


include/depthai/pipeline/DeviceNode.hpp
```diff
@@ -1,5 +1,8 @@
 #pragma once
 
+#include <spdlog/async_logger.h>
```
aljazkonec1: spdlog is private dependency and this is a real include taht will fail to compile on external consumers. You can add a forward declaration of spdlog and then include it DeviceNode.cpp. See example of forward declaration [here](https://github.com/luxonis/depthai-core/blob/3ef54b71e4abcc5cfa9133657a3e96884f859e4f/include/depthai/pipeline/node/PointCloud.hpp#L21-L24)

src/pipeline/node/internal/PipelineEventAggregation.cpp
```diff
@@ -638,6 +638,7 @@ std::tuple<std::shared_ptr<PipelineState>, bool> makeOutputState(PipelineEventHa
 
 void PipelineEventAggregation::run() {
     auto& logger = pimpl->logger;
+    logger->info("{} running on {}.", this->getName(), runOnHost() ? "host" : "device");
```
aljazkonec1: Can you remove logging as this is internal node.

src/beta/node/ImageOutputParser.cpp
```diff
@@ -301,7 +301,7 @@ bool ImageOutputParser::runOnHost() const {
 
 void ImageOutputParser::run() {
     auto& logger = ThreadedNode::pimpl->logger;
-    logger->debug("ImageOutputParser started");
+    logger->info("{} running on {}.", this->getName(), runOnHost() ? "host" : "device");
```
aljazkonec1: Using `runOnHost` here will produce the true / false statement because on device we override this function to always return true becasue device and host share the same src implementation of the parser. You have to use `runOnHostVar` variable here. Same goes for all other nodes in the diff.

src/pipeline/node/DynamicCalibrationNode.cpp
```diff
@@ -1017,6 +1017,7 @@ DynamicCalibration::ErrorCode DynamicCalibration::doWork(std::chrono::steady_clo
 }
 
 void DynamicCalibration::run() {
+    logger->info("{} running on {}.", this->getName(), runOnHost() ? "host" : "device");
```
aljazkonec1: Internal node. I would remove

src/pipeline/node/SegmentationParser.cpp
```diff
@@ -35,7 +35,8 @@ void SegmentationParser::buildInternal() {
         auto platform = device->getPlatform();
         if(platform == Platform::RVC2) {
             setRunOnHost(true);
-            std::cout << "SegmentationParser: For RVC2 platform, running on host." << std::endl;
+            auto& logger = ThreadedNode::pimpl->logger;
+            if (logger) logger->info("SegmentationParser: For RVC2 platform, running on host.");
```
aljazkonec1: Delete "SegmentationParser" in the string. Add else statement that still prints to `cout`

aljazkonec1 (changes requested):
Thanks! Left some design comments. Also some general comments that are useful for future PRs as well:
- merge develop branch into this branch (do the same on device side)
- You need to run `git submodule update --init --recursive` to [update all submodules ](https://github.com/luxonis/depthai-core#building) because your PR wants revert `pybing11_opencv_numpy` to previous commit
- need to run `cmake --build build --target clangformat`
- You need to bump the device side depthai-core commit hash to latest commit
- You need to bump the commit hashes in `cmake/Depthai/DepthaiDeviceRVC4Config.cmake` and `cmake/Depthai/DepthaiDeviceSideConfig.cmake` to the latest commit hashes and commit those here so I can build without needing to build device side aswell

aljazkonec1 (changes requested):
.

PR #1947, opened 2026-08-17
Add Stitching node

pheec:
## Summary

Introduce the Stitching node for combining multiple synchronized image streams
into a single output image.

The node runs on the host by default and can run on RVC4 by calling
`setRunOnHost(false)`.

## Features

- Accept a configurable number of image inputs.
- Synchronize inputs through an internal Sync node.
- Support two stitching modes:
  - Panorama stitching based on overlapping image content.
  - Calibration-aware planar projection onto a configurable plane.
- Support spherical, cylindrical, and pinhole panorama camera models.
- Support continuous registration or selecting the best transform from several
  estimation frames.
- Configure panorama confidence, seam finding, and maximum output dimensions.
- Configure planar projection plane, virtual camera, range, view size, and
  minimum incidence angle.
- Preserve image metadata and transformations on stitched output.
- Support host execution with OpenCV.
- Support device execution on RVC4 through `setRunOnHost(false)`.
- Make all device configuration and dynamic input information serializable.
- Keep the device-facing API available in builds without host OpenCV support.
- Reject device execution on platforms other than RVC4.

## Examples

Add Python examples for:

- Panorama stitching.
- Planar projection stitching.

## Testing

Add coverage for:

- Panorama registration and composition.
- Continuous and fixed-transform operation.
- Candidate selection and panorama size limits.
- Planar projection geometry and validation.
- Property serialization and firmware deserialization.
- Dynamic input reconstruction.
- RVC4-only device execution.
- Builds without host OpenCV support.

## New Features
- Added an experimental stitching node for creating panoramas and planar projections from multiple camera feeds.
- Supports host execution and RVC4 device execution, configurable camera models, seam finding, projection settings, and transform management.
- Added Python bindings and examples for panorama and planar stitching.
- Added validation for inputs, calibration, canvas sizes, and stitching configuration.

## Bug Fixes
- Improved const access for managed objects.
- Added clearer handling when host stitching is unavailable without OpenCV.


aljazkonec1:
If we are not certain on the API for this node I would prefer to have it in the beta namespace, where all the new parsers are. In this case we can have it out in the next release of dai. Thoughts @CenekAlbl ?

CMakeLists.txt
```diff
@@ -201,7 +201,7 @@ set(THIRDPARTY_OPENCV_LIBRARIES "" CACHE STRING "Optional libraries to link Open
 set(TARGET_OPENCV_NAME ${PROJECT_NAME}-opencv)
 set(TARGET_OPENCV_ALIAS opencv)
 if(DEPTHAI_OPENCV_SUPPORT)
-    set(REQUIRED_OPENCV_LIBRARIES "opencv_core" "opencv_imgproc" "opencv_videoio" "opencv_highgui" "opencv_calib3d")
+    set(REQUIRED_OPENCV_LIBRARIES "opencv_core" "opencv_imgproc" "opencv_videoio" "opencv_highgui" "opencv_calib3d" "opencv_features2d")
```
aljazkonec1: this will make `opencv_features2d` required for all builds, even if not building the stitching node. `opencv_features2d` should be included **only** if DEPTHAI_BUILD_BETA is true.

CMakeLists.txt
```diff
@@ -536,6 +541,17 @@ set(TARGET_OPENCV_SOURCES
...
+if(DEPTHAI_BUILD_BETA AND DEPTHAI_HAVE_OPENCV_STITCHING)
+    list(APPEND TARGET_OPENCV_SOURCES
+        src/beta/node/host/Stitching.cpp
+        src/beta/node/host/FixedPanoramaCompositor.cpp
+        src/beta/node/host/PlanarStitcher.cpp
+    )
+    list(APPEND DEPTHAI_OPENCV_LIBRARIES "opencv_stitching")
+elseif(DEPTHAI_BUILD_BETA)
+    list(APPEND TARGET_CORE_SOURCES src/beta/node/StitchingNoOpenCV.cpp)
+endif()
```
aljazkonec1: You dont need to have different cpp targets based on having openCV support. You can add a `#ifdef` statement in the run function based on OpenCV support. examples of how to handle it: `src/beta/utilities/PPText/PPTextUtils.cpp` or better yet `src/pipeline/node/ImageAlign.cpp`. You can then delete the `StitchingNoOpenCV` files

src/beta/utilities/Stitching/Stitching.cpp
```diff
@@ -0,0 +1,440 @@
+#include "depthai/beta/node/Stitching.hpp"
```
aljazkonec1, line 1: No need for the `host/Stitching.cpp`, you can define all of these functions in the `Stitching.cpp` or in a utils file in `src/beta/utilities`

src/beta/utilities/Stitching/PlanarStitcher.hpp
```diff
@@ -0,0 +1,133 @@
...
+ *
+ * Nothing is registered from the image content: every pixel of the rendered image is the intersection of a ray of the
+ * virtual camera with the plane, projected back into each input through the calibration carried by the input's
+ * `ImgTransformation`. Both the plane and the virtual camera live in the origin frame the input extrinsics are
+ * expressed in, so all inputs have to share that origin camera socket.
+ *
+ * The geometry is assumed to be fixed: the projection maps, the seams and the exposure gains are built from the first
+ * group and reused, so a frame only costs a remap and a blend. Call `reset()` to rebuild them.
+ */
+class PlanarStitcher {
```
aljazkonec1, line 30: Should not be `node` namespace as its not a node. If its a utility class, I would move it to `src/beta/utilities/Stitching/PlanarStitcher.hpp` or into `src/beta/impl/Stitching/PlanarStitcher.hpp`

src/beta/utilities/Stitching/FixedPanoramaCompositor.hpp
```diff
@@ -0,0 +1,67 @@
...
+namespace beta {
+namespace node {
+
+/**
+ * Composes a panorama whose registered camera geometry does not change.
+ *
+ * Projection maps, image regions, validity and seam masks, and exposure parameters are prepared from the first image
+ * group and reused. Later groups only resize (when requested), remap, compensate and blend.
+ */
+class FixedPanoramaCompositor {
```
aljazkonec1, line 23: Should not be `node` namespace as its not a node. If its a utility class, I would move it to `src/beta/utilities/Stitching/FixedPanoramaCompositor.hpp` or into `src/beta/impl/Stitching/FixedPanoramaCompositor.hpp`

src/utility/PimplImpl.hpp
```diff
@@ -20,9 +20,19 @@ T* Pimpl<T>::operator->() {
...
 template <typename T>
 T& Pimpl<T>::operator*() {
     return *m.get();
 }
 
+template <typename T>
+const T& Pimpl<T>::operator*() const {
+    return *m.get();
+}
+
```
aljazkonec1, line 37: Why is this change needed?

CMakeLists.txt
```diff
@@ -212,6 +212,10 @@ if(DEPTHAI_OPENCV_SUPPORT)
     if(OPENCV_SUPPORT_AVAILABLE)
         # Add public compile definition indicating that OpenCV support is available
         set(DEPTHAI_HAVE_OPENCV_SUPPORT ON)
+        set(DEPTHAI_OPENCV_LIBRARIES ${REQUIRED_OPENCV_LIBRARIES})
+        if("opencv_stitching" IN_LIST OpenCV_LIBS)
+            set(DEPTHAI_HAVE_OPENCV_STITCHING ON)
+        endif()
```
aljazkonec1: Same as above goes for `opencv_stitching`, add only if DEPTHAI_BUILD_BETA true

CMakeLists.txt
```diff
@@ -212,6 +212,10 @@ if(DEPTHAI_OPENCV_SUPPORT)
     if(OPENCV_SUPPORT_AVAILABLE)
         # Add public compile definition indicating that OpenCV support is available
         set(DEPTHAI_HAVE_OPENCV_SUPPORT ON)
+        set(DEPTHAI_OPENCV_LIBRARIES ${REQUIRED_OPENCV_LIBRARIES})
```
aljazkonec1: Needed? cant we append to REQUIRED_OPENCV_LIBRARIES ?

tests/CMakeLists.txt
```diff
@@ -346,7 +346,7 @@ function(dai_add_test test_name test_src)
 
     # Link to core and Catch2 testing framework
     set(DEPTHAI_TARGET depthai::core)
-    if(NOT DEPTHAI_MERGED_TARGET)
+    if(NOT DEPTHAI_MERGED_TARGET AND TARGET depthai::opencv)
```
aljazkonec1, line 349: Why is this change needed?

include/depthai/utility/Pimpl.hpp
```diff
@@ -17,7 +17,9 @@ class Pimpl {
     Pimpl(Args&&...);
     ~Pimpl();
     T* operator->();
+    const T* operator->() const;
```
aljazkonec1, line 20: See comment [here](https://github.com/luxonis/depthai-core/pull/1947#discussion_r3846112541)

src/beta/utilities/Stitching/StitchingCompositing.hpp
```diff
@@ -0,0 +1,95 @@
+#pragma once
```
aljazkonec1, line 1: I would move it to `src/beta/utilities/Stitching/StitchingCompositing.hpp`

src/beta/node/StitchingPlatform.hpp
```diff
@@ -0,0 +1,17 @@
...
+namespace stitching {
+
+constexpr bool isDevicePlatformSupported(Platform platform) {
+    return platform == Platform::RVC4;
+}
+
+}  // namespace stitching
+}  // namespace node
+}  // namespace beta
+}  // namespace dai
```
aljazkonec1: Is this needed? the base class BetaNode.cpp already handles the platform in `buildStage1()` . Depthai v3 cant run on RVC3 so you can remove that from the test all together.

include/depthai/beta/node/Stitching.hpp
```diff
@@ -0,0 +1,247 @@
...
+    void setSeamFinder(SeamFinder finder);
+    SeamFinder getSeamFinder() const;
+
+    void buildInternal() override;
+
+   private:
+    void run() override;
+    void initializeHostState();
+    void invalidateHostState();
+    void initializeInputNames(size_t numInputs);
```
aljazkonec1, line 232: I'm guessing these can be removed once the NoOpenCv src gets removed?

include/depthai/beta/node/Stitching.hpp
```diff
@@ -0,0 +1,247 @@
...
+    void setPanoConfidenceThreshold(double threshold);
+    double getPanoConfidenceThreshold() const;
+
+    void setSeamFinder(SeamFinder finder);
+    SeamFinder getSeamFinder() const;
+
+    void buildInternal() override;
+
+   private:
+    void run() override;
```
aljazkonec1: Nit: `run` is public in all nodes

aljazkonec1 (changes requested):
Thanks! Left some comments. Mostly design based to keep the structure the same as other nodes

src/beta/utilities/Stitching/Stitching.cpp
pheec: Do you mean moving the entire OpenCV implementation from host/Stitching.cpp into src/beta/node/Stitching.cpp behind #ifdef DEPTHAI_HAVE_OPENCV_SUPPORT, including the necessary target-split changes?
Or should the implementation remain a separate OpenCV-only source file, possibly under src/beta/utilities? Which structure would you prefer?

src/utility/PimplImpl.hpp
pheec: operator->() const is required because Stitching has const getter methods that access the implementation through impl->....
operator*() const is the matching const dereference overload. It may not be used directly by Stitching, but keeping it makes Pimpl behave consistently with std::unique_ptr and spimpl::impl_ptr, which are used elsewhere in the library.

tests/CMakeLists.txt
pheec: This check is needed because depthai::opencv is not created when OpenCV support is disabled. In that case, the tests use depthai::core instead.
This pattern is also used elsewhere, for example in tests/integration/CMakeLists.txt:

```
if(TARGET depthai::opencv)
    target_link_libraries(integration-test-opencv PRIVATE depthai::opencv)
endif()
```

src/beta/node/StitchingPlatform.hpp
pheec: Removed the RVC3 check because DepthAI v3 does not support RVC3. The platform helper is still needed because BetaNode only handles the RVC2 host fallback, while Stitching still needs to verify that device execution is supported on RVC4.

include/depthai/beta/node/Stitching.hpp
pheec: No, these methods are not only for the no-OpenCV implementation.
- initializeInputNames() is used when building the node.
- invalidateHostState() is called by the configuration setters and resetTransform().
- initializeHostState() is used by the OpenCV host implementation to create the internal state.

src/beta/utilities/Stitching/Stitching.cpp
aljazkonec1: Either structure works for me, for the sake of cleanliness they can be moved into utilities (they still need to be guarderd by #ifdef )

src/utility/PimplImpl.hpp
aljazkonec1: Yeah they arent needed anymore for this PR but they might be useful in other nodes down the line. thanks

src/beta/node/StitchingPlatform.hpp
aljazkonec1: Why does it need to verify that? Its already know if it runOnHostVar is set to `true` or `false` no?

aljazkonec1: Should the node not even run on device? why not?

src/beta/node/Stitching.cpp
```diff
@@ -0,0 +1,252 @@
...
+    if(device && !runOnHostVar && !stitching::isDevicePlatformSupported(device->getPlatform())) {
+        throw std::runtime_error("Stitching node is only supported on RVC4 devices. Use setRunOnHost(true) instead.");
+    }
+    return properties;
+}
+
+void Stitching::buildInternal() {
+    initializeHostState();
+    sync->out.link(inSync);
+    sync->setRunOnHost(true);
```
aljazkonec1: You cant unconditionally set running on host to true here. What if the stitching node is running on device? then you'll have device -> host sync -> device stitching link, making the pipeline very slow and laggy

src/beta/node/StitchingPlatform.hpp
pheec: The node should still run on the device on RVC4. Since BetaNode already switches RVC2 to host execution and RVC3 is not supported in DepthAI v3, I’ll remove the extra platform check.

aljazkonec1 (approved):
Thanks! need to bump FW, other than that looks good.

PR #1949, opened 2026-08-18
check only yolo output count when checking number of strides

ViacheslavTelelyukhin:
## Purpose
There was an incorrect check that would look for all outputs instead of just yolo outputs when checking amount of strides

## Specification
Made it only look at yolo outputs

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved YOLO output detection by counting only outputs explicitly identified as YOLO results.
  * Corrected stride validation and configuration when archives contain unrelated outputs.
  * Fixed Python access to the classification sequence parser’s input configuration, preserving its documented read-only behavior.

### DIFF
--- a/src/pipeline/node/DetectionParser.cpp
+++ b/src/pipeline/node/DetectionParser.cpp
@@ -149,7 +149,12 @@ void DetectionParser::configureYOLONetworkParser(DetectionParserOptions& parser,
         parser.strides = {1};
     }
     if(metadata.strides) {
-        const size_t numYoloOutputs = metadata.yoloOutputs ? metadata.yoloOutputs->size() : (head.outputs ? head.outputs->size() : 0);
+        size_t numYoloOutputs = 0;
+        if(metadata.yoloOutputs) numYoloOutputs = metadata.yoloOutputs->size();
+        else if(head.outputs.has_value()) {
+            for (const auto& name : *head.outputs)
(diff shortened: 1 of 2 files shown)


src/pipeline/node/DetectionParser.cpp
```diff
@@ -149,7 +149,12 @@ void DetectionParser::configureYOLONetworkParser(DetectionParserOptions& parser,
         parser.strides = {1};
     }
     if(metadata.strides) {
-        const size_t numYoloOutputs = metadata.yoloOutputs ? metadata.yoloOutputs->size() : (head.outputs ? head.outputs->size() : 0);
+        size_t numYoloOutputs = 0;
+        if(metadata.yoloOutputs) numYoloOutputs = metadata.yoloOutputs->size();
```
aljazkonec1: please run clang format as I'm pretty sure this will fail style check

aljazkonec1 (approved):
Thanks! Did you test it out on the provided MRE, one segmentation YOLo, one regular yolo and one keypoint yolo model? If it works on all, you can merge once style and detection_parser_test passes

ViacheslavTelelyukhin:
I tested it only with the provided model in the bug report. Is there a list of models I can maybe use to test cases like this? I'm not sure if digging through examples for models is a good approac and I couldn't find any list in the codebase.

ViacheslavTelelyukhin:
Tested again with more yolo models from hub. All I tried worked

PR #1950, opened 2026-08-18
Validate camera intrinsics consistently

pheec:
## Summary

Expand camera intrinsics validation so health checks reject calibration data that Camera nodes and downstream consumers cannot safely use.

Validation now requires:

- An exact, non-ragged 3×3 matrix
- Finite values
- Positive `fx` and `fy`
- Zero lower-triangular elements
- A homogeneous scale of `1`
- Valid, nonzero calibration dimensions representable by `CameraInfo`

The same validation is now applied when reading and setting intrinsics, preventing inconsistent calibration data from entering through the setter.

Principal points outside the image and an optional skew term remain supported to avoid rejecting unusual but valid calibrations.

## Testing

- Added unit coverage for malformed, non-finite, noncanonical, and valid unusual matrices
- `calibration_handler_test`: 68 test cases and 1,541 assertions passed
- Full on-host suite: 34/34 tests passed
- Verified checked-in calibration JSON files and the holistic recording’s `camera_info.json` against the new rules

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of camera calibration data, including intrinsic matrices and frame dimensions.
  * Invalid, malformed, singular, non-finite, negative, or structurally incorrect calibration values are now rejected.
  * Valid skewed intrinsic matrices continue to be accepted.
* **Tests**
  * Added coverage for calibration validation across default and manually configured camera settings.

### DIFF
--- a/src/device/CalibrationHandler.cpp
+++ b/src/device/CalibrationHandler.cpp
@@ -32,19 +32,8 @@ namespace {
 constexpr size_t kImuCalibrationRowCount = 3;
 constexpr size_t kImuCalibrationColumnCount = 4;
 
-bool isValidIntrinsicsMatrix(const std::vector<std::vector<float>>& intrinsics) {
-    if(intrinsics.size() != 3 || intrinsics[0].size() != 3 || intrinsics[1].size() != 3 || intrinsics[2].size() != 3) {
-        return false;
-    }
-
-    for(const auto& row : intrinsics) {
-        if(!std::all_of(row.begin(), row.end(), [](float value) { return std::isfinite(value); })) {
-            return false;
-        }
-    }
-
-    return intrinsics[0][0] > 0.0f && intrinsics[1][1] > 0.0f && intrinsics[1][0] == 0.0f && intrinsics[2][0] == 0.0f && intrinsics[2][1] == 0.0f
-           && intrinsics[2][2] == 1.0f;
+bool isValidCalibrationDimension(float dimension) {
+    return std::isfinite(dimension) && dimension > 0.0f && dimension <= std::numeric_limits<uint16_t>::max() && std::trunc(dimension) == dimension;
 }
 
 std::optional<std::array<float, 3>> lookupHousingEntry(const std::string& productName, HousingCoordinateSystem housingCs) {
@@ -1053,7 +1042,7 @@ void CalibrationHandler::setProductName(const std::string& productName) {
 
 void CalibrationHandler::setCameraIntrinsics(CameraBoardSocket cameraId, const std::vector<std::vector<float>>& intrinsics, int width, int height) {
     if(!isValidIntrinsicsMatrix(intrinsics)) {
-        throw std::runtime_error("Invalid Intrinsic Matrix entered!!");
+        throw std::runtime_error("Invalid Intrinsic Matrix entered: " + nlohmann::json(intrinsics).dump());
     }
     if(width <= 0 || height <= 0 || width > std::numeric_limits<uint16_t>::max() || height > std::numeric_limits<uint16_t>::max()) {
         throw std::runtime_error("Invalid calibration resolution entered!!");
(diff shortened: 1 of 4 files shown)


src/device/CalibrationHandler.cpp
```diff
@@ -31,6 +32,21 @@ namespace {
 constexpr size_t kImuCalibrationRowCount = 3;
 constexpr size_t kImuCalibrationColumnCount = 4;
 
+bool isValidIntrinsicsMatrix(const std::vector<std::vector<float>>& intrinsics) {
```
MaticTonin: Mybe best to move it to matrixOps.cpp since its a generic function

src/device/CalibrationHandler.cpp
```diff
@@ -1037,13 +1052,12 @@ void CalibrationHandler::setProductName(const std::string& productName) {
...
 void CalibrationHandler::setCameraIntrinsics(CameraBoardSocket cameraId, const std::vector<std::vector<float>>& intrinsics, int width, int height) {
-    if(intrinsics.size() != 3 || intrinsics[0].size() != 3) {
-        throw std::runtime_error("Intrinsic Matrix size should always be 3x3 ");
-    }
-
-    if(intrinsics[1][0] != 0 || intrinsics[2][0] != 0 || intrinsics[2][1] != 0) {
+    if(!isValidIntrinsicsMatrix(intrinsics)) {
         throw std::runtime_error("Invalid Intrinsic Matrix entered!!");
     }
+    if(width <= 0 || height <= 0 || width > std::numeric_limits<uint16_t>::max() || height > std::numeric_limits<uint16_t>::max()) {
```
MaticTonin, line 1047: I think adding upper limit is not required.

MaticTonin (changes requested):
Overall looks fine, small comments

src/device/CalibrationHandler.cpp
```diff
@@ -1037,13 +1052,12 @@ void CalibrationHandler::setProductName(const std::string& productName) {
 }
 
 void CalibrationHandler::setCameraIntrinsics(CameraBoardSocket cameraId, const std::vector<std::vector<float>>& intrinsics, int width, int height) {
-    if(intrinsics.size() != 3 || intrinsics[0].size() != 3) {
-        throw std::runtime_error("Intrinsic Matrix size should always be 3x3 ");
-    }
-
-    if(intrinsics[1][0] != 0 || intrinsics[2][0] != 0 || intrinsics[2][1] != 0) {
+    if(!isValidIntrinsicsMatrix(intrinsics)) {
         throw std::runtime_error("Invalid Intrinsic Matrix entered!!");
```
aljazkonec1: lets print the matrix when we throw

aljazkonec1 (approved):
Thanks! one small comment + everything that @MaticTonin mentioned :)

src/device/CalibrationHandler.cpp
pheec: I kept the check because CameraInfo stores width and height as uint16_t, so larger values could be silently truncated. Im not sure, please check it out @aljazkonec1

src/device/CalibrationHandler.cpp
```diff
@@ -1074,6 +1063,9 @@ void CalibrationHandler::setCameraIntrinsics(CameraBoardSocket cameraId, const s
 }
 
 void CalibrationHandler::setCameraIntrinsics(CameraBoardSocket cameraId, const std::vector<std::vector<float>>& intrinsics, Size2f frameSize) {
+    if(!isValidCalibrationDimension(frameSize.width) || !isValidCalibrationDimension(frameSize.height)) {
+        throw std::runtime_error("Invalid calibration resolution entered!!");
```
aljazkonec1: No need for two !!

src/device/CalibrationHandler.cpp
aljazkonec1: I'm fine either way

MaticTonin (approved):
LGTM now

PR #1951, opened 2026-08-19
report device in use more accurately

ViacheslavTelelyukhin:
## Purpose
deviceInUse was not being set to true for certain errors in device constructor if deviceInfo was out of date (had incorrect state)

## Specification
Matched more error messages

## AI Usage
None

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved device connection error handling by correctly identifying when a device is already in use.

### DIFF
--- a/src/device/HealthCheck.cpp
+++ b/src/device/HealthCheck.cpp
@@ -616,7 +616,7 @@ HealthCheckMetrics DeviceHealthCheck::run(const DeviceInfo& devInfo, const Healt
         device = std::make_shared<Device>(devInfo);
     } catch(const std::exception& ex) {
         const std::string errorMessage = ex.what();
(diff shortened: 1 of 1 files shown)


aljazdu:
Applied clang-format, removed redundant check. 
LGTM, but please test using the script which was provided in the thread (both RVC2 & RVC4 devices) - cover all three cases provided.

ViacheslavTelelyukhin:
I unfortuantely could not test on rvc2 usb because of my enviroment right now but behavior seems reasonable on rvc2 POE. Also all cases in proveded MRE are caller error and mostly can't reasonably be fixed without breaking changes.
Case 1 rvc4 and case 3 rvc2 for example are expected because the device no loger exists in that state and is therefore not found.

aljazdu (approved):
LGTM

PR #1958, opened 2026-08-24
Handle distorted images in PointCloud Node

pheec:
## What changed

- Compensate PointCloud depth rays for Perspective and Fisheye distortion.
- Cache undistorted rays for CPU deprojection.
- Fall back from GPU to CPU when distortion compensation is required.
- Reject unsupported distortion models with a clear error.
- Document that output distortion metadata describes the source image and is informational because points are already compensated.
- Add synthetic host and public-node regression tests for corrected optical rays.

## Validation

- `point_cloud_test`: 71 tests, 20,290 assertions passed.
- Public-node distortion test builds successfully and is intended for hardware/HIL execution.

## Summary by CodeRabbit

- **New Features**
  - Point-cloud generation now compensates for supported perspective and fisheye camera distortion.
  - Colored and depth-only point clouds preserve accurate geometry and color alignment.
  - Added a helper for displaying camera model names.

- **Bug Fixes**
  - Distorted inputs now use a safe CPU fallback when GPU processing does not support distortion.

- **Documentation**
  - Clarified that distortion metadata describes the source image, while generated 3D points are already corrected.

### DIFF
--- a/include/depthai/common/CameraModel.hpp
+++ b/include/depthai/common/CameraModel.hpp
@@ -1,11 +1,26 @@
 #pragma once
 
 #include <cstdint>
+#include <string>
 
 namespace dai {
 /**
  * Which CameraModel to initialize the calibration with.
  */
 enum class CameraModel : int8_t { Perspective = 0, Fisheye = 1, Equirectangular = 2, RadialDivision = 3 };
+
+inline std::string toString(CameraModel model) {
+    switch(model) {
+        case CameraModel::Perspective:
+            return "Perspective";
+        case CameraModel::Fisheye:
+            return "Fisheye";
+        case CameraModel::RadialDivision:
+            return "RadialDivision";
+        case CameraModel::Equirectangular:
+            return "Equirectangular";
+    }
+    return "Unknown";
+}
 
 }  // namespace dai
--- a/include/depthai/common/CameraModel.hpp
+++ b/include/depthai/common/CameraModel.hpp
@@ -1,7 +1,7 @@
 #pragma once
 
 #include <cstdint>
-#include <string>
+#include <string_view>
 
 namespace dai {
 /**
@@ -9,17 +9,14 @@ namespace dai {
  */
 enum class CameraModel : int8_t { Perspective = 0, Fisheye = 1, Equirectangular = 2, RadialDivision = 3 };
 
-inline std::string toString(CameraModel model) {
+[[nodiscard]] constexpr std::string_view toString(CameraModel model) {
     switch(model) {
-        case CameraModel::Perspective:
-            return "Perspective";
-        case CameraModel::Fisheye:
-            return "Fisheye";
-        case CameraModel::RadialDivision:
-            return "RadialDivision";
-        case CameraModel::Equirectangular:
-            return "Equirectangular";
+        case CameraModel::Perspective:    return "Perspective";
+        case CameraModel::Fisheye:         return "Fisheye";
+        case CameraModel::Equirectangular: return "Equirectangular";
+        case CameraModel::RadialDivision:  return "RadialDivision";
     }
+
     return "Unknown";
 }
 
--- a/include/depthai/common/CameraModel.hpp
+++ b/include/depthai/common/CameraModel.hpp
@@ -11,10 +11,14 @@ enum class CameraModel : int8_t { Perspective = 0, Fisheye = 1, Equirectangular 
 
 [[nodiscard]] constexpr std::string_view toString(CameraModel model) {
     switch(model) {
-        case CameraModel::Perspective:    return "Perspective";
-        case CameraModel::Fisheye:         return "Fisheye";
-        case CameraModel::Equirectangular: return "Equirectangular";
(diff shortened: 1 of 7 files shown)


src/pipeline/node/PointCloud.cpp
```diff
@@ -32,6 +35,24 @@
...
+        case CameraModel::RadialDivision:
+            return "RadialDivision";
+        case CameraModel::Equirectangular:
+            return "Equirectangular";
+    }
+    return "Unknown";
+}
+
+}  // namespace
+
```
JakubFara: Can we add this to `DistortionModel` class?

include/depthai/pipeline/node/PointCloud.hpp
```diff
@@ -61,13 +62,15 @@ class PointCloud : public DeviceNodeCRTP<DeviceNode, PointCloud, PointCloudPrope
         void useCPUMT(uint32_t numThreads);
         void useGPU(uint32_t device);
         void setIntrinsics(float fx, float fy, float cx, float cy, unsigned int width, unsigned int height);
+        void setDistortion(CameraModel model, std::vector<float> coefficients);
```
JakubFara, line 65: Move to private

JakubFara (commented):
Looks good: Two nits

include/depthai/common/CameraModel.hpp
```diff
@@ -1,11 +1,26 @@
...
+            return "Fisheye";
+        case CameraModel::RadialDivision:
+            return "RadialDivision";
+        case CameraModel::Equirectangular:
+            return "Equirectangular";
+    }
+    return "Unknown";
+}
+
+}  // namespace dai
```
JakubFara, line 27: This implementation is slightly better: it supports compile-time evaluation and avoids allocating a std::string by returning a std::string_view.

`constexpr` enables compile-time evaluation; it does not guarantee that every call is evaluated at compile time.

```
[[nodiscard]]
constexpr std::string_view toString(CameraModel model) {
    switch (model) {
        case CameraModel::Perspective:    return "Perspective";
        case CameraModel::Fisheye:         return "Fisheye";
        case CameraModel::Equirectangular: return "Equirectangular";
        case CameraModel::RadialDivision:  return "RadialDivision";
    }

    return "Unknown";
}
```

JakubFara (approved):
LGTM after changing this nit

include/depthai/pipeline/node/PointCloud.hpp
JakubFara: This is not needed since it is in `PointCloud::Impl` not in the `PointCloud` itself

src/pipeline/utilities/Alignment/AlignmentUtilities.cpp
```diff
@@ -315,6 +315,6 @@ inline float coeffAt(const std::vector<float>& coeffs, size_t idx) {
     return idx < coeffs.size() ? coeffs[idx] : 0.0f;
 }
 
-inline bool hasNonZeroDistortion(const std::vector<float>& coeffs) {
+bool hasNonZeroDistortion(const std::vector<float>& coeffs) {
```
aljazkonec1, line 318: why?

include/depthai/pipeline/node/PointCloud.hpp
```diff
@@ -101,6 +104,11 @@ class PointCloud : public DeviceNodeCRTP<DeviceNode, PointCloud, PointCloudPrope
         float lengthUnitMultiplier = DEFAULT_LENGTH_UNIT_MULTIPLIER;
 
         float fx = 0.0f, fy = 0.0f, cx = 0.0f, cy = 0.0f;
+        CameraModel distortionModel = CameraModel::Perspective;
+        std::vector<float> distortionCoefficients;
+        std::vector<Point2f> undistortedRays;
+        bool hasDistortion = false;
+        bool gpuDistortionFallbackWarned = false;
```
aljazkonec1, line 111: what do these coefficients pertain to? Is it only ToF ? doesnt Tof frame carry them in the ImgTransformations? Do they need to be public?

aljazkonec1 (approved):
Thanks, left some comments but looks OK

src/pipeline/utilities/Alignment/AlignmentUtilities.cpp
pheec: This helper was previously only used inside AlignmentUtilities.cpp. It is now also called from PointCloud.cpp, so the definition needs normal external linkage. Keeping inline while defining it only in the .cpp can lead to missing-symbol/linker issues.

include/depthai/pipeline/node/PointCloud.hpp
pheec: They describe the lens distortion of the camera that produced the depth frame, so they can come from StereoDepth or ToF.
We copy them from the frame’s ImgTransformation because PointCloud uses them to prepare the undistorted ray cache. They are in the private part of PointCloud::Impl, so they are not public API.

PR #1959, opened 2026-08-25
Fix/stereo default config

JakubFara:
## Purpose
On RVC4, in StereoDepth, the DEFAULT mode was not set if build() had not been called.

## Tests
New test added validating the fix

## Summary by CodeRabbit

* **New Features**
  * Added support for creating Stereo Depth nodes with an associated device.
  * Stereo Depth nodes now apply the default profile preset when initialized.

* **Bug Fixes**
  * Preset changes now reliably reset depth configuration before applying new settings.
  * Improved handling of post-processing hardware resources for RVC2 devices.

### DIFF
--- a/src/pipeline/node/StereoDepth.cpp
+++ b/src/pipeline/node/StereoDepth.cpp
@@ -41,6 +41,10 @@ StereoDepth::StereoDepth(std::unique_ptr<Properties> props)
     : DeviceNodeCRTP<DeviceNode, StereoDepth, StereoDepthProperties>(std::move(props)),
       initialConfig(std::make_shared<decltype(properties.initialConfig)>(properties.initialConfig)) {}
 
+StereoDepth::StereoDepth(const std::shared_ptr<Device>& device) : DeviceNodeCRTP<DeviceNode, StereoDepth, StereoDepthProperties>(device) {
+    setDefaultProfilePreset(PresetMode::DEFAULT);
+}
+
 StereoDepth::Properties& StereoDepth::getProperties() {
     properties.initialConfig = *initialConfig;
     return properties;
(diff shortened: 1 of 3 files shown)


src/pipeline/node/StereoDepth.cpp
```diff
@@ -41,6 +41,10 @@ StereoDepth::StereoDepth(std::unique_ptr<Properties> props)
     : DeviceNodeCRTP<DeviceNode, StereoDepth, StereoDepthProperties>(std::move(props)),
       initialConfig(std::make_shared<decltype(properties.initialConfig)>(properties.initialConfig)) {}
 
+StereoDepth::StereoDepth(const std::shared_ptr<Device>& device) : DeviceNodeCRTP<DeviceNode, StereoDepth, StereoDepthProperties>(device) {
+    setDefaultProfilePreset(PresetMode::DEFAULT);
```
aljazkonec1, line 45: This makes "constructor only stereo" use DEFAULT presets instead of making the DEFAULT presets equal to the constructor version of the presets. This will completely change behavior in all stereo examples. It also sets DEFAULT  preset for RVC2. Is that the same as not setting any preset?

src/pipeline/node/StereoDepth.cpp
```diff
@@ -206,6 +210,9 @@ void StereoDepth::setDefaultProfilePreset(PresetMode mode) {
 }
 
 void StereoDepth::setRvc2ProfilePreset(PresetMode mode) {
+    *initialConfig = StereoDepthConfig{};
+    setPostProcessingHardwareResources(Properties::AUTO, Properties::AUTO);
```
aljazkonec1, line 214: Why is this needed? Is this PR not ment to adjust only RVC4?

aljazkonec1 (changes requested):
using DEFAULT preset will completely change the behavior of stereo depth when the node is initialized without the build() function. To me the presets that are default constructed in initialConfig look better than when using the DEFAULT preset. Are we sure we want to do that?

src/pipeline/node/StereoDepth.cpp
JakubFara: The config does not need to set all the parameters so we "reset" the parameters at first

src/pipeline/node/StereoDepth.cpp
JakubFara: This unifies setting DEFAULT as default everywhere, seems intended to me.

JakubFara: cc: @CasualMathEnjoyer

PR #1961, opened 2026-08-25
Bugfix: avoid querying device after depthNode pipeline stops

MaticTonin:
Summary
Fixes a segfault in the Depth AUTO backend on-device tests on RVC2/network devices.
The test previously queried getConnectedCameraFeatures() after startPipelineAndRequireFirstFrames() returned. That helper stops the pipeline before returning, which closes the device RPC client. The subsequent capability query dereferenced the closed client and caused a SIGSEGV.
Capture the platform and ToF capability while the device is live, then use those cached values for post-stop backend assertions.

Tested with:
```
cmake --build build --target depth_node_test -j 4
DEPTHAI_PLATFORM=rvc2 ./build/tests/depth_node_test \
  'Depth: AUTO selects backend by platform and sensors' --reporter compact
```

https://github.com/luxonis/depthai-core/actions/runs/32837049490/job/97770930315#step:3:4404

## Summary by CodeRabbit

* **Tests**
  * Improved validation of automatic depth-camera backend selection across standard, user-camera, and camera-reuse scenarios.
  * Updated checks to account for the active platform and time-of-flight sensor availability.
  * Corrected depth-frame size validation for non-RVC4 platforms to reflect the full user-camera resolution.
  * Renamed a test case for clearer coverage and improved consistency across supported camera configurations.


aljazkonec1 (approved):

tests/src/ondevice_tests/pipeline/node/depth_node_test.cpp
```diff
@@ -377,6 +376,8 @@ void runUserCameraDepthTest(Pipeline& pipeline,
                             bool checkFrameSizes,
                             bool userPreviewStream) {
     PipelineStopGuard guard(pipeline);
+    const auto platform = device->getPlatform();
+    const auto autoUsesTof = platform == Platform::RVC2 && deviceReportsTofSensor(device);
```
JakubFara: Why we require RVC2 platform? You check the platform later.
I suggest rename it to `hasTofSensor` and just place there `deviceReportsTofSensor(device)`.

MaticTonin: Renamed and remived the RVC2 platform check yea, since its duplicated

JakubFara (approved):

PR #1962, opened 2026-08-26
report junit summaries to influxdb

ViacheslavTelelyukhin:
## Purpose
This branch will add logic to track test runs through influxdb

## Specification
added a script after junit generation that uploads them to influxdb. For now only to test_child.yml

## Deployment Plan
Do not merge yet, this is so I can run workflows on github test runner to check if everything work as expected

## Summary by CodeRabbit

* **New Features**
  * Added passed-test results, durations, and suite metadata to generated test reports.
  * Added automated publishing of individual test results and suite summaries to InfluxDB.
  * Included platform, protocol, and label details in test configuration reporting.

* **Bug Fixes**
  * Improved tracking of test status changes, including fixed and broken failures.
  * Added handling for oversized test logs and incomplete report data.
  * Improved reliability of automated test-result reporting.

### DIFF
--- a/.github/workflows/hil_testing_pull_request.yml
+++ b/.github/workflows/hil_testing_pull_request.yml
@@ -65,8 +65,7 @@ jobs:
           exit 1
         fi
   notify_slack:
-    needs:
-      - run_pr_tests
+    needs: [run_pr_tests, required_hil_test_report]
     if: >
       always() &&
       (
--- a/.github/workflows/hil_testing_pull_request.yml
+++ b/.github/workflows/hil_testing_pull_request.yml
@@ -20,14 +20,15 @@ jobs:
   run_pr_tests:
     name: Run Linux PR HIL tests
     if: >
-      (
+      (github.event_name == 'workflow_dispatch') ||
+      ((
         github.event.review.state == 'approved' ||
         (github.event_name == 'pull_request' && github.event.label.name == 'testable')
       ) &&
       (
         github.event.pull_request.base.ref == 'main' ||
         github.event.pull_request.base.ref == 'develop'
-      )
+      ))
     uses: ./.github/workflows/test_child.yml
     with:
       flavor: "vanilla"
--- a/.github/workflows/hil_testing_pull_request.yml
+++ b/.github/workflows/hil_testing_pull_request.yml
@@ -2,6 +2,7 @@ name: DepthAI Core HIL Testing PR
 
 # Trigger on submitted reviews or when manually marked as testable
 on:
+  workflow_dispatch:
   pull_request_review:
     types: [submitted]
   pull_request:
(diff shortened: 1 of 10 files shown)


ViacheslavTelelyukhin:
It should be acceptable at this point, only cosmetic and QOL changes remain.

.github/workflows/hil_testing_pull_request.yml
```diff
@@ -58,3 +64,41 @@ jobs:
           echo "run_pr_tests: $RESULT"
           exit 1
         fi
+  notify_slack:
```
aljazdu, line 67: I would avoid notifying slack for PR tests. Besides the "spamming", the PR tests would ideally be stable and would fail only in case of regressions introduced in the PR, and would as such be caught by the author.

.github/workflows/test_child.yml
```diff
@@ -68,7 +76,7 @@ jobs:
         fi
         BRANCH_NAME="${{ github.ref_name }}"
         PULL_REQUEST="false"
-        if [[ -n "${{ github.event.pull_request.number }}" ]]; then
+        if [[ "${{ github.workflow }}" = "DepthAI Core HIL Testing PR" ]]; then
```
aljazdu, line 79: Do we want to compare to a hard coded string value?

aljazdu (changes requested):
Slack notify step in PR tests is the only major thing I would change. Everything else looks ok.

.github/workflows/hil_testing_pull_request.yml
ViacheslavTelelyukhin: This is why we have an execution condition so it only notifies on develop/main/release runs after merge is approved. @aljazkonec1 explicitly wanted to have these. Do you think it's better if it only sends the notification if there are failing tests?

aljazdu: > @aljazkonec1 explicitly wanted to have these

In this case leave it as you've discussed.

> This is why we have an execution condition so it only notifies on develop/main/release runs after merge is approved.

This is not the case currently however.

The current conditions make `slack_notify` job trigger only on manual workflow dispatch.
The entire workflow is triggered on the following conditions:

- PR review submitted
- A PR is labeled
- manual trigger

then the `notify_slack` is triggered on the following:

- manual trigger
- when commit is pushed to main, develop, release


and the only time that both can happen is on workflow dispatch.

.github/workflows/test_child.yml
```diff
@@ -49,6 +49,14 @@ on:
         required: true
       CI_TELEMETRY_API_KEY:
         required: true
+      INFLUXDB_TEST_REPORT_BUCKET:
```
danilo-pejovic: Are we doing anything with these variables here? If not lets just remove them

.github/workflows/hil_testing_pull_request.yml
ViacheslavTelelyukhin: I corrected the execution condition. Now it's the same as for run_pr_tests except it doesn't notify for runs triggered by label.

aljazdu (approved):

PR #1970, opened 2026-09-03
Fix/python binding deadlock

ViacheslavTelelyukhin:
## Purpose
When I tried to send many messages form host to device to a script node input queue using python bindings the script would get stuck and XLInkOutHost wouldn't send more than 1 message to the node. It would just hand on the mainLoop() call indefinitly.
I believe this is because the queue on the device would fill up so inputQueue.send (host side) would wait untill it frees up but at the same time hold a lock on the queue so it couldn't get emptied.

## Specification
I changed the python binding for inputQueue.send and added a lock release.

This changes how consurrency is handled so should be considered carefully.

## Testing & Validation
I tested it on my example and it started working after the change.

## AI Usage
Codex told me to add this change

Submitted code was reviewed by a human: NO

The author is taking the responsibility for the contribution: NO

## Summary by CodeRabbit

* **Performance**
  * Improved Python threading responsiveness by allowing other threads to run while input data is being sent through the pipeline.


moratom (approved):
LGTM, interesting this wasn't identified for this long, thanks!

moratom:
@ViacheslavTelelyukhin maybe makes sense to ask an agent to go through all of the bindings and see if it can find more such cases where a long potentially blocking function doesn't release GIL even though it safely could. CC: @aljazkonec1

PR #1971, opened 2026-09-03
Fix/fusion confidence name

JakubFara:
Change confidence name in the `ToFStereoFusion`.
- previously it was `neuralConfidence` but the correct name is just `confidence`.

## Summary by CodeRabbit

- **Updates**
  - Renamed the `ToFStereoFusion` confidence output from `neuralConfidence` to `confidence` across supported interfaces.
  - The output type and configuration remain unchanged.
  - Applications using the previous property name must update references to `confidence`.
  - Updated the RVC4 device component version used by the SDK.


jakgra (approved):
LGTM

PR #1974, opened 2026-09-10
Add calibrated panorama stitching

pheec:
Adds a new calibrated panorama stitching mode through setUseInputCalibration(true).

When enabled, stitching:
- Uses camera intrinsics and rotations from each input frame.
- Skips feature matching.
- Treats camera centers as coincident and ignores translation.
- Reuses the prepared panorama geometry for subsequent frames.
- Aligns cylindrical panoramas around the mean Y-axis of all cameras instead of the first camera.
- Supports the existing seam finding, exposure compensation, and blending options.

## Summary by CodeRabbit

- **New Features**
  - Added an option to create panoramas using calibration data from input frames.
  - Added support for direct composition or full seam, exposure, and blending processing.
  - Improved cylindrical panorama alignment and reuse of calibration information.
- **Bug Fixes**
  - Improved handling of warped image dimensions and masks during panorama composition.
  - Added validation to reject incompatible camera geometry, distortion, frame sizes, and synchronization.
- **Tests**
  - Expanded coverage for calibration-based stitching, serialization, blending, overlap handling, and invalid configurations.

### DIFF
--- a/bindings/python/src/beta/node/StitchingBindings.cpp
+++ b/bindings/python/src/beta/node/StitchingBindings.cpp
@@ -64,6 +64,7 @@ void bind_beta_stitching(pybind11::module& m, void* pCallstack) {
 
     stitchingProperties.def_readwrite("mode", &StitchingProperties::mode)
         .def_readwrite("cameraModel", &StitchingProperties::cameraModel)
+        .def_readwrite("useInputCalibration", &StitchingProperties::useInputCalibration)
         .def_readwrite("continuous", &StitchingProperties::continuous)
         .def_readwrite("estimationFrames", &StitchingProperties::estimationFrames)
         .def_readwrite("maxPanoramaWidth", &StitchingProperties::maxPanoramaWidth)
@@ -116,6 +117,11 @@ void bind_beta_stitching(pybind11::module& m, void* pCallstack) {
         .def("getMinIncidenceAngle", &Stitching::getMinIncidenceAngle, DOC(dai, beta, node, Stitching, getMinIncidenceAngle))
         .def("setCameraModel", &Stitching::setCameraModel, py::arg("model"), DOC(dai, beta, node, Stitching, setCameraModel))
         .def("getCameraModel", &Stitching::getCameraModel, DOC(dai, beta, node, Stitching, getCameraModel))
+        .def("setUseInputCalibration",
+             &Stitching::setUseInputCalibration,
+             py::arg("useInputCalibration"),
+             DOC(dai, beta, node, Stitching, setUseInputCalibration))
+        .def("getUseInputCalibration", &Stitching::getUseInputCalibration, DOC(dai, beta, node, Stitching, getUseInputCalibration))
         .def("setContinuous", &Stitching::setContinuous, py::arg("continuous"), DOC(dai, beta, node, Stitching, setContinuous))
         .def("getContinuous", &Stitching::getContinuous, DOC(dai, beta, node, Stitching, getContinuous))
         .def("setEstimationFrames", &Stitching::setEstimationFrames, py::arg("frames"), DOC(dai, beta, node, Stitching, setEstimationFrames))
--- a/bindings/python/tests/beta_stitching_test.py
+++ b/bindings/python/tests/beta_stitching_test.py
@@ -11,3 +11,7 @@ def test_stitching_is_exposed_only_through_beta_api():
         assert isinstance(stitching, dai.beta.node.Stitching)
         assert stitching.getNumInputs() == 2
         assert stitching.getMode() == dai.beta.node.Stitching.Mode.PANORAMA
+        assert not stitching.getUseInputCalibration()
+
+        stitching.setUseInputCalibration(True)
+        assert stitching.getUseInputCalibration()
--- a/bindings/python/tests/beta_stitching_test.py
+++ b/bindings/python/tests/beta_stitching_test.py
@@ -1,7 +1,7 @@
 import depthai as dai
 
 
-def test_stitching_is_exposed_only_through_beta_api():
+def test_stitching_beta_api_and_input_calibration_configuration():
     assert not hasattr(dai.node, "Stitching")
     assert dai.beta.node.Stitching.Properties is dai.beta.StitchingProperties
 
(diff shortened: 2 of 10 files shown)


bindings/python/tests/beta_stitching_test.py
```diff
@@ -11,3 +11,7 @@ def test_stitching_is_exposed_only_through_beta_api():
         assert isinstance(stitching, dai.beta.node.Stitching)
         assert stitching.getNumInputs() == 2
         assert stitching.getMode() == dai.beta.node.Stitching.Mode.PANORAMA
+        assert not stitching.getUseInputCalibration()
```
MaticTonin, line 14: Does then this test require device? Pls check since otherwise it will fail on GH Actions

bindings/python/src/beta/node/StitchingBindings.cpp
```diff
@@ -116,6 +117,11 @@ void bind_beta_stitching(pybind11::module& m, void* pCallstack) {
         .def("getMinIncidenceAngle", &Stitching::getMinIncidenceAngle, DOC(dai, beta, node, Stitching, getMinIncidenceAngle))
         .def("setCameraModel", &Stitching::setCameraModel, py::arg("model"), DOC(dai, beta, node, Stitching, setCameraModel))
         .def("getCameraModel", &Stitching::getCameraModel, DOC(dai, beta, node, Stitching, getCameraModel))
+        .def("setUseInputCalibration",
+             &Stitching::setUseInputCalibration,
+             py::arg("useInputCalibration"),
+             DOC(dai, beta, node, Stitching, setUseInputCalibration))
+        .def("getUseInputCalibration", &Stitching::getUseInputCalibration, DOC(dai, beta, node, Stitching, getUseInputCalibration))
```
MaticTonin, line 124: Different name:

I would go for getInputCalibration and setInputCalibration, since else setUse ... its not nice wording.

include/depthai/beta/node/Stitching.hpp
```diff
@@ -182,6 +182,24 @@ class Stitching : public DeviceNodeCRTP<BetaNode, Stitching, StitchingProperties
...
+     * compensation, and multiband blending. The prepared composition is reused for subsequent groups. Only undistorted
+     * inputs are accepted. Only used in `Mode::PANORAMA`.
+     */
+    void setUseInputCalibration(bool useInputCalibration);
+
+    /**
+     * Return whether panorama composition uses the calibration carried by the input ImgTransformations.
+     */
+    bool getUseInputCalibration() const;
+
```
MaticTonin, line 202: Do you even require a getter? or just for test reasons? If for test reasons, its fine, renaming needed tho.

src/beta/utilities/Stitching/Stitching.cpp
```diff
@@ -45,6 +46,50 @@ std::string statusToString(cv::Stitcher::Status status) {
     return "unknown error";
 }
 
+bool isUndistorted(const ImgTransformation& transformation) {
+    const auto coefficients = transformation.getDistortionCoefficients();
+    return std::all_of(coefficients.begin(), coefficients.end(), [](float coefficient) {
+        return std::isfinite(coefficient) && std::abs(coefficient) <= matrix::MATRIX_EQ_EPSILON;
+    });
+}
```
MaticTonin, line 54: Probably best if this function exsists for ImgTransformations rather than here.

src/beta/utilities/Stitching/Stitching.cpp
```diff
@@ -178,18 +225,106 @@ class Stitching::Impl {
...
+            matrix::validateRotationMatrix3x3(extrinsics.rotationMatrix);
+
+            const auto intrinsics = transformation.getIntrinsicMatrix();
+            const float fx = intrinsics[0][0];
+            const float fy = intrinsics[1][1];
+            const float ppx = intrinsics[0][2];
+            const float ppy = intrinsics[1][2];
+            DAI_CHECK_V(std::isfinite(fx) && std::isfinite(fy) && std::isfinite(ppx) && std::isfinite(ppy) && fx > 0.0f && fy > 0.0f,
+                        "Calibrated panorama input {} carries invalid intrinsics",
+                        i);
```
MaticTonin, line 270: As far as I know, in transformations there exsists check similar to that? Could be that I am wrong

src/beta/utilities/Stitching/Stitching.cpp
```diff
@@ -327,6 +472,58 @@ void Stitching::run() {
             continue;
         }
 
+        if(currentProperties.useInputCalibration) {
+            cv::Mat pano;
+            try {
+                if(!impl->fixedPanorama.isPrepared()) {
+                    auto cameras = impl->camerasFromInputCalibration(transformations);
+                    if(currentProperties.cameraModel == CameraModel::CYLINDRICAL) {
```
MaticTonin, line 480: I would mybe name this as projectionModel, since it represents more on how images will be projected, than what device was calibrated in.

MaticTonin (changes requested):
Overall looks fine, some nitpicks

bindings/python/tests/beta_stitching_test.py
pheec: It does not

bindings/python/src/beta/node/StitchingBindings.cpp
pheec: I like it as it is

include/depthai/beta/node/Stitching.hpp
pheec: Test renamed

src/beta/utilities/Stitching/Stitching.cpp
pheec: There is matrix::isValidIntrinsicsMatrix(), but it validates calibration matrices. ImgTransformation::isValid() only checks image dimensions, and transformations may carry manually supplied or deserialized intrinsics. This local check validates exactly the values passed into OpenCV’s CameraParams, including guarding the fy / fx division, so I believe it should remain.

src/beta/utilities/Stitching/Stitching.cpp
pheec: I’d prefer to keep the ImgTransformation API as minimal as possible

src/beta/utilities/Stitching/Stitching.cpp
pheec: cameraModel is not a model in which it was calibrated. It describes how the images are warped, such as spherical, cylindrical, or pinhole. The existing Stitching API also uses this name in setCameraModel() and getCameraModel(), so I would prefer not to rename it in this PR.

MaticTonin (approved):
LGTM now

PR #1981, opened 2026-09-14
feat: add depthai::device-manager target

lukas-plojhar:
## Summary

Adds an opt-in library target, **`depthai::device-manager`** (`-DDEPTHAI_BUILD_DEVICE_MANAGER=ON`), for tools that only discover, flash and configure devices and never build a pipeline. First consumer: the oakctl RVC2 helper.

- Compiles 23 existing sources: `DeviceBootloader`, XLink connection/stream, `Resources` (embedded bootloader firmware), logging/environment/platform helpers, `Version`, `OpenVINO`/`BlobReader`, bspatch and the bootloader-shared protocol.
- Links XLink, spdlog/fmt, libarchive/lzma, bzip2, semver and the CMakeRC resource library. Exported and installed under the existing `depthai::` namespace like `depthai::opencv`.
- `depthai::core` is unchanged: it builds the same files with the same flags.

## Why the pipeline used to come along

Two reference chains dragged the whole pipeline into any consumer of `DeviceBootloader`:

1. The `Pipeline`-taking overloads in `DeviceBootloader.cpp` (`createDepthaiApplicationPackage`, `saveDepthaiApplicationPackage`, `flash(const Pipeline&)`) call `PipelineImpl`, which pulls every node and datatype.
2. `initialize()` calls `Telemetry::init()`, and `Telemetry.cpp` references `DeviceBase`.

Both are now excluded under `DEPTHAI_DEVICE_MANAGER_ONLY`, a public definition of the new target. `DeviceBootloader.hpp` forward-declares `Pipeline` in that mode and keeps including `Pipeline.hpp` otherwise, so existing consumers (including the Python bindings) see no change.

Also fixes two includes that were only satisfied transitively via `Pipeline.hpp`: `DeviceBootloader.hpp` uses `dai::Version`, and `Resources.cpp:184` uses `std::unordered_set`.

## Measured on the oakctl helper (macOS arm64, Release)

| | `depthai::core` | `depthai::device-manager` |
|---|---|---|
| helper binary | 7,146,192 B | 3,061,984 B |
| depthai-core objects compiled | 162 | 23 |
| serial compile time of those objects | 392.8 s | ~30 s |

Helper test suite and CLI smoke checks pass on the slim build; device discovery output is identical.

## Test plan

- [x] `DEPTHAI_BUILD_DEVICE_MANAGER=ON` builds and links a consumer on macOS arm64
- [x] Syntax-only compile of all 162 core TUs with the header change (only `Resources.cpp` needed the include fix)
- [ ] Linux and Windows CI (exercised by luxonis/oak-agent once pinned to this commit)
- [ ] Default configuration (`DEPTHAI_BUILD_DEVICE_MANAGER=OFF`) unaffected in depthai-core CI

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_014Y5Yop1fcfPL6Ucvuy7Qvx

## Summary by CodeRabbit

- **New Features**
  - Added an optional device-management library for bootloader operations and device connectivity without the full pipeline.
  - Added CMake configuration and presets for building and exporting the library.
  - Added optional embedded bootloader resources for device-manager builds.

- **Compatibility**
  - Existing builds retain their current behavior when device-manager support is disabled.
  - Device-manager builds exclude pipeline-based package creation, saving, and flashing APIs.

- **Build Validation**
  - Added automated validation for the device-manager configuration.
  - Improved handling of runtime dependencies during builds.


CMakeLists.txt
```diff
@@ -1056,6 +1056,121 @@ if(DEPTHAI_DYNAMIC_CALIBRATION_SUPPORT)
...
+    if(DEPTHAI_VCPKG_INTERNAL_ONLY)
+        exclude_archive_libs_symbols(${TARGET_DEVICE_MANAGER_NAME})
+    endif()
+    set_property(TARGET ${TARGET_DEVICE_MANAGER_NAME} PROPERTY CXX_STANDARD 17)
+    set_property(TARGET ${TARGET_DEVICE_MANAGER_NAME} PROPERTY CXX_STANDARD_REQUIRED ON)
+    set_property(TARGET ${TARGET_DEVICE_MANAGER_NAME} PROPERTY CXX_EXTENSIONS OFF)
+    target_compile_features(${TARGET_DEVICE_MANAGER_NAME} INTERFACE cxx_std_17)
+
+    if(DEPTHAI_ENABLE_LIBUSB AND NOT XLINK_LIBUSB_SYSTEM)
+        add_runtime_dependencies(${TARGET_DEVICE_MANAGER_NAME} usb-1.0)
```
lukas-plojhar, line 1165: **[1/10] add_runtime_dependencies is a macro that overwrites (set) the directory-scope required_dll_files on every call; the new device-manager call runs last, so with DEPTHAI_BUILD_DEVICE_MANAGER=ON the install step ships only the libusb runtime and drops dynamic_calibration.**

Windows, default options (DEPTHAI_DYNAMIC_CALIBRATION_SUPPORT=ON, DEPTHAI_ENABLE_LIBUSB=ON, vcpkg libusb so XLINK_LIBUSB_SYSTEM unset) plus -DDEPTHAI_BUILD_DEVICE_MANAGER=ON -DDEPTHAI_INSTALL=ON: the macro body at CMakeLists.txt:1025-1049 does `set(required_dll_files ${dlls})`, so after line 1056 sets it to dynamic_calibration and line 1165 resets it to usb-1.0, `install(FILES ${required_dll_files})` at ~1541 copies usb-1.0.dll but no dynamic_calibration.dll; an installed depthai::core consumer then fails at load with a missing dynamic-calibration DLL (pre-PR the same overwrite dropped usb-1.0.dll instead, so the PR changes which runtime is lost rather than fixing it). Fix: `list(APPEND required_dll_files ...)` in the macro, or move the device-manager runtime-dependency call before the dynamic-calibration one.

src/device/DeviceBootloader.cpp
```diff
@@ -8,13 +8,17 @@
...
 #include "depthai-bootloader-shared/Structure.hpp"
 #include "depthai-bootloader-shared/XLinkConstants.hpp"
+#ifndef DEPTHAI_DEVICE_MANAGER_ONLY
 #include "depthai/pipeline/Assets.hpp"
+#endif  // DEPTHAI_DEVICE_MANAGER_ONLY
 #include "depthai/utility/Serialization.hpp"
 #include "depthai/xlink/XLinkConstants.hpp"
 
 // project
+#ifndef DEPTHAI_DEVICE_MANAGER_ONLY
```
lukas-plojhar, line 18: _(anchored here; the line in question is `src/device/DeviceBootloader.cpp:31`, outside the diff hunk)_

**[2/10] `#include "zlib.h"` (line 31) stays unconditional although its only users (compressBound/compress2 inside createDepthaiApplicationPackage) are now guarded out, and the device-manager target neither links ZLIB::ZLIB nor guards the include.**

The DM compile of DeviceBootloader.cpp has no ZLIB-derived include flag (verified in the consumer's compile_commands.json); zlib.h is found only because fmt/spdlog/libarchive share one vcpkg include prefix (and static vcpkg's libarchive wrapper appends ZLIB::ZLIB). Configure with -DDEPTHAI_BOOTSTRAP_VCPKG=OFF -DDEPTHAI_BUILD_DEVICE_MANAGER=ON with zlib in its own prefix (ZLIB_ROOT on Windows, a cross-compile sysroot with zlib staged separately): find_package(ZLIB) succeeds, depthai-core compiles, but depthai-device-manager's DeviceBootloader.cpp fails with `zlib.h: No such file or directory`. Fix: move the include inside the existing `#ifndef DEPTHAI_DEVICE_MANAGER_ONLY` block (or link ZLIB::ZLIB PRIVATE like core does at line 920).

CMakeLists.txt
```diff
@@ -1056,6 +1056,121 @@ if(DEPTHAI_DYNAMIC_CALIBRATION_SUPPORT)
...
+########################
+# Device manager target
+########################
+# depthai::device-manager: DeviceBootloader + XLinkConnection + embedded bootloader firmware, without the
+# pipeline, nodes or datatypes. Used by device management tools (e.g. oakctl) that only flash and configure.
+if(DEPTHAI_BUILD_DEVICE_MANAGER)
+    set(TARGET_DEVICE_MANAGER_NAME ${PROJECT_NAME}-device-manager)
+    set(TARGET_DEVICE_MANAGER_ALIAS device-manager)
+
+    add_library(${TARGET_DEVICE_MANAGER_NAME}
```
lukas-plojhar, line 1068: **[3/10] depthai-device-manager recompiles a strict subset of core's TUs with a different macro set into a second archive exporting identical dai:: symbols, and nothing (CMake property, #error, docs) stops a consumer from linking both depthai::core and depthai::device-manager.**

nm shows both archives define dai::DeviceBootloader::init, dai::Resources::getInstance, all dai::initialize overloads, etc.; 22 of DM's 23 members are name-for-name duplicates of core members. A real link test (TU using DeviceBootloader(DeviceInfo) + flash(Pipeline), linked `libdepthai-device-manager.a libdepthai-core.a`) produces a mixed binary using core's DeviceBootloader.cpp.o but DM's Resources.cpp.o and Initialization.cpp.o, so a tool that links DM directly and picks up core transitively (e.g. via depthai::opencv or another library) gets `Device` on RVC2 throwing "DepthAI compiled without support for MyriadX Device FW" and Telemetry::init silently skipped, with no diagnostic; with GNU ld / whole-archive semantics it becomes duplicate-definition link errors. depthai::opencv/depthai::all set the precedent of composing over core rather than recompiling. At minimum add a guard such as COMPATIBLE_INTERFACE_STRING with differing values on core and device-manager so CMake rejects a target that links both, and document that the two are mutually exclusive.

src/device/DeviceBootloader.cpp
```diff
@@ -8,13 +8,17 @@
 #include "depthai-bootloader-shared/SBR.h"
 #include "depthai-bootloader-shared/Structure.hpp"
 #include "depthai-bootloader-shared/XLinkConstants.hpp"
+#ifndef DEPTHAI_DEVICE_MANAGER_ONLY
 #include "depthai/pipeline/Assets.hpp"
```
lukas-plojhar, line 12: **[4/10] Five newly added #include lines nested inside `#ifndef DEPTHAI_DEVICE_MANAGER_ONLY` blocks are not indented as .clang-format's `IndentPPDirectives: BeforeHash` requires, so the CI style job will fail on this PR.**

`clang-format --dry-run --Werror --style=file` at the PR head fails on DeviceBootloader.cpp:12,19,20, include/depthai/device/DeviceBootloader.hpp:14 and src/utility/Initialization.cpp:15 (the only reformat is adding four-space indentation, e.g. `    #include "depthai/pipeline/Assets.hpp"`), while the base-commit versions of all three files pass clean and existing nested includes in the repo are indented (Resources.cpp:42). The `style` job in .github/workflows/main.workflow.yml runs ci/check_format.sh, which builds the clangformat target and then `git diff-index --quiet HEAD`, so the reformat produces a diff and the job fails.

```suggestion
    #include "depthai/pipeline/Assets.hpp"
```
_(same indentation fix applies to DeviceBootloader.cpp:19-20, DeviceBootloader.hpp:14 and Initialization.cpp:15)_

CMakeLists.txt
```diff
@@ -1056,6 +1056,121 @@ if(DEPTHAI_DYNAMIC_CALIBRATION_SUPPORT)
...
+            DEPTHAI_DEVICE_RVC4_VERSION="${DEPTHAI_DEVICE_RVC4_VERSION}"
+    )
+    if(DEPTHAI_ENABLE_LIBUSB)
+        target_compile_definitions(${TARGET_DEVICE_MANAGER_NAME} PRIVATE DEPTHAI_ENABLE_LIBUSB DEPTHAI_HAVE_LIBUSB_SUPPORT)
+    endif()
+    if(DEPTHAI_ENABLE_DEVICE_BOOTLOADER_FW)
+        target_compile_definitions(${TARGET_DEVICE_MANAGER_NAME} PRIVATE DEPTHAI_ENABLE_DEVICE_BOOTLOADER_FW)
+    endif()
+    if(DEPTHAI_BINARIES_RESOURCE_COMPILE)
+        target_link_libraries(${TARGET_DEVICE_MANAGER_NAME} PRIVATE "${DEPTHAI_RESOURCE_LIBRARY_NAME}")
```
lukas-plojhar, line 1113: **[5/10] depthai::device-manager links the single global depthai-resources cmrc library, so under default options it embeds ~14.9 MB of RVC2 (8.9 MB) and RVC4 (6.0 MB) firmware it can never read; the "~3 MB slim helper" in the PR description is only true because the consumer script passes -DDEPTHAI_ENABLE_DEVICE_FW=OFF -DDEPTHAI_ENABLE_DEVICE_RVC4_FW=OFF.**

RESOURCE_COMPILED_FILES gets the device fwp (line ~783, DEPTHAI_ENABLE_DEVICE_FW default ON) and RVC4 fwp (line ~831, default ON); cmrc's generated lib.cpp references every file's _begin/_end symbols from root_index, and Resources::Resources() always calls cmrc::depthai::get_filesystem() for the bootloader tar, so every blob is pulled into any DM consumer. The DM target never defines DEPTHAI_ENABLE_DEVICE_FW/RVC3/RVC4, so Resources::getDeviceFirmware()/getDeviceRVC4Fwp() still throw "compiled without support" while ~17 MB of blobs sit in the binary instead of the 1.15 MB bootloader tar. Cheaper: a second `cmrc_add_resource_library(depthai-bootloader-resources NAMESPACE depthai ...)` built from only the bootloader resource list, linked by device-manager (cmrc's index is per-library and the bootloader path only opens CMRC_DEPTHAI_BOOTLOADER_TAR_XZ).

include/depthai/device/DeviceBootloader.hpp
```diff
@@ -9,7 +9,10 @@
 // project
 #include "CallbackHandler.hpp"
 #include "depthai/common/UsbSpeed.hpp"
+#include "depthai/device/Version.hpp"
+#ifndef DEPTHAI_DEVICE_MANAGER_ONLY
```
lukas-plojhar, line 13: **[6/10] The `#ifndef DEPTHAI_DEVICE_MANAGER_ONLY` include plus `#ifdef` forward-declaration pair makes an installed public header change shape based on a build macro, when an unconditional `class Pipeline;` (the pattern DeviceBase.hpp:51 and Node.hpp:30 already use) would do the same job with no macro in the header.**

All seven uses of Pipeline in this header (lines 140-257) are `const Pipeline&` parameters on out-of-line declarations; the only inline bodies are the operator<< overloads at 552-606, so nothing needs the complete type (proven by the DM build compiling this header without Pipeline.hpp). Of the six includers, depthai.hpp, Device.cpp and DeviceBootloader.cpp already include Pipeline.hpp themselves and the two src/utility headers never mention Pipeline; only bindings/python/src/DeviceBootloaderBindings.cpp (py::overload_cast<const Pipeline&,...>) needs one explicit `#include "depthai/pipeline/Pipeline.hpp"`. As shipped, a consumer TU that gets DEPTHAI_DEVICE_MANAGER_ONLY (or fails to) sees a different transitive include set than the library it links was built with, and every future header edit must keep both branches compiling. Simpler form: delete lines 13-15 and 27-29, add one unconditional `class Pipeline;`, keep the .cpp guards.

include/depthai/device/DeviceBootloader.hpp
```diff
@@ -9,7 +9,10 @@
 // project
 #include "CallbackHandler.hpp"
```
lukas-plojhar, line 10: **[7/10] `#include "CallbackHandler.hpp"` is unused by DeviceBootloader.hpp/.cpp and is the sole path by which the device-manager public header reaches ADatatype.hpp/Serialization.hpp and 42 libnop headers, which is the only reason `libnop` has to be a PUBLIC dependency of depthai::device-manager (CMakeLists.txt:1141).**

Compiling with the DM target's real flags and -H, the full header pulls 42 nop/ headers and 2 pipeline/datatype headers; a TU including every sibling header except CallbackHandler.hpp pulls 0 nop/ headers (the DM DeviceBootloader.cpp.o.d shows the same 42, XLinkConnection/XLinkStream/Version depfiles show 0). CallbackHandler is referenced only by its own header and by src/device/CallbackHandler.cpp, which is commented out of TARGET_CORE_SOURCES (line ~280). Every consumer TU of a library whose stated purpose is "without the pipeline, nodes or datatypes" therefore recompiles libnop templates and ADatatype/VectorMemory declarations for types the DM archive exports no symbols for, and must have libnop resolvable at find_package/export time. Fix: drop the include (verified with -fsyntax-only that std::function/std::atomic/std::shared_ptr still resolve via the remaining headers) and move libnop to PRIVATE on the DM target (Serialization.cpp is still compiled, so it cannot be dropped entirely; nlohmann_json must stay PUBLIC via bootloader-shared Config.hpp).

include/depthai/device/DeviceBootloader.hpp
```diff
@@ -21,6 +24,10 @@
 
 namespace dai {
 
+#ifdef DEPTHAI_DEVICE_MANAGER_ONLY
+class Pipeline;
```
lukas-plojhar, line 28: _(anchored here; the line in question is `include/depthai/device/DeviceBootloader.hpp:139`, outside the diff hunk)_

**[8/10] Under DEPTHAI_DEVICE_MANAGER_ONLY the header still declares the six Pipeline-taking members (createDepthaiApplicationPackage x2 at 139/149, saveDepthaiApplicationPackage x2 at 162/176, flash x2 at 243/256) whose definitions DeviceBootloader.cpp guards out, so the shipped header advertises API the device-manager archive does not contain.**

`nm -C libdepthai-device-manager.a` has no createDepthaiApplicationPackage / saveDepthaiApplicationPackage / flash(...Pipeline...) symbols, and install(DIRECTORY include/depthai) at ~1525 unconditionally ships Pipeline.hpp alongside. A consumer linking depthai::device-manager that includes <depthai/pipeline/Pipeline.hpp> and calls `bl.flash(cb, pipeline)` compiles cleanly and fails only at link time with an undefined reference, instead of a compile-time error naming the missing member. Since DEPTHAI_DEVICE_MANAGER_ONLY is PUBLIC, wrapping the two declaration blocks (~139-177 and ~239-257) in the same `#ifndef` turns this into a compile-time diagnostic. (Python bindings bind these members but link depthai::core only, so they are unaffected.)

CMakeLists.txt
```diff
@@ -1056,6 +1056,121 @@ if(DEPTHAI_DYNAMIC_CALIBRATION_SUPPORT)
...
+# Device manager target
+########################
+# depthai::device-manager: DeviceBootloader + XLinkConnection + embedded bootloader firmware, without the
+# pipeline, nodes or datatypes. Used by device management tools (e.g. oakctl) that only flash and configure.
+if(DEPTHAI_BUILD_DEVICE_MANAGER)
+    set(TARGET_DEVICE_MANAGER_NAME ${PROJECT_NAME}-device-manager)
+    set(TARGET_DEVICE_MANAGER_ALIAS device-manager)
+
+    add_library(${TARGET_DEVICE_MANAGER_NAME}
+        "${DEPTHAI_BOOTLOADER_SHARED_SOURCES}"
```
lukas-plojhar, line 1069: **[9/10] The 23-file device-manager source list is a hand-copied subset of TARGET_CORE_SOURCES with no CI coverage, and enabling the option neither shares objects with nor excludes the 162-TU core archive from `all`, so it is both redundant-compile-heavy and prone to silent drift.**

In the consumer build tree compile_commands.json has 162 depthai-core.dir entries plus 23 depthai-device-manager.dir entries, all 162 core objects and a 16 MB libdepthai-core.a exist, yet oakctl-rvc2-helper's link.txt links only libdepthai-device-manager.a; Makefile2 lists depthai-core.dir/all as the first dependency of `all`, and the consumer avoids paying for it only via `--target` in ci/scripts/build-rvc2-helper.sh:60. The only flag difference for e.g. XLinkConnection.cpp is -DDEPTHAI_DEVICE_MANAGER_ONLY vs -DMCAP_STATIC -DYAML_CPP_STATIC_DEFINE, and only 4 DM TUs even include DeviceBootloader.hpp where the macro matters, so 19 objects are compiled twice for identical output. No .github/workflows or ci/ file sets DEPTHAI_BUILD_DEVICE_MANAGER. A later commit that adds src/utility/NewHelper.cpp (used by Resources.cpp/XLinkConnection.cpp) only to TARGET_CORE_SOURCES keeps every core CI job green and breaks the oakctl helper link the next time it bumps the pin. Cheaper: define DEPTHAI_DEVICE_MANAGER_SOURCES once and `list(APPEND TARGET_CORE_SOURCES ${DEPTHAI_DEVICE_MANAGER_SOURCES})` (single source of truth), add one `-DDEPTHAI_BUILD_DEVICE_MANAGER=ON` configure+build step to main.workflow.yml, and consider an OBJECT library for the 19 identical TUs.

CMakeLists.txt
```diff
@@ -1056,6 +1056,121 @@ if(DEPTHAI_DYNAMIC_CALIBRATION_SUPPORT)
...
+        src/utility/MemoryWrappers.cpp
+        src/bspatch/bspatch.c
+    )
+    add_library("${PROJECT_NAME}::${TARGET_DEVICE_MANAGER_ALIAS}" ALIAS ${TARGET_DEVICE_MANAGER_NAME})
+    set_target_properties(${TARGET_DEVICE_MANAGER_NAME} PROPERTIES EXPORT_NAME ${TARGET_DEVICE_MANAGER_ALIAS})
+    list(APPEND targets_to_export ${TARGET_DEVICE_MANAGER_NAME})
+
+    target_compile_definitions(${TARGET_DEVICE_MANAGER_NAME}
+        PUBLIC
+            DEPTHAI_TARGET_CORE
```
lukas-plojhar, line 1098: **[10/10] DEPTHAI_TARGET_CORE is exported PUBLIC from a target that is not core; its only reader (include/depthai/config/config.hpp:4, which #undefs DEPTHAI_HAVE_OPENCV_SUPPORT when DEPTHAI_TARGET_OPENCV is absent) is a no-op for this target, and zoo_helper (line 621), the other non-core target compiling the same sources, sets it PRIVATE.**

grep across include/, src/, bindings/, cmake/ and the shared submodules finds no other reader; no DM public header reaches config.hpp (the one DM TU that includes it, Initialization.cpp, reads only build::HAVE_LIBUSB_SUPPORT), so no consumer TU needs the define. Exporting it PUBLIC puts a misleading "you are linking core" entry in the target's INTERFACE_COMPILE_DEFINITIONS for every downstream (e.g. oakctl), and a downstream that separately compiles OpenCV-enabled depthai code with -DDEPTHAI_HAVE_OPENCV_SUPPORT but without DEPTHAI_TARGET_OPENCV gets it silently #undef'd by config.hpp. Make it PRIVATE (matching zoo_helper) or drop it; DEPTHAI_DEVICE_MANAGER_ONLY is the one define that genuinely must be PUBLIC.

CMakeLists.txt
lukas-plojhar: Fixed in c09a0060472afe22ff03f30cf233ec93e517cc68 (https://github.com/luxonis/depthai-core/pull/1981/commits/c09a0060472afe22ff03f30cf233ec93e517cc68). `add_runtime_dependencies` now appends to `required_dll_files` instead of replacing it.

src/device/DeviceBootloader.cpp
lukas-plojhar: Fixed in 6ef278a0a9b78c56e66749a0a965d197e4ba6fd3 (https://github.com/luxonis/depthai-core/pull/1981/commits/6ef278a0a9b78c56e66749a0a965d197e4ba6fd3). `zlib.h` is included only outside `DEPTHAI_DEVICE_MANAGER_ONLY`, alongside its only users.

CMakeLists.txt
lukas-plojhar: Fixed in e071e650e2f0688bc86f1b13ff7728bfbc8179dc (https://github.com/luxonis/depthai-core/pull/1981/commits/e071e650e2f0688bc86f1b13ff7728bfbc8179dc). Both targets carry an `INTERFACE_DEPTHAI_FLAVOR` value under `COMPATIBLE_INTERFACE_STRING`; a consumer linking `depthai::core` and `depthai::device-manager` together now fails at generate time ("The INTERFACE_DEPTHAI_FLAVOR property of depthai-device-manager does not agree with the value of DEPTHAI_FLAVOR already determined"). Verified locally.

src/device/DeviceBootloader.cpp
lukas-plojhar: Fixed in dfaf2c1d0eaccff226354f2aef2c4f231d78bfc5 (https://github.com/luxonis/depthai-core/pull/1981/commits/dfaf2c1d0eaccff226354f2aef2c4f231d78bfc5). Guarded includes indented; all four touched files pass `clang-format --dry-run --Werror --style=file`.

CMakeLists.txt
lukas-plojhar: Fixed in 0aa6c94e2a0c705bddf58251df2962dde7dc4edf (https://github.com/luxonis/depthai-core/pull/1981/commits/0aa6c94e2a0c705bddf58251df2962dde7dc4edf). `depthai::device-manager` now links a separate `depthai-bootloader-resources` CMakeRC library holding only `DEPTHAI_BOOTLOADER_RESOURCE_LIST`; it is exported alongside the existing resource library for static consumers.

include/depthai/device/DeviceBootloader.hpp
lukas-plojhar: Fixed in 81ea9af9f877d40f5e247d930e1c4de7c74064a5 (https://github.com/luxonis/depthai-core/pull/1981/commits/81ea9af9f877d40f5e247d930e1c4de7c74064a5). Handled through the fix for [8/10]: the Pipeline-based declarations are guarded out of the header, so the forward declaration is gone. The `Pipeline.hpp` include stays conditional so `depthai::core` consumers (including the Python bindings) see exactly the header they saw before.

include/depthai/device/DeviceBootloader.hpp
lukas-plojhar: Fixed in d6d0a6de8bbe57fa5cc6d6216ac204b7f23df652 (https://github.com/luxonis/depthai-core/pull/1981/commits/d6d0a6de8bbe57fa5cc6d6216ac204b7f23df652). `CallbackHandler.hpp` include removed; libnop is now a PRIVATE dependency of the target (nlohmann stays PUBLIC for the bootloader Config header). All 162 core TUs still compile against the header.

include/depthai/device/DeviceBootloader.hpp
lukas-plojhar: Fixed in 81ea9af9f877d40f5e247d930e1c4de7c74064a5 (https://github.com/luxonis/depthai-core/pull/1981/commits/81ea9af9f877d40f5e247d930e1c4de7c74064a5). `createDepthaiApplicationPackage`, `saveDepthaiApplicationPackage` and `flash(const Pipeline&)` are now declared only outside `DEPTHAI_DEVICE_MANAGER_ONLY`, so a misuse is a compile error.

CMakeLists.txt
lukas-plojhar: Fixed in e25aa69c6452d10087092da48500aedff14fce9d (https://github.com/luxonis/depthai-core/pull/1981/commits/e25aa69c6452d10087092da48500aedff14fce9d). 9a and 9b fixed: `DEPTHAI_DEVICE_MANAGER_SOURCES` is defined once and `TARGET_CORE_SOURCES` includes it (the set of objects compiled into `depthai-core` is unchanged, verified by diffing the generated object list), and 88562e9d2301c34a07b1e124e309634051d32998 adds a `device-manager` preset plus a CI job that builds the target on every push and PR. 9c (double compile) and 9d (default-build participation) are deferred as agreed.

CMakeLists.txt
lukas-plojhar: Fixed in 5dc2e1a85d0ef1d3e26e1263a8d941dd3cbd57ba (https://github.com/luxonis/depthai-core/pull/1981/commits/5dc2e1a85d0ef1d3e26e1263a8d941dd3cbd57ba). `DEPTHAI_TARGET_CORE` is PRIVATE on the target; only `DEPTHAI_DEVICE_MANAGER_ONLY` remains PUBLIC.

lukas-plojhar (commented):
All review threads are addressed and resolved as of 88562e9 (see the per-thread replies for the fixing commits). Submitting this review to trigger CI, since the workflow only runs on pull_request events for PRs labelled "testable" and this account cannot add labels.

.github/workflows/main.workflow.yml
```diff
@@ -133,6 +133,32 @@ jobs:
     - name: Build minimal target set
       run: cmake --build --preset minimal
 
+  device-manager-build:
+    runs-on: ubuntu-24.04
+    needs: [precheck, consistency]
+    if: needs.precheck.outputs.should_run == 'true'
+
+    steps:
+    - uses: actions/checkout@v4
```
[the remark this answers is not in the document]
lukas-plojhar: Fixed in 798e7d937 and 4c56b720f. The device-manager job now pins actions/checkout and jwlawson/actions-setup-cmake to the full commit SHAs resolved from their existing v4 and v1.13 tags, respectively. I verified both commits against the upstream repositories.

Also addressed the token-permission note in the review: the workflow defaults to contents: read, the release job explicitly gets contents: write, and every checkout sets persist-credentials: false.

Validation: the workflow YAML parses, assertions for token permissions and checkout credential settings pass, and git diff --check passes. Full CI has not run: this account cannot add the upstream testable label.

.github/workflows/main.workflow.yml
```diff
@@ -304,6 +340,8 @@ jobs:
         fi
 
   release:
+    permissions:
+      contents: write
```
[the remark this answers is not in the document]
lukas-plojhar: Fixed in 318c97484. Pinned actions/create-release to 89e8dc236279a907f495cdc4ebff2056b7036b3d, verified as the current upstream master commit. The release inputs, tag-only condition, and contents: write permission are unchanged. Workflow YAML parsing, release pin/permission assertions, and git diff --check pass.

.github/workflows/main.workflow.yml
```diff
@@ -352,7 +391,7 @@ jobs:
         # zip -r depthai-core-${{ steps.tag.outputs.version }}-win32-no-opencv.zip depthai-core-${{ steps.tag.outputs.version }}-win32-no-opencv
 
     # Create GitHub release
-    - uses: actions/create-release@master
+    - uses: actions/create-release@89e8dc236279a907f495cdc4ebff2056b7036b3d # master
```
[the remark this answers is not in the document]
lukas-plojhar: Fixed in 6855fd728. All remaining actions in the release job (checkout, get-version, setup-python, download-artifact, and upload-release-asset) now use full commit SHAs. Each SHA was resolved from the existing upstream version ref and verified through the upstream commit API; get-version-action v2 is a branch, and its resolved commit is pinned too. Version comments remain for readability.

Validation: YAML parsing and assertions confirm every active release action uses a 40-character SHA and the parsed workflow is otherwise unchanged. git diff --check passes. The tag-only release condition, action inputs, and contents: write permission are preserved.

.github/workflows/main.workflow.yml
```diff
@@ -131,6 +138,7 @@ jobs:
     - uses: actions/checkout@v4
       with:
         submodules: 'recursive'
+        persist-credentials: false
```
zrezke, line 141: Why are all these changes needed?

CMakeLists.txt
```diff
@@ -1057,15 +1075,16 @@ macro(add_runtime_dependencies depending_target dependency)
...
         add_custom_command(TARGET ${depending_target} POST_BUILD COMMAND
-            "$<$<BOOL:${required_dll_files}>:${CMAKE_COMMAND};-E;copy_if_different;${required_dll_files};$<TARGET_FILE_DIR:${depending_target}>>"
+            "$<$<BOOL:${dlls}>:${CMAKE_COMMAND};-E;copy_if_different;${dlls};$<TARGET_FILE_DIR:${depending_target}>>"
             COMMAND_EXPAND_LISTS
             VERBATIM
         )
-        message(STATUS "Required dlls for core are: ${required_dll_files}")
+        message(STATUS "Required dlls for ${depending_target} are: ${dlls}")
     endif()
 endmacro()
```
zrezke: I think change is only needed because we allow building both "core" and "device manager" at the same time. We should make these two targets mutually exclusive and not require this "fix" in the first place.

zrezke (changes requested):

PR #1982, opened 2026-09-15
Crashdump upload size limit

ViacheslavTelelyukhin:
## Purpose
When uploading crashdumps we may run into an issue where the file is too big and is rejected by our server.

## Specification
This will not send crashdump payloads if they make the request over 128MB. Also added fields for size of the files in case they are not attached and emitting warning when payloads are skipped.

## Deployment Plan
We may still want to delegate the upload to a separate thread before merging into develop

## AI Usage
None

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved diagnostic log uploads by processing crash dumps before other pipeline data.
  - Enforced a 128 MiB upload limit and excluded files exceeding the remaining capacity.
  - Uploads now account for cumulative file sizes and include file-size information.
  - Crash dump collection now runs asynchronously, helping prevent delays.
  - Errors encountered while recording crash dumps are logged without interrupting device operations.


ViacheslavTelelyukhin:
Also need to determine/set the exact cloudflare limit before merging. I just guessed it to be 128MB but did not do any experminets to find it.

PR #1983, opened 2026-09-15
Feature/detection parser input config

ViacheslavTelelyukhin:
## Purpose
DetectionParser like all parsers needed the ability to change configs at runtime.

## Specification
Added a similar config input queue like in other parsers. For example SegmentaionParser.
The agent also added a test but I'm not completely sure if we need it or if it should be integrated into the existing detection parser test. Similar test cases exist for other parsers so I kept it.

## Deployment Plan
Check if we want the test for configuring runtime configs. Is it integrated correctly?

## Testing & Validation
Tested this on device and it works. Also reviewed the code and did not see any red flags.

## AI Usage
Assisted-by: Codex Terra 5.6 high

## Summary by CodeRabbit

- **New Features**
  - Detection thresholds can now be updated at runtime through configuration messages.
  - Added confidence and IoU threshold controls to Python bindings.
  - Detection parser bindings expose startup and runtime configuration inputs.
  - C++ and Python detection examples now support keyboard-based threshold adjustments.

- **Bug Fixes**
  - Runtime threshold updates now apply consistently across supported detection models.

- **Tests**
  - Added coverage for synchronous and asynchronous runtime configuration, serialization, and default thresholds.


jakgra (approved):
LGTM

PR #1984, opened 2026-09-15
Stabilize dynamic calibration test

pheec:
Run the synthetic recalibration test three times and pass when at least two attempts succeed. This tolerates the known non-deterministic calibration result (RANSAC) while preserving the existing checks.

## Summary by CodeRabbit

* **Tests**
  * Replaced the synthetic recalibration test with a holistic replay-based calibration test.
  * Added validation that replay images provide complete calibration coverage.
  * Added checks for rotation and translation accuracy against expected values.
  * Added verification that recalibration reduces Sampson error compared with the current calibration.
  * Added configuration support for selecting the holistic replay recording used by the test.


MaticTonin (approved):
LGTM

ViacheslavTelelyukhin (approved):

pheec (commented):
LGTM

PR #1985, opened 2026-09-16
fixed beta_parser_config_test race condition

ViacheslavTelelyukhin:
## Purpose
There was race condition in the test causing it to fail occasinally

## Specification
Sent data directly to the node inputQueue instead of creating an input queue. This preserves the semantics of testing the node.

## Summary by CodeRabbit

* **Tests**
  * Updated runtime configuration coverage for min-max scaling.
  * Simplified how runtime configuration is submitted during parser tests.


jakgra (approved):
LGTM

PR #1989, opened 2026-09-18
removed race condittion from pipeleine event handler

ViacheslavTelelyukhin:
## Purpose
Sometimes the pipeline_debugging_host_test "State callback test" case would fail with incomplete timing data (mainLoopTiming.durationStats.averageMicrosRecent == 0). I believe this is due to a race condition.

## Specification
Locked the PipelineEventHandler mutex while it's updating

## Testing & Validation
I couldn't reproduce the test failure because it onlt rarely happens on specific machines

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved synchronization during pipeline statistics updates to help ensure more consistent and reliable results.


asahtik:
I'm kind of worried about the overhead of this - a pipeline can spawn a lot of events quickly and locking a compute intensive chunk could cause some of the events to get dropped (IIRC the input is non-blocking).

Maybe we could change the `updateStats` method to do the calculations on a copy and only lock the two assignments?

ViacheslavTelelyukhin:
Would this work?

asahtik:
Some state changes and reads are not thread safe in the latest commit. I had codex review the file (the develop version) and it looks like a bigger refactor is necessary to get it truly thread safe :cry:

PR #1991, opened 2026-09-18
Fix/matrix ops precision

JakubFara:
# Increase `matrixToVector` precision
add test for it
**before**:
```
depthai-core/tests/src/onhost_tests/matrix_ops_tests.cpp:84: FAILED:
REQUIRE_THAT( convertedCenter, Catch::Matchers::WithinAbs(originalCenter, 1e-7) )
with expansion:
  0.00501367097848561 is within 0.0000001 of 0.00513399904849939
```
**after**:
pass

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved rotation conversion accuracy for very small rotations and rotations near 180°, helping preserve camera positioning during round-trip conversions.

- **Tests**
  - Added coverage validating that small rotation conversions preserve the calculated camera center.


jakgra (approved):
LGTM

PR #1994, opened 2026-09-18
Feature/tof expose parameters

JakubFara:
# Expose tof parameters and presets

## Summary by CodeRabbit

* **New Features**
  * Added configurable VD55H1 time-of-flight controls for phase unwrapping, bilateral filtering, temporal noise reduction, and flying-pixel filtering.
  * Exposed VD55H1 settings through Python bindings.
  * Added C++ and Python examples with live controls for tuning VD55H1 depth processing.
  * Profile presets now provide corresponding VD55H1 settings.

* **Bug Fixes**
  * Preserved existing RVC2 phase-unwrapping settings when applying profiles.

* **Tests**
  * Added coverage for profile configuration and settings serialization.


CasualMathEnjoyer:
Tested this together with the [FW](https://gitlab.luxonis.com/luxonis/depthai/depthai-device-kb/-/merge_requests/380) on RVC4. Observations:

All parameters except for bilateral kernel size and phase unwrap threshold work as expected, including when changed at runtime.

BUT

- `bilateralKernelSize` doesn't change anything . I suggest we remove it. The docs currently say "odd integers from 3 to 15", which makes it look supported.
- `phaseUnwrapErrorThreshold` only applies at pipeline start, sending it at runtime does nothing. 
- examples have sliders for these two, so they look broken. I'd drop the kernel slider and either drop the unwrap threshold or add an option to restart pipeline.

PR #1999, opened 2026-09-22
Codex/fix detection roi flipping

CenekAlbl:
# Fix normalized detection boxes flipping at rectangle angle boundaries

## Why

`ImgDetections::transformTo()` can abruptly change a detection's aspect ratio when mapping it to a non-square image. This also affects ordinary detection alignment through `Align`.

`remapRectTo()` currently fits a rotated rectangle in pixels, then normalizes its local width and height by the image dimensions while leaving its angle unchanged. When the fitter crosses 90°, it swaps width/height and wraps the angle. That representation change is valid in pixels, but the subsequent normalization distorts the box.

## Reproduction

The MRE uses only Camera + DetectionNetwork, freezes the largest detection, and maps it onto an 800×500 image with rotations of −0.001° and +0.001°. Green is the bounds of individually transformed corners; red is `transformTo().getOuterBoundingBox()`. The tiny rotation is deliberate, so detection jitter and camera alignment are not needed to trigger the bug.

For the captured detection, the expected box is approximately **333.84×293.66 px**. Before the fix, the −0.001° case becomes **534.14×183.54 px**, while +0.001° remains correct.

[Before: the same detection becomes wide on the left]image

Save the inline MRE as `repro.py` and run with an affected SDK on RVC4 (requires OpenCV and NumPy). It draws only rectangles on a black canvas and saves only frame dimensions and box coordinates:

```bash
python repro.py --capture /tmp/roi-capture.npz --output /tmp/before.png --show
```

<details>
<summary>MRE code (Camera + DetectionNetwork)</summary>

```python
#!/usr/bin/env python3
"""Reproduce detection rectangle flipping on a black canvas.

Red: transformTo().getOuterBoundingBox(); green: bounds of remapped corners.
The same frozen detection is used in both panels, with +/-0.001 degree rotation.
"""
import argparse
import time

import cv2
import depthai as dai
import numpy as np


def compare(size, detection):
    width, height = 800, 500  # Non-square output is essential to expose the bug.
    sw, sh = size
    # Work in model-input image coordinates; no calibration or second camera is needed.
    source = dai.ImgTransformation(sw, sh)
    message = dai.ImgDetections()
    message.setTransformation(source)
    message.detections = [detection]
    corners = detection.getBoundingBox().denormalize(sw, sh).getPoints()
    panels = []
    for angle in (-0.001, 0.001):
        target = message.getTransformation()  # Independent copy for each panel.
        target.addScale(width / sw, height / sh)
        target.addRotation(angle, dai.Point2f(width / 2, height / 2, False))
        transformed = message.transformTo(target).detections[0]
        actual = np.array(transformed.getOuterBoundingBox()) * [width, height, width, height]
        points = np.array([(p.x, p.y) for p in (source.remapPointTo(target, p) for p in corners)])
        expected = np.concatenate((points.min(axis=0), points.max(axis=0)))
        panel = np.zeros((height, width, 3), dtype=np.uint8)
        for box, color, thickness in ((expected, (0, 255, 0), 4), (actual, (0, 0, 255), 2)):
            x1, y1, x2, y2 = np.rint(box).astype(int)
            cv2.rectangle(panel, (x1, y1), (x2, y2), color, thickness)
        fittedAngle = transformed.getBoundingBox().angle
        actualSize = actual[2:] - actual[:2]
        expectedSize = expected[2:] - expected[:2]
        print(f"rotation={angle:+.3f} deg, fitted={fittedAngle:.3f} deg; "
              f"actual={actualSize.round(2)}, expected={expectedSize.round(2)} pixels", flush=True)
        for row, text in enumerate((f"Rotation {angle:+.3f} deg / fitted {fittedAngle:.3f} deg",
                                    "GREEN: remapped corners; RED: transformTo bounds")):
            cv2.putText(panel, text, (12, 25 + row * 25), cv2.FONT_HERSHEY_SIMPLEX, 0.55, (255, 255, 255), 2)
        panels.append(panel)
    return np.hstack(panels)


def main():
    parser = argparse.ArgumentParser(description=__doc__)
    parser.add_argument("--output", default="/tmp/detection-transform-repro.png")
    parser.add_argument("--show", action="store_true", help="Also open an OpenCV window")
    parser.add_argument("--capture", help="Save the frozen input as an NPZ file")
    parser.add_argument("--replay", help="Replay an NPZ capture without a device")
    args = parser.parse_args()
    if args.replay:
        with np.load(args.replay, allow_pickle=False) as capture:
            x, y, w, h, angle, normalized = capture["box"]
            box = dai.RotatedRect(dai.Point2f(x, y, bool(normalized)), dai.Size2f(w, h, bool(normalized)), angle)
            result = compare(capture["size"], dai.ImgDetection(box, 1.0, 0))
    else:
        size, detection = captureDetection()
        if args.capture:
            box = detection.getBoundingBox()
            np.savez_compressed(args.capture, size=size,
                                box=[box.center.x, box.center.y, box.size.width, box.size.height, box.angle, box.isNormalized()])
        result = compare(size, detection)
    if not cv2.imwrite(args.output, result):
        raise RuntimeError(f"Could not save {args.output}")
    print(f"Saved {args.output}")
    if args.show:
        cv2.imshow("Same detection, two tiny rotations", result)
        cv2.waitKey(0)
        cv2.destroyAllWindows()


def captureDetection():
    with dai.Pipeline() as pipeline:
        camera = pipeline.create(dai.node.Camera).build(dai.CameraBoardSocket.CAM_A, sensorFps=30)
        detector = pipeline.create(dai.node.DetectionNetwork).build(camera, dai.NNModelDescription("yolov6-nano"))
        detector.setConfidenceThreshold(0.25)
        frames = detector.passthrough.createOutputQueue(maxSize=4, blocking=True)
        detections = detector.out.createOutputQueue(maxSize=4, blocking=True)
        pipeline.start()
        warmupEnd = time.monotonic() + 2
        deadline = time.monotonic() + 30
        frame = message = None
        print("Waiting for a detection...", flush=True)
        while time.monotonic() < deadline:
            frame = frame if frame is not None else frames.tryGet()
            message = message if message is not None else detections.tryGet()
            if frame is None or message is None:
                time.sleep(0.005)
                continue
            if frame.getSequenceNum() < message.getSequenceNum():
                frame = None
            elif message.getSequenceNum() < frame.getSequenceNum():
                message = None
            elif message.detections and time.monotonic() >= warmupEnd:
                detection = max(message.detections, key=lambda d: d.getWidth() * d.getHeight())
                return (frame.getWidth(), frame.getHeight()), detection
            else:
                frame = message = None
        else:
            raise RuntimeError("No synchronized detection received within 30 seconds")


if __name__ == "__main__":
    main()
```

</details>

## Fix

In `ImgTransformation::remapRectTo()`, remap each corner in its original coordinate units and fit the rectangle in the destination coordinate space. For normalized input, `remapPointTo()` performs the normalization on each corner before fitting. Preserve the output's normalized/pixel flags explicitly.

The core change is:

```cpp
const bool normalized = rect.isNormalized();
const auto points = rect.getPoints();
std::vector<std::array<float, 2>> mapped(points.size());
for(std::size_t i = 0; i < points.size(); ++i) {
    const auto point = remapPointTo(to, points[i]);
    mapped[i] = {point.x, point.y};
}
auto result = impl::getOuterRotatedRect(mapped);
result.center = Point2f(result.center.x, result.center.y, normalized);
result.size = Size2f(result.size.width, result.size.height, normalized);
```

This removes the erroneous normalization of the fitted rotated rectangle. No node-specific workaround, public API, ABI, or serialization change is needed. The return value remains an enclosing rotated rectangle; this does not change the separate `projectRectTo()` path or standalone `RotatedRect::normalize()`.

Replay the same saved geometry with the fixed SDK:

```bash
python repro.py --replay /tmp/roi-capture.npz --output /tmp/after.png --show
```

After the fix, both rotations produce **333.84×293.66 px**, matching the corner-based reference. Both plots use exactly the same saved geometry, rendered on black backgrounds.

[After: both rotations preserve the detection bounds]image

## Validation

- Rebuilt the SDK on `develop` (`e5a4f431c`) with and without the fix and replayed the identical RVC4 capture. Live Camera + DetectionNetwork also passes with the fixed SDK.
- Both new regression tests fail without the fix (37 failed assertions).
- With the fix: **50,581 assertions pass across all 20 image-transformation test cases**, covering angle boundaries, non-square images, and coordinate units.
- C++ formatting and datatype/protobuf consistency checks pass. Public API and serialization are unchanged.

- Full SDK/Python rebuild succeeded, including all 114 C++ test executables. **90/97 C++ executables passed** (39/40 host, 51/57 RVC4); all seven failures also reproduce with the original `develop` implementation. The remaining 17 require unavailable hardware.
- **110 Python cases passed across runs**; the remote-connection service test aborts. All alignment tests passed, and device calibration is unchanged. The full suite is not green; no regression attributable to this patch was identified.

## Review

- [ ] Review `remapRectTo()` logic: fit transformed corners in the destination coordinate units and preserve normalized/pixel flags, including near 0°/90° boundaries on non-square images.
- [ ] Run the inline MRE before and after the fix using the same captured geometry; confirm both rotations match the green reference bounds after the fix.
- [ ] Review regression coverage and HIL results on RVC4, including alignment tests, baseline-reproduced failures, and hardware coverage gaps.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved rotated rectangle remapping so bounding boxes retain accurate bounds across image scaling, rotation, and angle wrapping.
  - Preserved the coordinate units of rectangles during transformations, including small pixel-based coordinates.
  - Added validation to reject normalized rectangles when source or target dimensions are zero.


MaticTonin:
Please change it to develop, then we will revirew.

MaticTonin (approved):
Tested on both, RVC2 and RVC4 and it worked fine yes

PR #2000, opened 2026-09-22
fixed pipeline debuggging test checking unused queue

ViacheslavTelelyukhin:
## Purpose
The pipeline_debugging_rvc4_test would consistently fail because it would assert the timing for inputConfig on detection parser exists but would not send any messsages to that queue meaning no timings could be produced.

## Summary by CodeRabbit

- **Tests**
  - Updated pipeline debugging validation to ignore unused inputs while preserving timing checks for relevant inputs.
  - Adjusted detection parser test formatting and include ordering without changing behavior.


aljazdu:
Closed as its a duplicate

PR #2002, opened 2026-09-23
fixed config input queue delivery race conditions in several tests

ViacheslavTelelyukhin:
## Purpose
Simialr to 1985 there was a race condition in multiple tests where the delivery of a config through input queue could happend after the asertion reliant on that config.

## Specification
Instad configs are now delivered to the node directly without a scheduling queue in the middle

We probably need to write some separate tests for queues.

## Testing & Validation
Ran beta_parser_config_test and detection_parser_runtime_test 100 times. The rest a few times as well but they were slower.

## AI Usage
Codex astra high was responsible for identifying instances of the bug across the test suite.

## Summary by CodeRabbit

* **Tests**
  * Updated runtime parser configuration test coverage across Python, on-host, and on-device scenarios to validate configuration delivery through the parser’s configuration input.
  * Adjusted test handling for asynchronous device configuration and parsers that may need an input frame before returning output.


aljazkonec1 (approved):
Side note: this only fixes the race issue because pipeline is running on host so the message is written directly to the parsers queue when using `parser.inputConfig.send(config)` If pipeline is running on device, the queue would stay empty on the device

